PR #969, opened 2026-07-31
test: fix the 2 reds + 1 flake in the scheduled full e2e run

mersahin:
Run [30608852159](https://github.com/21072026/Internship/actions/runs/30608852159) (2026-07-31 03:00 UTC) reported **288/291**: two deterministic failures and one flake. None of the three is a product bug.

Refs #963.

### 1. `evaluation-goals` — stale assertions kept by a merge

The spec still asserted the goals panel's old progress bar:

```ts
await expect(page.getByText('0/2 completed')).toBeVisible();
await expect(page.getByText('0/2 completed').locator('..').getByText('0%', { exact: true })).toBeVisible();
```

#785 (PR #786) replaced that bar with the `goals-active-count` / `goals-completed-count` counters. The merge (85df9f7) took **main's** `GoalsPanel` (the counters version from #918) together with **the branch's** spec, so the spec asserts markup that no longer renders. `goals-archive-sort.spec.ts`, which shipped with the same feature, already asserts the counters — `evaluation-goals` now does the same (3 goals: 2 active, 1 completed).

### 2. `meeting-requests` — sign-in clicked a button on the wrong page

```
locator resolved to 5 elements. Proceeding with the first one:
<button disabled type="submit" …>Add</button>
  - element is not enabled … (15s)
```

The test switches user mid-test, and `clearCookies()` alone does not end the old session: the page being left keeps calling `/api/auth/session`, and NextAuth re-issues the session cookie on those responses — one landing just after the clear restores it. `/auth/signin` then sees `status === 'authenticated'` and `router.replace()`s to the *previous* user's dashboard while the test is still typing, so the unscoped `page.click('button[type="submit"]')` re-resolves against that dashboard. The failure screenshot in `playwright-report-full-shard-3` is the **mentee portal**, and the button being retried is its disabled "Add goal".

New `signInAsFreshUser()` in `e2e/helpers/auth.ts`:
- `goto('about:blank')` first, so the old page and its in-flight requests are gone *before* the session cookie is dropped;
- clears only the NextAuth session cookie — a blanket `clearCookies()` also drops the cookie-consent state seeded via `storageState`, which brings the banner back over the form;
- clicks the submit button **inside the sign-in form**, so a stray redirect fails fast instead of clicking something unrelated.

`evaluation-goals` carried the same copy of the old helper and uses it too.

### 3. `notes` — the run's flake

`expect(page.getByText('Prepare portfolio for interview')).toBeVisible()` followed immediately by a Prisma read. Playwright's text matching includes `<textarea>` values, so that matched the text just typed into the still-open editor — before the PATCH was even sent — and the read raced the write. It now waits for the editor to close (the panel only drops the textarea after the PATCH resolves).

### Verification

`npx tsc --noEmit` clean. No local MySQL on this machine, so the specs are verified by a `workflow_dispatch` run of **e2e-full** on this branch — link in a comment below; merging once it and the smoke gate are green.

No version bump: test-only changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,32 @@ version is shown in the sidebar footer of every page (links to the
 ## [Unreleased]
 
 ### Fixed
+- **Scheduled full e2e suite: the 2 reds and 1 flake left by run
+  [30608852159](https://github.com/21072026/Internship/actions/runs/30608852159)**
+  (#963). Again no product bug — all three are test defects.
+  - **`evaluation-goals`** still asserted the goals panel's old `"0/2 completed"`
+    progress bar and its `0%` label. #785 (PR #786) replaced that bar with the
+    `goals-active-count` / `goals-completed-count` counters, but the merge kept
+    main's `GoalsPanel` (from #918) *and* the branch's spec, so the spec asserted
+    markup that no longer exists. It now asserts the counters, like the
+    `goals-archive-sort` spec that shipped with the same feature.
+  - **`meeting-requests`** switches user mid-test, and `clearCookies()` alone does
+    not end the old session: the page being left keeps hitting
+    `/api/auth/session`, and NextAuth re-issues the session cookie on those
+    responses — one landing just after the clear restores it. `/auth/signin` then
+    saw `status === 'authenticated'` and redirected to the *previous* user's
+    dashboard mid-typing, so `page.click('button[type="submit"]')` re-resolved
+    against the mentee portal and spent the whole action timeout retrying its
+    disabled "Add goal" button. New `signInAsFreshUser()` in `e2e/helpers/auth.ts`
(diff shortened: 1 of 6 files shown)


mersahin:
Full-suite verification run for this branch: https://github.com/21072026/Internship/actions/runs/30618267526

mersahin:
Verified: full-suite dispatch on this branch is **green — 291/291, 0 failed, 0 flaky** ([run 30618267526](https://github.com/21072026/Internship/actions/runs/30618267526)).

- `evaluation-goals` ✓ (shard 2, 2.0s)
- `meeting-requests` ✓ (shard 3, 1.7s — previously burned 17s into the action timeout)
- `notes` ✓ (shard 3, 1.6s — passed without a retry)

Smoke gate + Lint/Typecheck/Build also green. Merging.

PR #991, opened 2026-07-31
fix(messages): project group chat emptied the "New chat" picker

mersahin:
## Why

The scheduled full e2e run ([run 30648248486](https://github.com/21072026/Internship/actions/runs/30648248486)) went red on `e2e/project-dm.spec.ts`:

```
TimeoutError: locator.click: Timeout 15000ms exceeded.
  - waiting for getByTestId('new-chat-toggle')
```

The toggle wasn't slow — it was never rendered, and this is a real user-facing regression, not a flaky test.

## Root cause

[src/app/messages/page.tsx](src/app/messages/page.tsx) excludes people the viewer already has a DM with from the "new chat" candidate list, and built that exclusion set from **every** conversation it had loaded:

```ts
const existingDmPartnerIds = new Set(
  conversations.flatMap((c) => c.participants.map((p) => p.userId)).filter((id) => id !== me),
);
```

That was correct while the query was `type: 'DIRECT'` only. 988d791 ("expose project group chats") widened it to `DIRECT ∪ the viewer's project GROUP chats` — and every project co-member is a participant of the shared project's group chat. So every candidate matched the exclusion, `candidates` came out empty, and `StartConversationPicker` returns `null` on an empty list.

Net effect: **anyone who is a member of any project saw no "New chat" option at all** and could not start a project DM from the UI. Not just the test.

## Fix

Build the exclusion set from `DIRECT` conversations only.

## Also in this PR

`E2E Tests` takes an optional `grep` input on manual dispatch (default `@smoke`), so a single non-smoke spec can be re-verified on a branch without dispatching the 4-shard `e2e-full` suite and its summary email. The filter is passed through an env var rather than interpolated into the shell (a dispatch input is untrusted text).

## Verification

- `tsc --noEmit` and `npm run check:i18n` clean.
- No local MySQL in this environment, so the spec itself is verified in CI: `E2E Tests` dispatched on this branch with `grep` set to the failing test title (linked in a comment below).

## Checklist

- [x] `package.json` → `0.32.2-beta`
- [x] `CHANGELOG.md` entry
- [x] `src/lib/releaseNotes.ts` entry (EN/TR/DE)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -4,7 +4,15 @@ on:
   # Re-enabled 2026-07-21 after the repo transfer reset the Actions quota. This
   # is the @smoke-only PR gate (~3.5 min) — cheap enough to run per PR. The full
   # suite (e2e-full) stays workflow_dispatch-only to protect the monthly budget.
+  # A manual run takes an optional Playwright --grep filter, so a single
+  # non-smoke spec can be re-verified on a branch without dispatching the whole
+  # 4-shard e2e-full suite (which also emails a summary on every run).
   workflow_dispatch:
+    inputs:
+      grep:
+        description: 'Playwright --grep filter (e.g. @smoke, or a test title)'
+        required: false
+        default: '@smoke'
   push:
     branches: [main]
   pull_request:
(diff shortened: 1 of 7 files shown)


mersahin:
Verification: dispatched `E2E Tests` on this branch with the new `grep` input set to the failing test title — [run 30663869932](https://github.com/21072026/Internship/actions/runs/30663869932):

```
Running 1 test using 1 worker
  ✓  1 [chromium] › e2e/project-dm.spec.ts:23:5 › project co-members can start a DM, and lose the composer when the project link goes away (2.2s)
  1 passed (7.8s)
```

For the record, the other 5 non-green tests in [run 30648248486](https://github.com/21072026/Internship/actions/runs/30648248486) were all flaky (`failed,passed`), unrelated to this change, and all one shape — a post-sign-in redirect interrupting an immediate `page.goto` (`Navigation to /mentor/mentees/X is interrupted by another navigation to /mentor`), or the sign-in form not rendering within 15s: `comms`, `company-shortlist`, `mentee-detail-feedback`, `mentorship-pagination`, `profile-activity`. Not addressed here.

PR #992, opened 2026-07-31
feat(announcements): attach an image to a broadcast

mersahin:
## What

`/admin/announcements` only accepted text + a link, so anything visual had to be hosted elsewhere and pasted in as a URL. Announcements can now carry an image.

- **Composer**: optional picker (PNG/JPEG/WebP/GIF, ≤ 5 MB) with a local object-URL preview, replace/remove.
- **Readers**: the image renders on `/announcements` (full width, capped height), as a thumbnail in the dashboard `AnnouncementsCard`, and in the admin history list.
- **Email**: with *"also send by email"* checked, the image goes out as an **inline `cid:` attachment** — the serving route requires a session, so an `` would have been a broken image in every mail client. `sendEmail`'s `attachments` gained an optional `cid`.

## How

- New `AnnouncementImage` model — blob in the DB, same approach as `AvatarFile`/`CvFile`, so it survives container redeploys. Cascade-deleted with its announcement.
- Served from `GET /api/announcements/<id>/image`, gated on any authenticated session (matching `GET /api/announcements` — every broadcast targets all active users) with `nosniff` + `Content-Disposition: inline`.
- **No `imageUrl` column.** The URL is a pure function of the announcement id, so mirroring it would only add a second write that can disagree with reality. Both feed APIs select `image: { id: true }` — never the blob — and derive the URL.
- `POST /api/admin/announcements` now accepts `multipart/form-data` **in addition to** JSON; the JSON contract is untouched, so existing API clients and the three specs that post JSON keep working. An unparseable body is now a 400 instead of an unhandled 500.

## Validation

`src/lib/announcementImage.ts` is the single source of truth, imported by both the picker and the route, so a rejected file is reported at the picker instead of as a bare 400 after Broadcast (the `textLimits.ts` reasoning).

- **SVG deliberately excluded** — it can carry `<script>` and the blob is served from our own origin, so allowing it would turn "attach an image" into stored XSS against every reader.
- **Leading bytes checked**, so a file that merely *claims* to be a PNG is refused — and refused *before* the notification fan-out, so a rejected image can't leave behind notifications for a broadcast that never landed.
- The email attachment filename is derived from the allow-listed MIME type, not from the upload.

## Tests

New `e2e/announcement-image.spec.ts` (4 tests, left out of `@smoke` to keep the PR gate small):
1. admin attaches an image → 201, `AnnouncementImage` row correct, endpoint serves `image/png` + `nosniff` with the right byte length, and a *mentee* sees it on `/announcements`.
2. the image endpoint is 401 from a session-less context.
3. a `text/plain` file (`setInputFiles` bypasses `accept`) is rejected at the picker, no preview.
4. bytes that only claim to be `image/png` → 400, and no announcement/notification rows.

## Verification done locally

`npx tsc --noEmit`, `npm run lint`, `npm run build` (new route registered), `npm run check:i18n` (1591 keys × 3 locales) all clean. **No browser/e2e run locally**: this machine has no MySQL and no Docker daemon, and the only `DATABASE_URL` on the box points at the shared preview DB, which is off-limits for development — so the functional proof is the e2e suite in CI against its own MySQL service.

## Checklist
- [x] `package.json` → `0.33.0-beta` (minor: user-visible feature)
- [x] `CHANGELOG.md` entry
- [x] `src/lib/releaseNotes.ts` entry (EN/TR/DE)
- [x] i18n keys in all three dictionaries
- [x] `prisma format && validate && generate` (schema-first, no SQL migration — CI runs `db push`)


mersahin:
Kapatıldı: GitHub bu PR için ilk açılıştan sonra hiçbir `pull_request` (synchronize/reopened) olayı tetiklemedi, bu yüzden zorunlu "Deploy topic environment" check'i head commit'te hiç oluşmadı ve auto-merge ilerleyemedi. Aynı daldan yeni bir PR açıyorum; kod aynı.

PR #976, opened 2026-07-31
feat: add global search to mentor portal

sudebegumyksl:
## Summary

Adds the existing global search component to the mentor portal. The search API already supported the `MENTOR` role and restricted results to the mentor's own mentees; this change exposes that functionality in the mentor UI and routes results to the appropriate mentor pages.

Closes #<issue-number>

## Changes

- Added `GlobalSearch` to the mentor layout.
- Updated mentor search results to navigate to `/mentor/mentees/<relationId>`.
- Included `relationId` in mentor search API responses while preserving existing admin behavior.
- Added a dedicated `data-testid` to the global search input for reliable E2E locators.
- Added an E2E test verifying:
  - mentors can find their own mentees,
  - mentors cannot see another mentor's mentees,
  - mentor search results navigate to the correct page.
- Updated package version, CHANGELOG, and release notes.

## Verification

- [x] `npm run check:i18n`
- [x] `npm run build`
- [x] `npx playwright test e2e/mentor-global-search.spec.ts e2e/search.spec.ts e2e/admin-nav-filter.spec.ts`
- [x] `git diff --check`
- [x] Manually verified mentor global search and navigation

PR #1013, opened 2026-08-01
feat(evaluations): delete an evaluation recorded by mistake

mersahin:
## Problem

The evaluation panel only ever appended. A rating clicked by mistake stayed on the record forever — there was no `DELETE` route and no UI affordance.

## What changed

- **`DELETE /api/evaluations/[id]`** — removes one evaluation. Only its **own author** (or an `ADMIN`) may do so: an evaluation is the author's judgement, so the other side of the relation cannot erase one written about them. The removal is written to the activity log as `evaluation.deleted` (actor, target, relation, type, request origin).
- **`GET /api/evaluations`** now returns a `canDelete` flag per row, so the panel offers the button exactly where the DELETE route would allow it.
- **`EvaluationPanel`** — trash icon on each entry it can delete, `window.confirm` first, list reloads after.
- i18n: `evaluation.delete` / `evaluation.confirmDelete` in EN/TR/DE (`check:i18n` green).
- e2e: mentor deletes their own evaluation (200, row gone) and is refused the mentee's (403, row intact).
- Version `0.38.0-beta` + CHANGELOG + release-notes entry.

## Verification

- `npm run lint` and `tsc --noEmit` clean (only pre-existing warnings and the two optional-dep errors in `inboundMailBridge.ts`).
- The new spec is **not** `@smoke`, so the PR gate does not run it — dispatched `e2e.yml` on this branch with `--grep "delete an evaluation they recorded"` to exercise it directly; result reported in a comment below.
- No local MySQL in this environment, so the panel itself was not clicked through in a browser.

## Notes

- Rebased onto `main` after #1014 and #976 landed mid-session; the version bump was redone on top (`0.37.0-beta` → `0.38.0-beta`).
- The first push produced **zero** check-runs on the head commit — the dropped-`pull_request`-event failure already recorded in `docs/agent-experience.md`. The force-push after the rebase re-triggered all five workflows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ version is shown in the sidebar footer of every page (links to the
 [user-facing release notes](src/lib/releaseNotes.ts), rendered at
 `/release-notes`) and in the landing-page footer.
 
+## [0.38.0-beta] - 2026-08-01
+
+### Added
+- **An evaluation can be deleted** (`DELETE /api/evaluations/[id]`, trash icon in
+  `EvaluationPanel`). Until now a mis-clicked rating was permanent: the panel only ever
+  appended, and there was no route to remove a row. Only the evaluation's **own author**
+  (or an ADMIN) may delete it — an evaluation is the author's judgement, so the other
+  side of the relation cannot erase one written about them. The list endpoint now returns
+  a `canDelete` flag per evaluation so the button only appears where the DELETE route
+  would actually allow it, and the deletion is recorded in the activity log
+  (`evaluation.deleted`).
+
 ## [0.37.0-beta] - 2026-08-01
 
 ### Added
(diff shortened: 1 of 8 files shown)


mersahin:
The new spec is not `@smoke`, so the PR gate doesn't run it. Dispatched `e2e.yml` on this branch with `--grep "delete an evaluation they recorded"` ([run 30699277247](https://github.com/21072026/Internship/actions/runs/30699277247)) — it exercised the authorization matrix against a real MySQL:

```
Running 1 test using 1 worker
  ✓  1 [chromium] › e2e/evaluation.spec.ts:36:5 › a mentor can delete an evaluation they recorded by mistake, but not the mentee's (1.4s)
  1 passed (7.1s)
```

So both branches are covered for real: `403` on the mentee's evaluation with the row still present, `200` on the mentor's own with the row gone.

PR #1017, opened 2026-08-01
fix(e2e): repair the two specs the scheduled full run broke on

mersahin:
## Why

Scheduled full run [30696957141](https://github.com/21072026/Internship/actions/runs/30696957141) went red on two specs. Both were `failed,failed` — no retry rescued either — so both are real breakage, not flakes. Both came from changes merged *after* the last green full run.

## 1. `inbound-email.spec.ts` — 401 instead of 200

#870 (cea7a06) removed the inbound webhook's fail-open outside development:

```ts
if (!expected) return process.env.NODE_ENV !== 'production';
```

CI serves a production build (`npm run start`), so `NODE_ENV=production`, `INBOUND_SECRET` is unset, and `/api/inbound-email` answered **401 to every call** in the suite. The endpoint is behaving exactly as intended — the test environment was relying on the fail-open that #870 deliberately closed.

Fix: give the Playwright web server an `INBOUND_SECRET` (alongside the existing `HEALTH_TOKEN`, same pattern) and send it from the spec. This is the better test anyway — it now exercises the shape production actually runs: shared secret **required and supplied**, with the HMAC reply token as the second factor rather than the only one. Added a case pinning the gate itself (no header → 401), which nothing covered before.

`cron-start.spec.ts` still passes: with the secret configured, `/api/inbound-email/poll` returns 401 for both a missing and a wrong header, and that spec accepts `[401, 503]`.

## 2. `date-formatter-locale.spec.ts` — got `"unknown"`, expected a date

#881 added an IP chip to each `/admin/activity` row:

```tsx
{e.ip && <span className="hidden sm:inline font-mono text-xs text-gray-400 …" title={e.userAgent}>{e.ip}</span>}
<span className="text-xs text-gray-400 …">{formatDateTime(e.createdAt, locale)}</span>
```

The spec's `span.text-gray-400.first()` now resolves to the IP chip, which comes first in the DOM. It read `"unknown"` — `clientIp()` returns that sentinel when there is no proxy header, and the e2e server deliberately runs with `TRUSTED_PROXY_COUNT=0`.

Fix: `data-testid="activity-date"` on the date span, and the spec targets it. This is the exact pitfall CLAUDE.md documents under "E2E locator pitfalls".

## Verification

`E2E Tests` dispatched on this branch with `grep` covering both specs (linked in a comment below), plus `tsc --noEmit` and `check:i18n` clean.

## Not in this PR

- **The 6 flakies** (`announcements`, `comms`, `mentee-detail-feedback`, `mentee-setpw`, `profile-activity`, `two-factor`) — all `failed,passed`, all the same sign-in/navigation race. Worth a dedicated fix in the shared sign-in helper; not bundled with a red-run repair.
- **No version bump.** The only `src/` change is a `data-testid` attribute; everything else is test and CI config. Per the CHANGELOG rule, non-user-visible changes don't get a bump.

### DIFF
--- a/e2e/date-formatter-locale.spec.ts
+++ b/e2e/date-formatter-locale.spec.ts
@@ -24,7 +24,9 @@ test('dates on admin/activity follow the selected app language, not the browser 
     await page.goto('/admin/activity');
 
     // TR formatting uses dots (gg.aa.yyyy), never the en-US slash format.
-    const firstDate = page.locator('span.text-gray-400').first();
+    // Target the test id, not `span.text-gray-400` — the row also carries an
+    // IP chip in that colour (#881), and it comes first in the DOM.
+    const firstDate = page.getByTestId('activity-date').first();
     await expect(firstDate).toBeVisible();
     await expect(firstDate).toHaveText(/\d{2}\.\d{2}\.\d{4}/);
     await expect(firstDate).not.toHaveText(/\d{1,2}\/\d{1,2}\/\d{4}/);
--- a/e2e/inbound-email.spec.ts
+++ b/e2e/inbound-email.spec.ts
@@ -1,6 +1,13 @@
 import { test, expect } from '@playwright/test';
 import { createHmac } from 'crypto';
 import { prisma, seedUser, cleanupByEmail, uniqueEmail } from './helpers/db';
+import { E2E_INBOUND_SECRET } from '../playwright.config';
+
+// The webhook's shared secret is mandatory outside development (#870), and CI
+// serves a production build — so every call here carries it. That is also the
+// production shape: the token gate below is the *second* factor, not the only
+// one.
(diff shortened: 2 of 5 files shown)


mersahin:
Verification — [run 30702268448](https://github.com/21072026/Internship/actions/runs/30702268448), `E2E Tests` dispatched on this branch with `grep` covering both specs:

```
Running 2 tests using 1 worker
  ✓  1 [chromium] › e2e/date-formatter-locale.spec.ts:7:5 › dates on admin/activity follow the selected app language, not the browser default (1.8s)
  ✓  2 [chromium] › e2e/inbound-email.spec.ts:26:5 › inbound email reply is routed to the thread (token + sender verified) (471ms)
  2 passed (8.7s)
```

PR #1018, opened 2026-08-01
test(e2e): put the six flaky specs on the shared sign-in helpers

mersahin:
## Why

The six flakies in scheduled run [30696957141](https://github.com/21072026/Internship/actions/runs/30696957141) (`failed,passed` — green on retry) are two known shapes, and [e2e/helpers/auth.ts](e2e/helpers/auth.ts) already exists to prevent both. These specs simply never adopted it. The same race also showed up in the previous scheduled run, so it isn't settling on its own.

| shape | specs |
|---|---|
| `Navigation to /X is interrupted by another navigation to /admin` | `announcements`, `mentee-detail-feedback`, `mentee-setpw` |
| action times out on an element belonging to the **previous** page | `comms`, `profile-activity`, `two-factor` |

The second shape is the blanket-`clearCookies()` trap already documented on `signInAsFreshUser`: it also drops the consent cookie seeded via `storageState`, and the page being left keeps calling `/api/auth/session`, which re-issues the session cookie — so `/auth/signin` sees `authenticated` and `router.replace()`s to the old dashboard while the test is still typing. Concretely, `two-factor`'s submit resolved against the previous page's **disabled "Add note" button** and retried it until the 15s action timeout:

```
locator resolved to <button disabled type="submit" …>Add note</button>
```

`comms` sat on a password field that was about to be replaced.

## What changed

- `signInAndSettle` + `gotoSettled` for the deep-link cases.
- `signInAsFreshUser` for the user switches.
- `two-factor` needs those guards **without** the landing wait — a 2FA account is *supposed* to stay on `/auth/signin` and show the authenticator field. The body of `signInAsFreshUser` is now exported as `submitSignInForm`, and both share it.

No product code changes.

## Verification

`E2E Tests` dispatched on this branch with `grep` covering all eight affected tests (linked in a comment below). `tsc --noEmit` clean.

## Not in this PR

16 other specs still call blanket `clearCookies()`. Some do it deliberately (`legal-consent`, `cookie-consent-categories` need the consent cookie gone), so a blind sweep would break them. Worth a follow-up pass that goes file by file — this PR fixes the ones that actually flaked.

## No version bump

Test-only change, no user-visible behaviour.

### DIFF
--- a/docs/agent-experience.md
+++ b/docs/agent-experience.md
@@ -1913,3 +1913,19 @@ dışa aç, `signInAsFreshUser` onu çağırsın — guard'lar tek yerde kalır.
 **Süreç notu:** iki iş (kırmızı onarımı + flaky onarımı) ayrı PR'lara ayrıldı. Aynı dala
 yığmak, kırmızı düzeltmesinin merge'ünü flaky işinin CI'sine bağlardı; `git stash` + yeni
 dal (`git checkout -B <yeni> origin/main` + `stash pop`) bunu 10 saniyede ayırıyor.
+
+**Ek ders (aynı oturum, ilk denemem kırdı):** flaky'leri helper'a taşırken
+`submitSignInForm`'u "about:blank → sadece session cookie'sini sil → /auth/signin" olarak
+yazmak `two-factor.spec.ts`'i **deterministik olarak** bozdu (failed,failed) — parola alanı
+hiç gelmedi. Sebep: `/auth/signin` girişi tamamlarken `/api/auth/session`'ı yoklayıp
+`window.location.assign(roleHome)` yapıyor. `waitForURL()` dönüyor ama **terk edilen sayfa
+hâlâ session okuyor**; `clearCookies()`'ten sonra düşen bir yanıt cookie'yi geri yazıyor,
+sıradaki `/auth/signin` `authenticated` görüp dashboard'a `router.replace()` ediyor.
+Ekranda bunu söyleyen hiçbir şey yok — belirti sadece "parola alanı 15 sn'de gelmedi".
+Çözüm: cookie'yi düşürmeden **önce** çıkan sayfanın susmasını bekle (`networkidle`), form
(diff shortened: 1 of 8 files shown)


mersahin:
Verification — [run 30702780693](https://github.com/21072026/Internship/actions/runs/30702780693), all eight affected tests, no retries used:

```
Running 8 tests using 1 worker
  ✓  1 announcements.spec.ts:9 › sending an announcement persists it… (2.7s)
  ✓  2 comms.spec.ts:12 › messages track read/unread… (5.6s)
  ✓  3 comms.spec.ts:45 › admin broadcasts an announcement… (5.8s)
  ✓  4 mentee-detail-feedback.spec.ts:11 › empty interaction log… (2.3s)
  ✓  5 mentee-setpw.spec.ts:9 › set-password link… (4.7s)
  ✓  6 profile-activity.spec.ts:9 › activity visible to admin and mentor… (4.8s)
  ✓  7 two-factor.spec.ts:14 › enables TOTP 2FA… (5.7s)
  ✓  8 two-factor.spec.ts:61 › failed authenticator codes are rate limited (9.7s)
  8 passed (47.9s)
```

**Worth reading before approving** — the first cut of this PR ([run 30702528249](https://github.com/21072026/Internship/actions/runs/30702528249)) passed 7/8 and broke `two-factor` *deterministically*. Writing `submitSignInForm` as "about:blank → clear the session cookie → /auth/signin" wasn't enough: [signin/page.tsx](src/app/auth/signin/page.tsx#L104) finishes sign-in by polling `/api/auth/session` and then doing a full-page `window.location.assign(roleHome)`, so `waitForURL()` returns while the outgoing page is still reading the session. A read landing after `clearCookies()` re-issues the cookie, `/auth/signin` sees `authenticated` and `router.replace()`s away — and the only symptom is a timeout on the password field.

The old blanket `clearCookies()` was masking that by side effect, not by design. Commit 88f094f waits for the outgoing page to go quiet before dropping the cookie, and reloads once if the form still isn't there.

PR #2146, opened 2026-09-03
fix(auth,infra): stop leaking Prisma errors to the login form; add a MySQL watchdog

### LINKED ISSUE #2145: Login form prints the raw Prisma error when MySQL is OOM-killed; nothing restarts the database
## What happens

While MySQL is down, the sign-in form shows the driver's own message to the visitor:

```
Invalid `prisma.user.findUnique()` invocation: Can't reach database server at
`localhost:3306` Please make sure your database server is running at `localhost:3306`.
```

`service mysql start` fixes it. The database is being **OOM-killed** on the server, so
this recurs — and nothing on the box brings it back, so it stays down until somebody
notices.

## Two separate problems

**1. The message should never reach the browser.** #1150 added `guardProviders` in
`src/lib/auth.ts` exactly to prevent this, and it does not run: `CredentialsProvider({…})`
keeps the caller's real `authorize` in its `options` bag (the top-level one is a stub
returning `null`), and NextAuth's `parseProviders` merges that bag back *over* the
provider — reinstating the unwrapped function. So the wrapper wraps a function that is
never called. A raw driver message on a login form is useless to the user and useful to
an attacker (host, port, ORM, query shape).

**2. Nothing restarts the database.** The packaged systemd unit either does not restart
after a SIGKILL (which is what an OOM kill is) or gives up after a few rapid restarts
(`StartLimitBurst`), and no alert goes anywhere.

## What "done" looks like

- An unreachable database produces a stable, opaque code; the sign-in page shows a
  localized "temporarily unavailable, try again shortly" (EN/TR/DE) and the real cause
  is logged server-side only.
- The sign-in page renders only allow-listed messages, so a message that slips past the
  server guard still cannot leak internals.
- A regression test that exercises the guard through NextAuth's own provider merge —
  the reason the previous fix passed review while doing nothing.
- On the server: the unit restarts itself after an OOM kill, a watchdog covers the cases
  systemd will not, and somebody gets an email when the database is down and could not be
  recovered.
- The restart is first aid, not the cure: the recovery should record the kernel's OOM
  evidence and say plainly when it is flapping, so the memory problem gets fixed
  (swap, `innodb_buffer_pool_size`) instead of being papered over.

mersahin:
## Summary

While MySQL was down, the sign-in form printed the driver's own words
(``Invalid `prisma.user.findUnique()` invocation: Can't reach database server at `localhost:3306` ``).
Two independent problems sat behind that one screenshot: the guard meant to prevent it
never ran, and nothing on the server brought the OOM-killed database back.

Closes #2145

## Changes

**The message can no longer reach the browser**

- **Root cause of the leak.** `CredentialsProvider({…})` keeps the caller's real
  `authorize` in its `options` bag (the top-level one is a stub returning `null`), and
  NextAuth's `parseProviders` merges that bag back **over** the provider — reinstating the
  unwrapped function. `guardProviders` (added in #1150) was wrapping a stub that is never
  called. It now wraps both, so the guard survives that merge whichever one wins.
- **New `SERVICE_UNAVAILABLE` code.** An unreachable database (`P1001`, `P2024`,
  `PrismaClientInitializationError`, …) is distinguished from a generic fault, because the
  honest advice differs: "try again in a few moments" is actually true. Rendered as a
  localized message in EN/TR/DE; the Prisma stack goes to the server log only.
- **Allow-list on the client too.** The sign-in page renders only messages in
  `INTENTIONAL_AUTH_ERRORS` (now shared by both sides); anything else — including a
  NextAuth-generated code or a fault thrown outside the guard — is generic by construction.
- The guard moved to `src/lib/authGuard.ts` so it can be unit tested without pulling Prisma
  into the test process.

**The database restarts itself**

`infra/mysql-watchdog.sh`, in three layers, because each one alone has a gap:

| Layer | Covers | Where |
|---|---|---|
| systemd drop-in (`Restart=always`, no start-rate limit, `OOMScoreAdjust=-500`) | brings mysqld back within ~10s of a kill | written by `--install` |
| on-box timer, every minute | a unit left `inactive`, a start that failed on a transient, a `stop` nobody meant to leave | `internship-mysql-watchdog.timer` |
| `mysql-watchdog.yml`, every 10 min, **GitHub-hosted** | the box being unresponsive; re-installs the timer; the only layer that **emails** | new workflow |

Two decisions worth calling out:

- **Liveness = the server answered the protocol**, not "we could log in". A `mysqladmin ping`
  answered with `Access denied` is a healthy server refusing an anonymous client; reading
  that as down would restart a perfectly good database every minute. No credentials are
  read or stored anywhere in the script.
- **The restart is first aid.** Every recovery logs the kernel's OOM lines and a memory
  snapshot, and three restarts in an hour are reported as *flapping*, not health — the box
  needs headroom (swap, `innodb_buffer_pool_size`), and the alert says so.

Docs: `infra/README.md` § 6. Retrospective in `docs/agent-experience.md`.

## Verification

- [x] New unit spec `e2e/auth-error-guard.unit.spec.ts` (5 tests) — and it **fails** with
      the `options` wrapping removed, i.e. it actually catches the bug that shipped
- [x] New `infra/test/mysql-watchdog.test.sh` (28 checks, wired into `ci.yml`)
- [x] `e2e/auth.spec.ts`, `e2e/two-factor.spec.ts` (which assert the pass-through error
      strings) and `e2e/auth-corrupt-json-column.spec.ts` pass against a local MariaDB
- [x] Tested manually end to end against a real MariaDB: sign-in works; stopping the
      server yields `SERVICE_UNAVAILABLE` at `/api/auth/callback/credentials` and
      "Giriş şu anda geçici olarak kullanılamıyor…" / "Sign-in is temporarily unavailable…"
      / "Die Anmeldung ist vorübergehend nicht verfügbar…" on the form, with the Prisma
      stack only in the server log; the watchdog detected the stopped server, restarted it,
      and stayed silent on the next run

One bug the stubs could not have caught, found by running the script against a real
server: `mysqladmin` accepts `--no-defaults` **only as the first argument** and otherwise
exits 2 with `unknown option`, which made the probe report a healthy server as down — it
would have restarted production's database once a minute. Fixed, and the test stub is now
as picky as the real client. The probe also distinguishes "could not ask" from "answered
no", instead of treating both as down.

## Contributor terms

https://claude.ai/code/session_017uwfZq9B3XsneUWk1UpgeH

---
_Generated by [Claude Code](https://claude.ai/code/session_017uwfZq9B3XsneUWk1UpgeH)_


mersahin:
**Status:** every check on `f61ac14` is green — CI (lint · typecheck · build, incl. the new `infra/test/mysql-watchdog.test.sh`, 28/28), Playwright smoke, CodeQL, Analyze, Security audit, and the `topic-pr2146` image build.

**The preview env is not up yet, and it isn't this PR.** `Deploy topic environment` has sat *queued* on the self-hosted runner since 05:43 UTC — 70+ minutes with no other job occupying it. It is not specific to this branch: #2148's topic deploy has been queued the same way since 06:38. `runner-watchdog` reported the runner service **active** at 06:04, and a manual re-run changed nothing, which fits a runner whose service is up but no longer picking work off the queue (it only restarts an *inactive* one).

The box itself is healthy from outside: `crm.ersah.in` and `crm-preview.ersah.in` both answer `/api/health` with 200. So this is the runner process, not the host being down.

Nothing to fix in this diff. `https://crm-pr2146.ersah.in` will come up on its own once the runner takes the job; if it doesn't, the runner service on the box needs a restart by hand. I'll keep watching and report when the deploy lands.

---

PR #2200, opened 2026-09-06
fix: give a mentor application's admin note its own column (#1806)

### LINKED ISSUE #1806: 🐛 🌱 [Mentor] Task · Saving an admin note overwrites the rejection reason on a mentor application
## 🎯 Goal
Two different pieces of text are being written into one column. An admin saving a private note on a mentor application silently destroys the recorded reason that application was rejected — the exact record a later reviewer, or a complaining applicant, needs. Give the note its own column and stop the collision.

## 🔍 Current state
- `src/app/api/mentor-applications/[id]/route.ts:78` — the `note` action:
  ```ts
  await prisma.mentorApplication.update({ where: { id }, data: { rejectReason: note ?? '' } });
  ```
- `src/app/api/mentor-applications/[id]/route.ts:121` — the `reject` action writes the decision reason into the same column:
  ```ts
  data: { status: 'REJECTED', decidedAt: new Date(), decidedById: session.user.id, rejectReason: reason }
  ```
- The column's own comment says what it is for and does not mention notes: `prisma/schema.prisma:1257-1258` — *"Set when rejecting; explains the decision to the reviewing admin later."*
- Consequence chain: reject an application with a reason → later save a note on it → the reason is gone, replaced by the note, with no audit trail beyond a generic `mentor_application.note_saved` activity row (`:79-86`). Passing an empty note wipes the column entirely (`note ?? ''`).
- The reject path also requires a non-empty reason (`:66-68`), so the data loss is guaranteed to destroy something that was deliberately collected.

## 🛠 What to build
1. Add `adminNote String? @db.Text` to `MentorApplication` in `prisma/schema.prisma` (additive, nullable — safe under `db push`), with a comment saying explicitly that it is **private to admins and never sent to the applicant**, unlike `rejectReason`.
2. Point the `note` action at `adminNote` and stop coercing to `''` — an empty note should store `null`, not blank the field.
3. Return and render both values wherever an application is shown: the list/detail route (`GET /api/mentor-applications`) and the admin review UI, with the note visually marked as internal.
4. One-off backfill consideration, stated in the PR description rather than automated: existing `rejectReason` values on **non-`REJECTED`** applications are almost certainly notes that overwrote nothing, and can be copied to `adminNote`. Do not touch `rejectReason` on `REJECTED` rows — that value may be either the real reason or a note that already destroyed it, and guessing would make it worse. Add the copy as an idempotent script under `prisma/` following the existing backfill convention.
5. Keep the activity log entry, and add the application id to its `detail` so a future incident review can follow it.

## ✅ Acceptance criteria
- [ ] Saving a note on a `REJECTED` application leaves `rejectReason` byte-identical
- [ ] Saving an empty note stores `null`, not `''`
- [ ] Both the note and the rejection reason are visible in the admin UI, the note marked internal
- [ ] The applicant-facing rejection e-mail is unaffected and never contains the note
- [ ] The backfill script is idempotent and touches only non-`REJECTED` rows
- [ ] `npm run build` passes
- [ ] `npx prisma validate` passes
- [ ] an e2e spec rejects an application with a reason, saves a note, and asserts the reason survived
- [ ] a release fragment `releases/unreleased/mentor-application-admin-note.json` is added (bump: `patch`)

## 🇹🇷 Türkçe özet
İki farklı metin tek bir sütuna yazılıyor. Yönetici bir mentor başvurusuna özel bir not kaydettiğinde, o başvurunun neden reddedildiği bilgisi sessizce siliniyor — yani sonradan bakacak bir başka yöneticinin ya da itiraz eden bir başvuranın tam olarak ihtiyaç duyduğu kayıt. Nota kendi sütununu veriyoruz (`adminNote`), boş not artık `''` değil `null` yazıyor ve iki değer de yönetici ekranında ayrı ayrı, notu "dâhili" işaretiyle görünüyor. Geriye dönük düzeltme yalnızca `REJECTED` olmayan satırlara dokunuyor: reddedilmiş bir satırdaki değerin gerçek gerekçe mi yoksa onu ezmiş bir not mu olduğunu bilemeyiz, tahmin etmek durumu kötüleştirir.

## 🤖 Agent prompt
```
Bug: the "save a note" action on a mentor application overwrites the rejection reason,
because both write to MentorApplication.rejectReason.

Open first:
  src/app/api/mentor-applications/[id]/route.ts:60-130
      :78   -> data: { rejectReason: note ?? '' }        <-- the defect
      :121  -> data: { ..., rejectReason: reason }       <-- the legitimate writer
  prisma/schema.prisma:1236-1268   (MentorApplication; the column comment at :1257-1258)
  the admin mentor-application review UI under src/app/admin/
  prisma/  (the existing ad-hoc backfill scripts — follow their convention)

Do:
1. Add adminNote String? @db.Text to MentorApplication. Comment it as private-to-admins,
   never sent to the applicant.
2. Point the note action at adminNote; store null for an empty note instead of ''.
3. Show both values in the admin UI, the note marked internal.
4. Idempotent backfill script copying rejectReason -> adminNote for NON-REJECTED rows only.
   Do not touch REJECTED rows: that value may be the real reason or a note that already
   destroyed it, and guessing makes it worse. Explain this in the PR description.
5. Keep the activity entry; add the application id to its detail.

Constraints:
- No migrations folder (#1515) — one additive nullable column, nothing else.
- The rejection e-mail must never include adminNote.
- EN/TR/DE for the new "internal note" label.
- Add releases/unreleased/mentor-application-admin-note.json with bump: patch.

Done when: rejecting with a reason and then saving a note leaves the reason untouched, and an
e2e spec proves it.
```

**Zorluk** · S · **Tahmini** · 0.5 day · **Bağımlılık** · yok

mersahin:
## Summary

Saving a review note on a mentor application destroyed the rejection reason. Both actions wrote the same column: the note action did `data: { rejectReason: note ?? '' }` while the reject action wrote the real reason to `rejectReason`. So an admin who rejected an application with a documented reason and then jotted a note next to it overwrote that reason with the note — silently, with no way back, because the old code did a plain `UPDATE`. The detail page then completed the confusion by loading the note box *from* `rejectReason`, so a rejection reason was displayed as if it were the note and the next save laundered it into one.

The two are different things and now live in different columns. `MentorApplication` gains `adminNote String? @db.Text` — nullable and additive, so `db push` is safe — documented in the schema as private to admins and never sent to the applicant, unlike `rejectReason`, which at least documents a decision that was communicated.

Closes #1806

## Changes

- **`prisma/schema.prisma`** — new `adminNote String? @db.Text` on `MentorApplication`, with the privacy boundary between it and `rejectReason` written down. `rejectReason`'s comment now says who is allowed to write it. No SQL migration (this repo uses `db push`).
- **`src/app/api/mentor-applications/[id]/route.ts`** — the `note` action writes `adminNote`, and stores an emptied note as `NULL` instead of `''`, so "no note" is one state rather than two. The activity entry (`mentor_application.note_saved`) stays and now carries the application id in its `detail`; the note text itself is deliberately *not* logged — it is private commentary and the activity log has a wider audience than the application page.
- **List + detail routes** return both columns (neither uses a `select`, so this is already true — now stated in a comment, since a future `select` is exactly how it would quietly stop being true).
- **Admin review UI** — the note box reads `adminNote`, its card is marked with an **Internal** chip, and a rejected application shows its recorded reason in its own read-only card (`mentor-application-recorded-reject-reason`). The list rows show both too: the reason under a `REJECTED` row, and the note behind the same Internal chip. New EN/TR/DE keys `rejectReasonRecorded`, `rejectReasonRecordedHint`, `internalBadge`.
- **`src/lib/mentorApplicationRejectReason.ts`** — the one predicate deciding whether a `REJECTED` row's `rejectReason` may be *labelled* as the reason; see below.
- **`prisma/backfill-mentor-application-admin-note.mjs`** (idempotent, wired into `infra/deploy-prod.sh`) — see below.
- **`scripts/test/mentor-application-note-privacy.test.mjs`** + a `ci.yml` step — the hard constraint, enforced.
- **`scripts/sanitize-db.mjs`** drops `adminNote` alongside `rejectReason`; it is the same class of free text about a real person.
- **`e2e/mentor-application-review.spec.ts`** — a new `@smoke` case for the regression.

## Why the backfill skips REJECTED rows — and what that costs

The script copies `rejectReason` → `adminNote` for **non-rejected rows only** (`PENDING` / `UNDER_REVIEW` / `APPROVED`), and only where `adminNote` is still `NULL`, which is what makes it idempotent.

On a **non-rejected** row the value is unambiguous. The reject action is the only writer of a genuine reason, and it sets `status = REJECTED` in the same write — so a value sitting on a row that is not `REJECTED` can only have come from the note action. It *is* a note, and moving it into the field it belongs in loses nothing.

On a **REJECTED** row the value is ambiguous, and no inspection of the row resolves it: it is either the real rejection reason, or a note that was saved afterwards and already destroyed that reason. Nothing distinguishes the two cases, so any rule I could write would be wrong roughly half the time — and each direction of error is its own damage. Treating a reason as a note strips the documentation from a decision that was communicated to a person; treating a note as a reason fabricates a justification that nobody wrote. The already-overwritten reasons are not recoverable — the old code overwrote them in place.

**Be precise about what declining to guess actually buys, because an earlier revision of this description overstated it.** It claimed that leaving the value put *"keeps the surviving text visible under the label it was last written with"*. That is false for exactly the damaged rows: such a row reads `status=REJECTED`, `rejectReason=<the note>`, `adminNote=NULL`, and the review UI renders a `REJECTED` row's `rejectReason` as the recorded decision. So not moving the data would still have ended with a private note displayed where a rejection reason belongs — the fabrication merely moves from the data into the label, and the CRM would assert that, say, *"Applicant's brother works here — check conflict of interest"* was why somebody was turned down.

The script cannot fix that (it cannot tell the rows apart either), so **the UI carries the ambiguity instead**: a `REJECTED` row decided before the split — or with no `decidedAt` at all — is shown under **"Recorded text (may be a review note)"** with a caveat naming both possibilities, rather than under the affirmative "Recorded rejection reason" heading and its hint. Rows decided after the split are unaffected and keep the confident label. The rule lives in `src/lib/mentorApplicationRejectReason.ts`; the cutoff is deliberately rounded up by a day, because over-flagging a handful of correct rows costs one hedging sentence while under-flagging restates the defect.

The script does not clear `rejectReason` on the rows it copies either. A copy is reversible; a delete based on an inference is not. The UI simply stops showing a "rejection reason" for a non-rejected status, which is where a stale copy would otherwise be misread.

## The rejection e-mail cannot contain the note

`sendMentorApplicationRejectedEmail` takes `{ to, fullName, locale, orgId }` and nothing else. Its body is `M.rejected.heading` / `M.rejected.body` from the dictionary plus the recipient's own name — no application text of any kind reaches it, which is also why the applicant only ever receives a general decline. And the PATCH handler's `findUnique` — the object every e-mail branch reads from — selects `id, status, fullName, email, locale, orgId, capacity, expertise`: neither free-text column is even loaded.

That is true of today's code, and nothing in the type system keeps it true: `adminNote` is just another string on the row, and a future "include the note so the reply has context" change would compile, pass every runtime test, and be discovered by the person who received it. So it is asserted at the file level in `scripts/test/mentor-application-note-privacy.test.mjs` (`npm run test:note-privacy`, wired into `ci.yml`):

- **no reference to `adminNote` in any file that composes an e-mail** — the scan walks `src/`, keeps every file containing a `sendEmail(` call (12 today, not just `emailService.ts`, which was the original blind spot: ~10 routes and libs build their own HTML) and asserts none mentions the column, so a new mail-composing route is covered automatically;
- **no reason/note text in the rejection sender's body** — extracted by walking the parameter list by paren depth and then brace-matching the body. A naive `indexOf('\n}')` matches the multi-line destructure's `}: {` at column 0 and yields 95 characters of parameter names, which is why the check also asserts a length floor and that the slice contains `sendEmail({` and `html:`;
- **no free-text column in the PATCH handler's select.**

## Verification

Static only — this container has no database and no Docker daemon, so the app and the Playwright suite could not be run here. The new e2e case is written but unexecuted; CI runs it.

- [x] `npm run lint` clean (two pre-existing `react-hooks/exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`, untouched by this PR)
- [x] `npx prisma format && npx prisma validate && npx prisma generate` clean
- [x] `npm run check:i18n` — 4156 keys × 3 locales
- [x] `npm run check:release-fragments` — valid, ships as a patch
- [x] `npm run check:query-scalars`, `npm run test:note-privacy` (3/3) green — and both privacy assertions mutation-verified: referencing `adminNote` from a `sendEmail(` caller and interpolating `rejectReason` into the rejection body each fail the suite
- [x] Added an e2e test: rejects with a reason, saves a note, asserts the reason survived byte-identical, reloads to confirm the two fields never swap places, and empties the note to confirm it clears to `NULL` while the reason stays

Preview: https://pr2200.interncrm.com

## Contributor terms


mersahin:
## Review follow-ups — both findings confirmed and fixed (`680b0af`)

I reproduced both before changing anything. Neither was wrong; I rejected nothing.

### A — the UI re-labelled a note as a rejection reason

Confirmed, and the reviewer's framing is right: the backfill refusing to guess on `REJECTED` rows is **not** the same as leaving those rows labelled correctly. A row whose reason was destroyed by a note ends up as `status=REJECTED`, `rejectReason=<the note>`, `adminNote=NULL` — and the new card rendered exactly that under **"Recorded rejection reason"** with *"Kept for the record — the applicant only received a general decline message"*, while the queue row prefixed it **"Rejection reason:"**. So for `"Applicant's brother works here — check conflict of interest"` the CRM would have stated, on its own authority, that this was why the person was turned down. Before this PR the same text at least appeared under *"Review note (admin only)"* — the PR made that row strictly worse, which is not acceptable in a data-integrity change.

I did (a), (b) and (c):

- **`src/lib/mentorApplicationRejectReason.ts`** (new) — one predicate for the trust boundary, so the two pages cannot drift. A `REJECTED` row decided before the split, **or with no `decidedAt` at all**, is ambiguous. The cutoff is a date because that is the only discriminator that exists, and it is **rounded up to the day after this ships on purpose**: over-flagging a few correct rows costs one hedging sentence for under a day, whereas under-flagging restates the defect. That trade-off is written into the file.
- **Detail card + queue row** — ambiguous rows are titled **"Recorded text (may be a review note)"** and carry a caveat naming both possibilities *("…either the rejection reason or a note that was saved afterwards and replaced it — the CRM cannot tell which")*. The affirmative title and hint now render only for rows where they are true. Unambiguous rows look exactly as they did. New EN/TR/DE keys `rejectReasonRecordedUnverified{,Hint}`.
- **(c), done regardless** — the backfill's header comment claimed that leaving the value put *"keeps the surviving text visible under the label it was last written with"*. That is false for precisely the damaged rows: it was last written as a note and was about to be displayed as a decision. Replaced with what actually happens, plus a pointer to where the ambiguity is now carried. The PR description above is stale on the same sentence — same correction applies.

Both routes use no `select`, so `decidedAt` already reaches both pages; no API change was needed.

### B — the privacy test checked nothing

Confirmed empirically before touching it. `emailService.indexOf('\n}', start)` matches the multi-line parameter destructure's `}: {` at column 0:

```
start 60209 end 60304 len 95
export async function sendMentorApplicationRejectedEmail({
  to,
  fullName,
  locale,
  orgId,
```

**95 characters of parameter names.** It never reached `const brand = await emailBrand(orgId)`, let alone the HTML — so `['adminNote','rejectReason','reason','note']` were all asserted against a slice that could not contain them. The CI step named *"no e-mail template can read a private admin note (#1806)"* was reporting a guarantee it had not checked.

Both halves fixed:

- **Real body extraction** — walks the parameter list by paren depth first (so the destructure and the type literal cannot be mistaken for the body), then brace-matches the body. Plus the assertion that would have caught the original bug: `body.length > 400`, and an anchor check that the slice actually contains `sendEmail({` and `html:`. A silent re-shrink now fails instead of passing.
- **Scan follows the callers, not one file** — the `adminNote` grep now walks `src/`, keeps every file containing `sendEmail(` (12 today: the announcements route, `mentor/email`, `outcomeComms.server.ts`, `newsletterDispatch.ts`, …) and asserts none references `adminNote`. A future "reply to the applicant with context" route is covered the moment it calls `sendEmail(`, with a floor on the discovered count so a rename of the helper fails loudly rather than emptying the set.

**Mutation-verified**, since a test whose whole failure mode was passing vacuously deserves it:

| mutation | result |
|---|---|
| `adminNote` referenced in `src/lib/outcomeComms.server.ts` | `not ok 1` ✅ |
| `${esc(rejectReason ?? '')}` interpolated into the rejection body | `not ok 2` ✅ |

Both reverted; `git diff` on those files is empty.

### Checks

`npx tsc --noEmit` clean · `npm run lint` clean (only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`) · `npm run check:i18n` 4156 keys × 3 locales · `npm run check:release-fragments` valid · `npm run check:query-scalars` clean · `npm run test:note-privacy` 3/3. Schema untouched, so no `prisma validate` run. No DB and no browser here, so the e2e case is still unexecuted by me; its existing assertions target `data-testid`s that did not change, so the hedged label does not affect it.

The release fragment prose was updated to describe the hedged label and the strengthened test, since both change what the fragment claims shipped.

---

Issue #1733, opened 2026-08-30
🔀 [Billing] Task · One entitled(orgId, feature) resolver — CompanyEntitlement becomes a shim


mersahin:
## 🎯 Goal
Collapse the three gating mechanisms into one function so that from this point on there is exactly one place to ask whether a tenant may use a feature — and so that the enforcement work in Story B has a single call to make instead of three variants.

## 🔍 Current state
- `src/lib/entitlements.ts:16-23` — `hasFeature(companyId, feature)` does one indexed lookup on `CompanyEntitlement`; row presence means on. Null company → false.
- `src/lib/entitlements.ts:27-36` / `:39-49` — `getCompanyFeatures()` and `setCompanyFeature()` back the admin toggle grid (`src/components/admin/CompanyEntitlements.tsx`, `src/app/api/admin/companies/[id]/entitlements/route.ts`).
- Only four call sites exist: `src/app/api/company/talent-pool/route.ts:20` (`TALENT_POOL_SEARCH`), `:61` (`EARLY_ACCESS`), `src/app/api/company/candidates/[id]/route.ts:59` (`VERIFIED_CANDIDATE_CARD`), `src/app/company/layout.tsx:28` (nav visibility).
- The fifth consumer is a raw query, not a call site: `src/services/emailService.ts:2661-2680` selects companies via `entitlements.some.feature = 'COMPANY_NEED_MATCH_ALERTS'`.
- `Company.orgId` is nullable (`prisma/schema.prisma:1046-1047`), so a company can exist without a tenant — the resolver has to decide what that means rather than crash.
- A separate, parallel gate: `src/lib/settings.ts:74-77` `getSetting('premiumAnalytics')`, read at four places (moved in a sibling task).

## 🛠 What to build
1. In `src/lib/entitlements.ts`, add `entitled(orgId: string | null | undefined, feature: PremiumFeature): Promise<boolean>` resolving in this order:
   1. no `orgId` → **false** (fail closed for paid features; the free core never reaches this function),
   2. active `Subscription` for the org → `planFeatures(planKey)` from `src/lib/plans.ts` contains the feature → true,
   3. an `OrgEntitlement` row for `(orgId, feature)` with `expiresAt` null or in the future → true,
   4. otherwise false.
[...]

Issue #1731, opened 2026-08-30
🗃 [Billing] Task · Subscription + OrgEntitlement schema and the plan→feature matrix in src/lib/plans.ts


mersahin:
## 🎯 Goal
Give the Organization somewhere to record what it is paying for. Right now a tenant's commercial state is one enum column (`Organization.plan`) with no customer, no period, no trial and no way to grant a single feature outside the plan — which means every non-standard deal has to be modelled as "put them on ENTERPRISE and hope".

> 🔗 Absorbs #1737 (closed as duplicate) — merged in its `educationDiscount` flag, the lazy `getOrCreateSubscription(orgId)` accessor and its legacy-plan seeding map. There is **one** subscription table for a tenant and it is `Subscription`; `OrgSubscription` is not created.

## 🔍 Current state
- `prisma/schema.prisma:61-107` — `Organization` carries `plan OrgPlan @default(FREE)` at `:66` and nothing else commercial: no `stripeCustomerId`, no period, no trial, no seat count.
- `prisma/schema.prisma:26-30` — `enum OrgPlan { FREE PRO ENTERPRISE }`, three tiers that do not match the five we intend to sell.
- `src/lib/orgPlans.ts:13-26` — `OrgPlanLimits { maxUsers, maxActiveRelations, maxProjects }`; the header comment at `:7-11` states the limits are advisory.
- `src/lib/entitlementsCatalog.ts:5-15` — nine feature keys, no plan association at all; a plan does not know which features it contains.
- `prisma/schema.prisma:1077-1092` — `CompanyEntitlement` is the only entitlement storage and is keyed on `companyId`.
- `src/lib/orgContext.ts:43-55` — `TENANT_MODELS` currently lists 11 models; anything new that holds tenant data must be added here.

## 🛠 What to build
1. `prisma/schema.prisma` — add to `Organization`: `stripeCustomerId String? @unique`, `billingEmail String?`, `vatId String?`, `billingCountry String?` (ISO-3166 alpha-2). Purely additive.
[...]

Issue #1535, opened 2026-08-30
🐛 [Trust] Task · Gate /api/admin/organizations behind a super-admin capability on User


mersahin:
## 🎯 Goal
Any customer's admin can currently overwrite any other customer's SSO signing certificate and IdP entry point through `PATCH /api/admin/organizations`, because the only check is `role === 'ADMIN'`. Introduce the missing distinction between "runs this deployment" and "runs this tenant", and gate the organisation-management surface on it. This is the single change that turns "we have an isolation engine" from a claim into something a security reviewer can verify.

## 🔍 Current state
- `src/app/api/admin/organizations/route.ts:19` — `GET`: `if (!session || session.user.role !== 'ADMIN') return 401`, then `prisma.organization.findMany()` with no filter, returning every tenant's plan, branding, usage counts and SSO state.
- `src/app/api/admin/organizations/route.ts:95` — `POST`: same check, creates an organisation.
- `src/app/api/admin/organizations/route.ts:147` — `PATCH`: same check. The target is `parsed.data.id` from the **request body** (`patchSchema`, `:131`), and the writable fields include `ssoEnabled`, `ssoProvider`, `ssoIssuer`, `ssoEntryPoint`, `ssoCertificate` (`:184-199`). The org is loaded by that id at `:174` with no ownership test.
- `src/app/api/admin/organizations/[id]/pipeline-stages/route.ts` is the only other route under this path — check it for the same pattern.
- `grep -rn "SUPER_ADMIN" src/ prisma/` → no results. `prisma/schema.prisma` `User` has `role`, `orgId`, `isActive` (`:278`) and no capability column.
- `src/lib/orgScope.ts:29-33` — `resolveOrgId(session)` already gives the caller's own org.
- The session shape is assembled in `src/lib/auth.ts:459-475` (`session.user.orgId` is already carried).
- e2e: `e2e/admin-organizations.spec.ts`.

## 🛠 What to build
1. `prisma/schema.prisma` — add `isSuperAdmin Boolean @default(false)` to `User`. Additive with a default, so `prisma db push` resolves it non-destructively. Run `prisma format && prisma validate && prisma generate`.
2. New `src/lib/superAdmin.ts` with one server-side helper — e.g. `requireSuperAdmin(session)` returning the session or null — that reads the flag from the database (not from the JWT: a freshly-revoked capability must take effect on the next request, the same reasoning as the `sessionsValidFrom` lookup at `src/lib/auth.ts:449-455`).
[...]

Issue #1736, opened 2026-08-30
🔒 [Billing] Story · Make the decorative feature keys real — five premium toggles that enforce nothing


mersahin:
## 🎯 Goal
Half the premium menu we would print on a pricing page is fiction. `AI_PACKAGE`, `ADVANCED_ANALYTICS`, `REPORT_EXPORT`, `WHITE_LABEL` and `SSO_SAML` are catalogued, translated into three languages and rendered as admin toggles — and `grep` finds **zero** enforcement call sites for any of them. A packaging model nobody can trust is worse than no packaging: it makes every price conversation a code audit, and the first customer who notices is the first refund. This story gives every catalogued key at least one server-side gate, and adds the one key the integration tier needs and does not have.

## 🔍 Current state
- `src/lib/entitlementsCatalog.ts:5-15` — nine keys. The only `hasFeature()` call sites in the entire repo are `src/app/api/company/talent-pool/route.ts:20` and `:61`, `src/app/api/company/candidates/[id]/route.ts:59`, and `src/app/company/layout.tsx:28` — covering `TALENT_POOL_SEARCH`, `EARLY_ACCESS` and `VERIFIED_CANDIDATE_CARD`. `COMPANY_NEED_MATCH_ALERTS` is enforced via a raw query at `src/services/emailService.ts:2661-2680`.
- Analytics is gated by a completely different mechanism — the global `premiumAnalytics` Setting (`src/lib/settings.ts:25`), read at `src/app/api/admin/analytics/cohorts/route.ts:20`, `.../sources/route.ts:18`, `.../benchmark/route.ts:27` and `src/services/emailService.ts:2879`.
- White-label and SSO are gated, when at all, by `Organization.plan` or by plain ADMIN role — a third mechanism. `src/app/api/admin/organizations/route.ts:19` is bare `role === 'ADMIN'`, and the per-tenant pipeline-stage editor rejects only the `FREE` plan.
- AI has a quota but no entitlement: `src/lib/aiGate.ts:24-30` counts `AiUsage` rows globally against one `Setting`.
- Exports are produced **client-side** — `src/lib/excel.ts` is a browser-only SheetJS wrapper, called from `src/app/admin/candidates/page.tsx:125`, `src/app/admin/analytics/page.tsx:203/:217` and `src/components/admin/ContributorTermsReport.tsx`. There is nothing server-side to gate, which is exactly why `REPORT_EXPORT` has no call site.
[...]

PR #2203, opened 2026-09-06
feat: gate WHITE_LABEL and SSO_SAML at the write boundary (#1742)

### LINKED ISSUE #1742: 🌱 🎨 [Admin] Task · WHITE_LABEL and SSO_SAML enforced at the write boundary, not by role
## 🎯 Goal
White-label branding and SAML SSO are the two headline Program Plus / Enterprise features, and both are currently reachable by any tenant that can reach the admin screen — the entitlement keys for them exist and enforce nothing. This is a small, well-bounded task: put the gate on the **write**, so an unentitled tenant cannot store branding or an IdP config in the first place, and the read path needs no change at all.

## 🔍 Current state
- `src/lib/entitlementsCatalog.ts:13-14` declares `WHITE_LABEL` and `SSO_SAML`. Neither has a `hasFeature()` call site anywhere.
- Branding is written by `PATCH /api/admin/organizations` — `src/app/api/admin/organizations/route.ts:145-220`, with hex/SSRF validation at `:157-169`. The only authorization is `session.user.role !== 'ADMIN'` at `:147`.
- SSO config is written by the same handler (`:181-202`, effective-config merge before validating; the change is logged at `:207-218`). Validation helpers live in `src/lib/sso.ts` (`validateSsoConfig`, `isSsoActive`).
- Read paths that must keep working untouched: `src/lib/branding.ts:26-84` (`resolveBranding` with product defaults), `src/lib/orgBranding.ts:7-13`, `src/components/BrandWordmark.tsx:11-27`, the 25 branded e-mail templates via `src/services/emailService.ts:42-63`, and the certificate PDF at `src/app/api/mentorship/[id]/certificate/route.ts:79-88`.
- `Organization` stores the branding and SSO columns at `prisma/schema.prisma:74-90`.

## 🛠 What to build
1. In the `PATCH` handler of `src/app/api/admin/organizations/route.ts`, before applying any of `brandName` / `brandLogoUrl` / `brandColor` / `supportEmail`, require `await entitled(targetOrgId, 'WHITE_LABEL')`. Reject with the standard `403 { code: 'feature_locked', feature: 'WHITE_LABEL' }`.
2. Same for the SSO block (`ssoEnabled`, `ssoProvider`, `ssoIssuer`, `ssoEntryPoint`, `ssoCertificate`) against `SSO_SAML`.
3. **Downgrade behaviour is explicit, not accidental**: losing the entitlement does *not* wipe stored branding or SSO config. `isSsoActive()` in `src/lib/sso.ts` additionally requires the entitlement, so an unentitled tenant's IdP simply stops being accepted at sign-in while the config survives an upgrade back. Document that decision in the function's header comment.
4. The admin screen (`src/app/admin/organizations/page.tsx:101-138` branding editor, `:140-180` SSO editor) renders a locked state with an upgrade link. This is cosmetic — the server check is the gate.
5. New locked-state copy gets EN/TR/DE keys.

## ✅ Acceptance criteria
- [ ] An admin of an unentitled org gets 403 when PATCHing branding — verified by calling the API directly with the UI bypassed.
- [ ] Same for the SSO block.
- [ ] Losing the entitlement leaves stored branding rows intact and only stops SSO sign-in; re-granting it restores both with no re-entry.
- [ ] Every branding **read** path still works: wordmark, e-mail templates, certificate PDF.
- [ ] `e2e/sso-saml-mapping.spec.ts` and `e2e/admin-organizations.spec.ts` pass (grant the entitlement in the fixture where needed).
- [ ] `npm run build` passes
- [ ] an e2e spec covers the 403
- [ ] a release fragment `releases/unreleased/<kebab-slug>.json` is added

## 🇹🇷 Türkçe özet
White-label ve SAML SSO, üst planların iki vitrin özelliği — ama ikisinin de entitlement anahtarı hiçbir şeyi kapatmıyor; admin ekranına ulaşan her kiracı ikisini de kullanabiliyor. Bu küçük görev kapıyı **yazma** anına koyuyor: hakkı olmayan kiracı marka bilgisini veya IdP ayarını hiç kaydedemiyor. Plan düşerse kayıtlı veriler silinmiyor — sadece SSO girişi kabul edilmiyor; müşteri geri yükseldiğinde her şey yerinde duruyor. Okuma yolları (e-posta şablonları, sertifika PDF'i, uygulama logosu) hiç değişmiyor.

## 🤖 Agent prompt
```
Objective: gate white-label branding writes on WHITE_LABEL and SSO config writes on
SSO_SAML, at the API write boundary.

Open first:
  src/app/api/admin/organizations/route.ts  (PATCH :145-220; branding validation :157-169;
                                             SSO merge/validate :181-202; logging :207-218;
                                             note the bare role check at :147)
  src/lib/sso.ts (isSsoConfigComplete / isSsoActive / validateSsoConfig)
  src/lib/entitlements.ts (entitled), src/lib/entitlementsCatalog.ts (:13-14)
  src/app/admin/organizations/page.tsx (:101-138 branding editor, :140-180 SSO editor)
  Read paths that must NOT change: src/lib/branding.ts, src/lib/orgBranding.ts,
  src/components/BrandWordmark.tsx, src/services/emailService.ts:42-63

Constraints:
  - Server-side gate on the WRITE. Rendering a locked card in the UI is not the gate.
  - Do NOT delete stored branding or SSO config when the entitlement is absent - make
    isSsoActive() require it instead, and say why in the comment.
  - Never touch the read paths; a tenant that already has branding keeps seeing it.
  - New copy needs EN + TR + DE keys or check:i18n fails.
  - This route currently authorizes on `role === 'ADMIN'` only, which is the cross-tenant
    escalation #1535 is fixing separately. Do not make it worse; do not try to fix it here.

Done when: direct PATCH from an unentitled org 403s for both blocks, existing SSO and
org-admin e2e specs green, build + check:i18n green, release fragment, PR open.
```

**Zorluk** · S · **Tahmini** · 0.5-1 day · **Bağımlılık** · #1733

mersahin:
## Summary

White-label branding and SAML SSO are the two headline premium features, and both were reachable by any tenant that could open the admin screen: `WHITE_LABEL` and `SSO_SAML` existed in `src/lib/entitlementsCatalog.ts` with **zero call sites**. This puts the gate on the **write** — `PATCH /api/admin/organizations` is the only place either config is ever created, so an unentitled tenant can no longer store branding or an IdP config in the first place, and no read path had to change.

The branding block (`brandName`, `brandLogoUrl`, `brandColor`, `supportEmail`) now requires `WHITE_LABEL`; the SSO block (`ssoEnabled`, `ssoProvider`, `ssoIssuer`, `ssoEntryPoint`, `ssoCertificate`) requires `SSO_SAML`. Either refusal is `403 { code: 'feature_locked', feature, requiredPlan }`. The check runs **before** the hex/SSRF field validation, so the entitlement answer never depends on whether the payload happened to be well-formed, and **after** the existing cross-tenant ownership check, so the 403-before-404 ordering that stops a foreign admin probing org ids is unchanged.

**Three rules make the gate mean something** (all three came out of review — see the second commit):

1. **The gate cannot be bought by the actor it gates.** The entitlement is read from `Organization.plan`, and this endpoint used to let a plain tenant ADMIN write that field on their own org — so `{plan: 'ENTERPRISE', …SAML}` in a single request walked straight through, and the two-step version needed no bundling at all. Setting the tier is a **billing** act, not a tenant-administration one: **super admin only**, same as creating the tenant in `POST`. A no-op (the plan the org is already on) is not a change and still passes, so a client that echoes the current value is unaffected. The row's plan `<select>` is disabled for non-super-admins to match.
2. **Undoing is always allowed.** The gate stops a tenant *acquiring* a paid feature, not *removing* one it no longer pays for. Every branding read path is deliberately untouched, so refusing the delete too would mean white-label is bought once and renders in every branded e-mail and on the certificate PDF forever — with no way for anyone, super admin included, to remove it without a billing-visible plan change. A payload whose premium fields can only null columns is exempt; `ssoEnabled: false` counts as a clear, so the dead SSO switch can be switched off. The locked cards carry the matching actions rather than being a dead end.
3. **Nothing that was legal when it was written goes dark on deploy.** See the backfill below.

**Where the entitlement comes from — read this before reviewing.** #1733's `entitled(orgId, feature)` and #1731's `Subscription`/`OrgEntitlement` schema have not landed, and `src/lib/entitlements.ts` only offers `hasFeature(companyId, feature)`, which is **company**-scoped — the wrong axis for a tenant-level feature. So the entitlement is derived from the org's `Organization.plan` by a new pure helper `orgPlanHasFeature(plan, feature)` in `src/lib/orgPlans.ts`. **Both features are `ENTERPRISE`**, matching `docs/premium-model-calismasi.md`, which puts white-label in Enterprise in both places it enumerates tiers (Pro buys scale, analytics and the AI package). **This is an interim source of truth that #1733 will replace**, and the helper is deliberately the single place that has to change when it does. No schema change, no new entitlement system. The grandfathered `default` org is already `ENTERPRISE` (`prisma/backfill-organization.mjs`), so existing prod/preview tenants keep both features.

**Downgrade behaviour is explicit, not accidental.** Losing the entitlement does **not** wipe stored branding or SSO config. Instead `isSsoActive()` in `src/lib/sso.ts` additionally requires `SSO_SAML`, so an unentitled tenant's IdP stops being accepted at `/login` and `/acs` while issuer, entry point and signing certificate survive untouched — re-grant the plan and SSO resumes with nothing re-entered by the customer's IT team.

**Deploy safety.** That new requirement would have taken *existing* SSO tenants offline: `plan` defaults to `FREE`, SSO was configurable on any plan until now, and JIT-provisioned SSO users are stored with `password: '!sso-no-login'`, so there is no password to fall back to. `prisma/backfill-sso-plan.mjs` grandfathers every org whose SSO was **on and complete** to `ENTERPRISE` and runs from `infra/deploy-prod.sh` (which the preview pipeline reuses) before the container swaps. It is idempotent — a no-op on a fresh install and on every deploy after the first — and it *reports*, without touching, any org holding branding it can no longer edit.

Closes #1742

## Changes

- `src/lib/orgPlans.ts` — new `PLAN_FEATURES` map (both features `ENTERPRISE`, packaging cited in the comment), pure `orgPlanHasFeature()`, and `planIncludingFeature()` so a refusal can name the tier that sells it.
- `src/app/api/admin/organizations/route.ts` — `PATCH` refuses the branding and SSO blocks without the matching entitlement (`403 feature_locked` + `requiredPlan`); the plan field is super-admin only; a pure-clear payload is exempt from the gate; `existing` is fetched before field validation.
- `src/lib/sso.ts` — `isSsoActive()` requires `SSO_SAML`; `SsoActivation.plan` is **required**, so narrowing the login/ACS query to just the SAML columns is a compile error instead of a silent product-wide SSO outage (the guard is fail-closed on a missing plan).
- `prisma/backfill-sso-plan.mjs` (new) + `infra/deploy-prod.sh` — the idempotent grandfathering pass described above.
- `src/app/admin/organizations/page.tsx` — locked note + disabled inputs/save on both premium editors, plus a "remove stored branding" / "switch SSO off" action so a locked card is never a dead end; the plan `<select>` is read-only for tenant admins.
- `src/lib/entitlementsCatalog.ts` + `src/components/admin/CompanyEntitlements.tsx` — each catalogue entry declares its `scope`; the two `org`-scoped keys render read-only on the company screen with the reason, because a `CompanyEntitlement` row for them grants nothing and an operator who ticked one used to get an unexplained 403 from the other screen.
- `src/i18n/dictionaries.ts` — `brandingLocked`, `ssoLocked`, `featureLocked`, `brandingClear`, `ssoDisable`, `entitlements.orgScoped` in EN/TR/DE.
- `e2e/org-premium-write-gate.spec.ts` (new) — the full contract through `page.request`, UI bypassed: FREE 403s both blocks and stores nothing; a tenant admin cannot self-upgrade, alone or bundled with the config it would unlock; PRO is still locked; ENTERPRISE goes active; a downgrade keeps branding + the IdP config while `sso.active` flips false and `/api/auth/sso/<slug>/login` redirects with `sso_unavailable`; clearing works while unentitled but a blank field next to a real value does not; re-upgrade resumes with no re-entry; the locked cards' escape hatches work and the plan selector is disabled.
- `e2e/admin-organizations.spec.ts` — fixtures granted what they now need, and the `403 feature_locked` + `requiredPlan` assertion for both blocks on PRO.
- `releases/unreleased/org-premium-feature-write-gate.json` — `patch`, with EN/TR/DE notes.

**Untouched on purpose:** `src/lib/branding.ts`, `src/lib/orgBranding.ts`, `src/components/BrandWordmark.tsx`, the branded templates in `src/services/emailService.ts`, and the certificate PDF route — a tenant that already has branding keeps seeing it everywhere. The bare `role === 'ADMIN'` authorization on this route is #1535's cross-tenant escalation and is left exactly as it was.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint reports only the two pre-existing `exhaustive-deps` warnings in unrelated admin pages)
- [ ] `npm run test:e2e` passing (or CI green) — no database or Docker daemon in this container, so the specs are written but not executed here; CI runs them
- [x] Tested the change manually / added an e2e test — added `e2e/org-premium-write-gate.spec.ts` and fixed the two existing specs; `npm run check:i18n` (4157 keys × 3 locales) and `npm run check:release-fragments` both green

Preview: https://pr2203.interncrm.com

## Contributor terms


mersahin:
## Review response — all six findings addressed (111e151)

I verified each one against the code before touching anything. **All six were real; none were rejected.** The gate's shape (refuse the write, keep the stored config) is unchanged — what changed is that it can no longer be bought, no longer traps the tenant, and now matches the documented packaging.

### 1. Blocker — the gated actor could write the field the gate reads ✅ fixed

Confirmed exactly as described: `PATCH` checked only `role === 'ADMIN'` plus own-org ownership, nothing restricted `plan`, and the row's plan `<select>` sits outside the `{superAdmin && …}` block. `effectivePlan = plan ?? existing.plan` then honoured the change in the same request. The two-step version did not even need bundling.

Fixed at the field, not at the bundle: **setting the tier is now super-admin only** on this endpoint, same as creating the tenant in `POST`. It is a billing act, not a tenant-administration one, and once only a super admin can move the plan, `effectivePlan` bundling is the operator legitimately doing two things at once rather than an escalation — so I kept that behaviour instead of banning the combined payload. A refusal is audited through `logCrossTenantDenial` with the route tagged `(plan)`. A **no-op** (the plan the org is already on) is not a change and still passes, so a client that echoes the current value is not broken. The UI's plan selector is disabled for non-super-admins to match.

New assertions in `e2e/org-premium-write-gate.spec.ts`: a tenant admin's `{plan:'ENTERPRISE'}` is refused and the row is unchanged; `{plan:'ENTERPRISE', …SAML}` is refused and neither the plan nor `ssoEntryPoint` moves; the plan `<select>` renders disabled.

### 2. Major — deploy-time lockout for tenants that configured SSO under the old rules ✅ fixed

Confirmed: `plan` defaults to `FREE` (schema line 77), SSO was configurable on any plan, and `src/lib/ssoProvisioning.ts:70` provisions with `password: '!sso-no-login'` — so it is a hard lockout, not a downgrade to password auth.

`prisma/backfill-sso-plan.mjs` grandfathers to `ENTERPRISE` every org whose SSO was **on, complete and on an implemented provider** before the gate. Wired into `infra/deploy-prod.sh` §4 alongside the existing backfills, so it runs on prod *and* preview (which reuses that script) before the container swaps. Idempotent — it only touches rows that qualify and are not already `ENTERPRISE`, so it converges to a no-op. Branding is deliberately **not** grandfathered (losing the ability to edit is the intended gate, not an outage, and clearing is now exempt anyway) but the script *reports* any org in that position so the operator sees them.

The second half of your point — "their own admin cannot even switch the dead switch off" — is fixed by finding 3.

### 3. Major — the gate refused clearing as well as setting ✅ fixed

Confirmed, including the one-step variant: `{plan:'FREE', brandName:''}` computed `effectivePlan = 'FREE'` and 403'd, dropping the plan change with it.

A payload whose premium fields **can only null columns** is now exempt (`isPureClear`). `ssoEnabled: false` counts as a clear; `true` does not. Any non-empty value in the payload puts the whole block back behind the entitlement, so this is not a smuggling route — asserted: `{brandName:'', brandColor:'#ff0000'}` is still 403 and `brandColor` stays null. The locked cards in the admin screen grew the matching actions (`brand-clear`, `sso-disable`) so the UI route is open too, and the spec drives both buttons.

### 4. Major — WHITE_LABEL on PRO contradicted the documented packaging ✅ fixed

Confirmed at `docs/premium-model-calismasi.md` lines 119 and 159; Pro is scale/analytics/AI in both. `PRO: []`, white-label is `ENTERPRISE`-only, and the reasoning (with the doc cited) is now a comment on the map so the next person does not have to go looking. The EN/TR/DE `brandingLocked` copy no longer promises Pro customers an Enterprise feature — it names Enterprise and adds that stored branding stays applied and can still be removed. No doc change needed: the code now matches what is written.

### 5. Minor — optional `SsoActivation.plan` ✅ fixed

`plan: OrgPlan | string | null`, required. The narrowed-`select` optimisation you describe is now a compile error rather than a silent product-wide SSO outage, and the header comment says why the field is required so nobody "tidies" it back to optional. All three call sites pass full rows and still typecheck.

### 6. Minor — two contradictory sources of truth, and the 403 payload ✅ both fixed

`PREMIUM_FEATURES` entries now declare a `scope` (`'company'` | `'org'`). The two `org`-scoped keys render **read-only on the company entitlements modal** with a one-line reason in EN/TR/DE ("Decided by the organization's plan, not by this toggle — set it on the Organizations screen"), so an operator can no longer tick a switch that grants nothing and send the customer into an unexplained 403. The keys stay in the catalogue, so the API's `features` map shape and the "unknown key is 400" contract are untouched. `#1733` deletes the distinction when it unifies the axis.

The 403 body now carries **`requiredPlan`** (from a new `planIncludingFeature()`), so the uniform upgrade CTA #1736 specifies can name the tier without re-deriving the packaging a second time. Asserted in both specs.

### Checks

`npx tsc --noEmit`, `npm run lint` (only the two pre-existing `exhaustive-deps` warnings in unrelated admin pages), `npm run check:i18n` (4157 keys × 3 locales) and `npm run check:release-fragments` all green. No schema change, so no `prisma validate` run. The e2e specs are written but not executed here — there is no database in this container; CI runs them.

Thanks for the review — finding 1 in particular was the difference between a gate and a speed bump.

---

PR #2204, opened 2026-09-06
fix: stop GET /api/meetings leaking the mentee rsvpToken (#1548)

### LINKED ISSUE #1548: 🐛🌱 [Calendar] Task · GET /api/meetings hands the mentee's rsvpToken to every mentor and admin
## 🎯 Goal
`GET /api/meetings` returns whole `Meeting` rows with no field `select`, so `rsvpToken` — the full public credential for `/rsvp/<token>` and `/api/calendar/<token>` — is serialised into the response every ADMIN and MENTOR viewer receives. Holding that token lets you answer an invitation **as that mentee** and pull their calendar entry with no session at all. The mentee portal legitimately needs its own token; the widening to other roles looks unintended, and it is precisely the kind of thing a security review finds in ten minutes.

## 🔍 Current state
- `src/app/api/meetings/route.ts:44-80` — the `GET` handler. Role scoping is careful and explicit (`:48-51`: "Every role's scope is spelled out — an unlisted role is rejected, never silently given everything"), the `where` is correct per role (`:59-65`), and the `guests` list is **already** conditioned on role (`:64-75`, with a comment explaining that a mentee must not see third parties' addresses).
- What is missing is a field `select` on the meeting itself: `prisma.meeting.findMany({ where, include: { … }, orderBy })` at `:66-79` returns every column, `rsvpToken` included.
- The token's power: `/rsvp/<token>` is on the middleware allowlist as a public route (`src/middleware.ts:20` — "Public meeting RSVP (the unguessable token is the credential)"), and there is an ICS feed at `/api/calendar/<token>`.
- Consumers of this endpoint read `m.relation.mentee.fullName` and the meeting fields: `src/components/MeetingsManager.tsx`, `src/components/MeetingSchedulerPanel.tsx`, and the mentee portal view. Check each for a `rsvpToken` reference before removing the field — the portal genuinely needs its own.
- Meetings are created at `:82+`, where each row gets its own RSVP token.

## 🛠 What to build
1. Replace the bare `include` in the `GET` at `src/app/api/meetings/route.ts:66` with an explicit `select` listing exactly the meeting fields the three consumers use (`id`, `title`, `scheduledAt`, `durationMinutes`, `location`/`meetingUrl`, `status`, `notes`, `relationId`, `createdAt`, plus the nested `relation` and conditional `guests` blocks already there).
2. Include `rsvpToken` **only** when `role === 'MENTEE'` — the same conditional-spread shape the `guests` block already uses two lines above, so the fix matches the file's own idiom.
3. Grep the consumers (`grep -rn "rsvpToken" src/`) and confirm nothing outside the portal path depends on it. If a mentor/admin component does read it, that is a second finding — fix or file it, do not silently keep the leak.
4. Add an e2e case: sign in as a MENTOR, fetch `/api/meetings`, assert no `rsvpToken` appears in the payload; sign in as the MENTEE on the same meeting and assert it does. `e2e/authz-idor.spec.ts` and `e2e/idor-hardening.spec.ts` are the right neighbourhood.

## ✅ Acceptance criteria
- [ ] `GET /api/meetings` returns `rsvpToken` only to the `MENTEE` the meeting belongs to
- [ ] The endpoint uses an explicit field `select`, so a future column is not leaked by default
- [ ] `MeetingsManager`, `MeetingSchedulerPanel` and the portal meeting views all still render (no consumer lost a field)
- [ ] An e2e spec asserts the token's absence for a MENTOR and its presence for the MENTEE
- [ ] `npm run build` passes
- [ ] a release fragment `releases/unreleased/meetings-rsvp-token-leak.json` is added (`bump: patch`)

## 🇹🇷 Türkçe özet
`GET /api/meetings` alan seçimi yapmadan bütün `Meeting` satırını döndürüyor; içinde `rsvpToken` de var. O jeton `/rsvp/<token>` ve `/api/calendar/<token>` için tam yetkili genel kimlik: elinde tutan kişi oturum açmadan mentee'nin yerine davete yanıt verebilir ve takvim kaydını çekebilir. Bugün bu jeton her mentöre ve her admine gidiyor. İlginç olan, aynı sorgudaki `guests` listesinin zaten role göre koşullandırılmış olması (`:64-75`) — yani düzeltmenin kalıbı iki satır yukarıda hazır duruyor. Açık bir `select` yazılacak ve `rsvpToken` yalnızca `MENTEE` rolüne eklenecek. Küçük, tek dosyalık, stajyer için uygun ama gerçek bir sızıntıyı kapatıyor.

## 🤖 Agent prompt
```
Objective: stop GET /api/meetings from returning the mentee's rsvpToken to mentors and
admins in the Internship CRM.

This is a small, well-bounded task — a good first issue.

Repo: Next.js 15 App Router, Prisma 5. Read CLAUDE.md first.

Open, in order:
  src/app/api/meetings/route.ts:44-80   (the GET; note the conditional `guests` block at
                                         64-75 — that spread is EXACTLY the shape your fix
                                         should take)
  src/middleware.ts:20                  (why the token is a credential: /api/rsvp is a
                                         public route and the token IS the auth)
  src/components/MeetingsManager.tsx, src/components/MeetingSchedulerPanel.tsx
  e2e/authz-idor.spec.ts, e2e/idor-hardening.spec.ts

Steps:
1. Run `grep -rn "rsvpToken" src/` and list every consumer before you change anything.
2. Replace the bare `include` at :66 with an explicit `select` naming only the meeting
   fields those consumers actually use, keeping the existing nested `relation` and the
   role-conditional `guests` blocks.
3. Add rsvpToken to the select ONLY when role === 'MENTEE'.
4. Add an e2e case: as a MENTOR, GET /api/meetings and assert the payload contains no
   rsvpToken; as the MENTEE on that meeting, assert it does.

Constraints:
- The mentee portal LEGITIMATELY needs its own token. Do not remove the field outright.
- Do not change the `where` clauses — role scoping there is already correct and
  deliberate (see the comment at :48-51).
- No schema change, so no prisma commands.
- e2e pitfall: getByText does substring matching; scope locators to a container or use
  { exact: true }.

Done when: the e2e spec proves both directions, `npm run build` and `npm run lint` pass,
releases/unreleased/meetings-rsvp-token-leak.json exists, and a PR is open with the
https://crm-pr<N>.ersah.in preview URL posted.
```

**Zorluk** · S · **Tahmini** · 3-4 hours · **Bağımlılık** · yok

mersahin:
## Summary

`GET /api/meetings` read whole `Meeting` rows through a bare `include`, so every column was serialised into the response — including `rsvpToken`. That token is a bearer credential, not an internal id: `/rsvp/<token>` is on the middleware's public allowlist (the unguessable token *is* the auth) and `/api/calendar/<token>` serves the event. Any ADMIN or MENTOR calling the endpoint was handed the mentee's token and could answer the invitation as them, or pull their calendar entry, with no session at all.

The fix is at the route boundary and takes the shape the file already uses two lines below: the query now names its fields with an explicit `select` — an allowlist, so a future column is not leaked by default rather than removed after the fact — and adds `rsvpToken` only when the caller's role is `MENTEE`, exactly like the role-conditional `guests` block. The mentee portal keeps its token: `UpcomingMeetings` RSVPs in-app and builds the `/api/calendar/<token>` link from that field, so removing it outright would have broken the mentee's own screen.

No authorization was touched: the per-role `where` clauses are unchanged, and the mentor/admin payload keeps every field its consumers read. `relation` was narrowed from `include` to a `select` of the `mentee` `id`/`fullName` that `MeetingsManager` actually renders.

I checked every other read of the model for the same leak. `/api/rsvp`, `/api/calendar/feed/<token>`, `/api/meetings/[id]/end`, `/api/meetings/[id]/guests`, `lib/meetingAccess.ts`, `lib/upcomingMeeting.ts` and `lib/meetingGuests.ts` all select explicitly already (the guest lib deliberately strips the token from what it returns), `/api/meetings/instant` and `/api/meeting-requests/[id]` return only ids, and `/api/calendar-events` over-fetches from the DB but maps to a fixed event shape, so nothing else serialises a token. `/api/meetings` GET was the only leak.

Closes #1548

## Changes

- `src/app/api/meetings/route.ts` — the GET query's `include` becomes an explicit field `select` (`id`, `relationId`, `title`, `scheduledAt`, `timeZone`, `meetLink`, `rsvp`, `endedAt`, `createdAt`), with `rsvpToken` spread in only for `role === 'MENTEE'`; the existing role-conditional `guests` block is preserved verbatim and `relation` is narrowed to the mentee fields the consumers use. Comment explains why the token is a credential.
- `e2e/meetings-rsvp-token-scope.spec.ts` — new spec: seeds an admin, a mentor, a mentee, a relation and a meeting; signs in as the mentor and asserts the row has no `rsvpToken` property *and* that the token string appears nowhere in the payload (while title, `meetLink` and `relation.mentee` are still there — proving nothing but the credential was removed); the same for the admin; then signs in as the mentee and asserts the token is still served.
- `releases/unreleased/meetings-rsvp-token-leak.json` — release fragment, `bump: patch`. No `notes`: nothing changes on any screen, so there is no user-facing highlight to write.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint prints only the two pre-existing `react-hooks/exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`)
- [ ] `npm run test:e2e` passing (or CI green) — this container has no database and no browser, so the new spec was written but not executed; leaving it to CI
- [x] Tested the change manually / added an e2e test — added the e2e test
- [x] `npm run check:i18n` OK (no new strings) · `npm run check:release-fragments` OK (ships as `0.156.23-beta`)

Preview: https://pr2204.interncrm.com

## Contributor terms


mersahin:
## Review follow-up: the RSVP-token scope check now runs in the PR gate

Pushed as `5324823`.

**Accepted the finding.** The reviewer is right, and it is the sharpest kind of gap: `e2e/meetings-rsvp-token-scope.spec.ts` was the only spec proving the mentee's `rsvpToken` is withheld from a mentor and an admin, and none of its three cases carried a tag. Since `.github/workflows/e2e.yml` runs `--grep @smoke` only, the sole regression test for a bearer-token leak guarded neither this PR nor the PR that eventually reintroduces the leak — and `main` auto-deploys straight to production. A test that runs only 4×/day is a detector, not a gate.

Tagged the **mentor** case `@smoke`.

**Smoke set size: 101 → 102 tests** (measured with `npx playwright test --grep @smoke --list`, before and after; 58 → 59 files). Worth flagging separately: CLAUDE.md still describes the smoke set as "~15-20 tests", which is off by 5× from what the gate actually runs today. I have not touched that line — it is out of scope here, but it means "keep the set small" is currently guidance nobody is measuring against.

**Why only one of the three.** I tagged the mentor case and deliberately left the other two untagged:

- The **admin** case exercises the *same* single conditional. The fix is one `...(role === 'MENTEE' ? { rsvpToken: true } : {})` spread inside one `select`; MENTOR and ADMIN differ only in the `where` clause, which is not what is under test. Tagging it would add gate time for a second pass through identical logic.
- The **mentee** case guards the opposite over-correction (someone drops the field outright and breaks the portal's RSVP + "add to calendar"). That is a loud, visible breakage rather than a silent credential leak, and the scheduled suite catches it soon enough.

The mentor case is also the richer of the two non-owner tests — it asserts the meeting is still returned intact (`title`, `meetLink`, `relation.mentee.id`), so it pins that this stayed a field fix and did not quietly become an authorization change.

**Self-containment: confirmed, no change needed.** A `@smoke` spec runs against the isolated MySQL service, so I checked this rather than assumed it. `beforeAll` seeds its own admin, mentor and mentee via `uniqueEmail()` (collision-proof), its own `MentorshipRelation`, and its own `Meeting` with an `rsvpToken` it generates locally with `randomBytes` — so the assertion compares against a token the test itself knows, not one it scraped from shared state. `afterAll` removes notifications, meetings, the relation and all three users. It borrows nothing from another spec and leaves nothing behind.

**Rejected one sub-point, respectfully: the assertion was already correct.** The review suggested the assertion might be of the `expect(body.rsvpToken).toBeUndefined()` variety that passes vacuously on a reshaped payload or a 500. It is not — that hazard was already designed out:

```ts
const res = await page.request.get('/api/meetings');
expect(res.status()).toBe(200);          // a 500 fails here, before any parsing
...
const own = meetings.find((m) => m.id === meetingId);
expect(own).toBeDefined();               // a reshaped payload fails here
expect(own).not.toHaveProperty('rsvpToken');            // on the meeting object itself
expect(JSON.stringify(meetings)).not.toContain(token);  // and nowhere else in the payload
```

The request success is asserted first, the lookup is by id so a nesting change fails loudly instead of silently, and the property check runs on the actual meeting object — plus a whole-payload sweep for the literal token string, which catches the leak even if it reappears under a different key or in a nested relation. Changing working code here would have made it weaker, not stronger, so I left it alone.

**Fragment updated.** `releases/unreleased/meetings-rsvp-token-leak.json` claimed the spec "asserts the token is absent for a mentor and an admin"; that claim is now backed by the PR gate rather than only the scheduled suite, so the prose says so.

**Static checks** (no DB, no browser in this container, so no Playwright run): `npx tsc --noEmit` clean, `npm run lint` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`, untouched by this branch), `npm run check:i18n` OK (4151 keys × 3 locales), `npm run check:release-fragments` OK (2 pending, ships as `0.156.23-beta`). Schema untouched, so no `prisma validate`. The tag itself is verified by the `--list` run above, which resolves to exactly `meetings-rsvp-token-scope.spec.ts:90 › …withholds the rsvpToken from the mentor`.

---

Issue #1820, opened 2026-08-30
🐛 🌱 [Matching] Task · The mentor directory truncates at 500 — move skill filtering into SQL


mersahin:
## 🎯 Goal
Search the mentor directory for a skill and you get a confidently wrong answer: the route reads the first 500 consented mentors, filters those in JavaScript, and reports the count of what survived — so a mentor who holds the skill but sorts 501st alphabetically simply does not exist, and nothing on screen says the result is partial. With the taxonomy join table in place this becomes an ordinary database query.

## 🔍 Current state
- `src/app/api/mentors/route.ts:38-64` — `prisma.user.findMany` with `take: 500`, ordered by `fullName: 'asc'`. The comment at `:45-47` states the reason honestly: the skill/language filters match inside JSON arrays, which MySQL `contains` cannot do case-insensitively.
- `:102-107` — `mentors.filter(...)` matching `skills.some(s => s.toLowerCase().includes(skill))` **or** a substring of the free-text `interests` field.
- `:108-113` — language and `acceptingOnly` filtered in JS too.
- `:115-116` — `total` and the page slice both come from the already-truncated, JS-filtered array.
- The two sibling routes already fixed their *totals* under #1392 but still fetch-then-filter: `src/app/api/candidates/route.ts:127-141` and `src/app/api/company/talent-pool/route.ts:83-105`. Both carry the same comment about MySQL JSON arrays.
- Visibility is a double opt-in that must not move: `publicProfile: true` **and** an active `MENTOR_DIRECTORY_VISIBILITY` consent (`src/app/api/mentors/route.ts:38-44`), plus a fail-closed role gate at `:21-25` (COMPANY/SOURCE get 403).
- Availability is derived by the shared `getMentorAvailability` (`src/lib/mentorAvailability.ts`), fed by one `groupBy` over active relations at `:66-76` — keep that; it is the single source of truth reused by four screens.
- Specs to extend: `e2e/mentor-directory.spec.ts`, `e2e/mentor-directory-consent.spec.ts`, `e2e/skill-filter.spec.ts`.

## 🛠 What to build
1. Change `?skill=` to accept a canonical skill id (keep accepting a free-text term for one release, resolving it through `resolveSkill()` server-side) and express the filter as a relation `where`: `skills: { some: { skillId: { in: [...] } } }` over `UserSkill`.
[...]

Issue #1815, opened 2026-08-30
🗃 [Matching] Task · Skill/SkillAlias/SkillEdge/UserSkill schema and the src/lib/skills.ts resolver


mersahin:
## 🎯 Goal
Turn "a skill" from a string somebody typed into a row the database can join, count and reason about. This is the one schema change the whole epic stands on, and the one the CTO review flagged as destructive under `db push` — so it lands additively, behind the migrations epic, with a pure resolver module that every later task imports instead of re-implementing string comparison.

## 🔍 Current state
- No skill model exists in `prisma/schema.prisma` (118 models, none of them a skill). Skills live as `User.skills Json @default("[]")` at `prisma/schema.prisma:221`, `User.skillLevels Json @default("{}")` at `:225`, and `Requisition.requiredSkills Json @default("[]")` at `:2562`.
- The org-scoped curated-vocabulary pattern to copy is `Tag` / `UserTag` — `prisma/schema.prisma:143-174`: required `orgId`, `@@unique([orgId, name])`, `@@index([orgId])`, cascade from `Organization` and from the owner, and a join table whose deletion removes the label but never the person.
- Normalisation helpers already exist and should be reused rather than reinvented: `normalizeTagName` / `tagKey` (`src/lib/tags.ts:33-45`, keying on `toLocaleLowerCase('tr')` so `İ`/`ı` behave) and `normalizeSkills` with `transliterate()` (`src/lib/requisitions.ts:47-59`).
- `TENANT_MODELS` (`src/lib/orgContext.ts:43-55`) lists 11 models today; the middleware only auto-scopes what is registered there, and it is dormant unless `MT_ENFORCE_ISOLATION=true` (`src/lib/orgContext.ts:16-19`).

## 🛠 What to build
1. In `prisma/schema.prisma`, add — all additive, nothing dropped:
   - `Skill` — `id`, `orgId`, `key` (normalised, unique per org), `labelEn` / `labelTr` / `labelDe`, `category`, `status` (`ACTIVE` | `RETIRED`), `mergedIntoId` (self-relation, so a merged skill still resolves), `createdById`, `createdAt`. `@@unique([orgId, key])`, `@@index([orgId, status])`.
   - `SkillAlias` — `id`, `orgId`, `skillId`, `key`. `@@unique([orgId, key])` so an alias can never point at two canonical skills.
[...]

PR #2205, opened 2026-09-06
fix: mentor directory reports an honest total instead of a silent 500-row cap (#1820)

mersahin:
## Summary

Search the mentor directory today and you get a confidently wrong answer. `/api/mentors` read the first 500 consented mentors with `take: 500`, filtered *those* in JavaScript, and reported the number of survivors as `total`. A mentor who holds the searched skill but sorts 501st by name simply did not exist as far as the directory was concerned — and nothing on screen said the result was partial. Silence was the bug.

This PR removes the truncation from the path that never needed it, and makes the remaining one speak up.

**Refs #1820** — this PR deliberately does **not** close the issue; see *What stays open* below.

## Changes

**`src/app/api/mentors/route.ts`**

- `take: 500` is gone.
- The handler now has two honest branches, mirroring the pattern `/api/candidates` already uses (#1392):
  - **No in-memory filter** — the whole query is expressible in SQL, so the page comes from `skip`/`take` and `total` comes from `prisma.user.count({ where })` over the **same** `where` the page was sliced from. No cap, no partial.
  - **An in-memory filter is active** — `skill` and `language` match *inside* the `skills` / `languages` JSON arrays, which MySQL cannot search case-insensitively; `accepting` is derived by `getMentorAvailability()` from a *counted* number of ACTIVE relations, so it is not a column either. That pass stays until the taxonomy join table (#1815) lands — the comment in the code says exactly that. What changes is the honesty: `total` describes the **filtered** set, the scan cap is a named `JS_FILTER_SCAN_CAP = 2000` rather than a bare `500`, and *reaching* it sets `partial: true` and returns the `cap` that was applied.
- The response always carries `partial` (a boolean), so the client never has to infer completeness from silence.

**`src/app/mentors/page.tsx`**

- Renders a **"Showing 1–12 of 37 mentors"** line above the cards (`data-testid="mentors-total"`).
- Renders an explicit amber **"this list may be incomplete — only the first {cap} listed mentors could be searched"** notice (`data-testid="mentors-partial-notice"`) whenever the API sets `partial`. It sits above the results, so it is also visible on a zero-result page — a capped scan finding nothing is precisely the case where "no matches" would be a lie.
- The notice uses the established `bg-amber-50 / border-amber-200 / text-amber-800` combination, which the existing `html.dark .bg-amber-50.text-amber-800` compound override in `globals.css` already retints for dark mode.

**`src/i18n/dictionaries.ts`** — new `mentorDirectory.showing` and `mentorDirectory.partialNotice` keys in EN/TR/DE.

**`e2e/mentor-directory.spec.ts`** — a new case asserting the total is honest: with `pageSize=1` the unfiltered `total` exceeds the returned rows (so it came from a `count`, not from the page), the filtered `total` equals the filtered rows, `partial` is `false` for both, the "showing" line renders and the incompleteness notice does not.

**`releases/unreleased/mentor-directory-honest-total.json`** — `patch`, with EN/TR/DE user-facing notes.

## What is deliberately unchanged

Verified against the existing consent spec, which passes untouched:

- The double opt-in visibility rule — `publicProfile: true` **and** an active `MENTOR_DIRECTORY_VISIBILITY` consent — and the fail-closed role gate that 403s COMPANY and SOURCE.
- The strict `select` allowlist: `email` / `phone` / `whatsapp` are never read, so they can never leak. Not widened.
- `getMentorAvailability()` in `src/lib/mentorAvailability.ts`, shared by four screens and fed by one `groupBy`. Kept, not forked.

## What stays OPEN for the #1815 follow-up

Scope is cut on purpose, because the SQL half depends on work that does not exist on `main` yet — there is no `model UserSkill` in `prisma/schema.prisma` and no `resolveSkill()` anywhere in `src/lib`. So this PR is **Refs**, not **Closes**:

- **Acceptance criterion 1** (600+ consented mentors seeded; filtering by a skill held only by a mentor sorting after the 500th returns that mentor) is **not** met. The skill filter still runs in JavaScript over a capped read — the cap is now 2000 and announced rather than 500 and hidden, but it is still a cap.
- **`e2e/skill-filter.spec.ts` is not extended with the beyond-500 case.** There is no database and no browser in this container, so a spec that seeds 600 mentors could be neither run nor trusted here; writing an unexecuted 600-row seed would be a guess wearing a test's clothes.
- **`/api/candidates` and `/api/company/talent-pool` are untouched.** Both already report honest totals from #1392; their remaining unbounded fetch-then-filter is only fixable with the join table.
- No taxonomy schema, no relation `where`, no `resolveSkill()`, no canonical-skill-id `?skill=` param.

## Verification

Static checks only — no database and no Docker daemon in this container, so the app and the Playwright suite could not be run here.

- [x] `npm run lint` — clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`)
- [x] `npm run check:i18n` — OK, 4153 keys × 3 locales
- [x] `npm run check:release-fragments` — OK, ships as `0.156.23-beta`
- [ ] `npm run test:e2e` — not runnable here; CI is the gate
- [x] Added an e2e test (`e2e/mentor-directory.spec.ts`)

Preview: https://pr2205.interncrm.com

## Contributor terms


mersahin:
Review follow-up pushed as `346dcd9`. All four findings were verified against the code; three were real defects and are fixed, the fourth is a scope call I agree with and have addressed as far as it can be addressed without #1815.

**1 — unstable paging (major). Confirmed, fixed.** `prisma/schema.prisma` really does carry only `@@index([role])` and `@@index([referredById])` on `User`, so `ORDER BY fullName ASC LIMIT n OFFSET m` is a filesort over a non-unique column and ties are free to move between plans. With `skip`/`take` on top of that, two mentors called the same thing could appear on page 1 and page 2, or on neither — exactly the failure #1820 is about, newly created by the SQL branch. The order is now `[{ fullName: 'asc' }, { id: 'asc' }]`, matching the precedent at `src/app/api/offers/route.ts:180`, and both branches share it.

I did **not** add `@@index([fullName])`. The `where` leads with a `consents: { some: … }` subquery plus four column predicates, so the optimiser filters first and would not pick the index for the sort anyway; it is a schema change that makes the next prod `db push` add an index to `User` for a code path #1815 is about to restructure. Worth revisiting if the directory ever grows past a few thousand rows, but not as part of this fix.

**2 — cap detection off by one (minor). Confirmed, fixed.** `rows.length >= CAP` cannot distinguish "exactly CAP rows exist" from "more exist", so a complete answer of exactly 2000 mentors rendered the amber "narrow your filters" banner — advice that cannot help, because nothing was left out. Now a `limit + 1` probe: read `CAP + 1`, drop the probe row, `truncated = rows.length > CAP`.

**3 — read amplification on the debounced path (minor). Confirmed, fixed via the second option.** Rather than reverting the cap to 500, the scan now reads only the six columns the JS filter actually touches (`id`, `skills`, `languages`, `interests`, `mentorCapacity`, `acceptingMentees`) — no `bio`, no `avatarUrl` — and the full card payload is re-read for the ≤ `pageSize` rows that survive. The availability `groupBy` moved after the text filters, so the ~54 KB `IN` list only happens when `accepting=1` is set on its own; the unfiltered branch now runs it over the page alone instead of the whole read. The second read re-applies the visibility `where` alongside the ids, so the allowlisted select is still unreachable through a bare id.

**4 — AC 1 unmet, `partial` branch untested (minor). Agreed on both halves.** The prerequisite is real: there is no `Skill`/`UserSkill` model on main, so the skill filter cannot leave JavaScript yet. The PR stays `Refs #1820` and the issue stays open.

The coverage gap I could close, and did. The scan-cap and filter arithmetic moved into `src/lib/mentorDirectory.ts` as I/O-free functions, with `scripts/test/mentor-directory.test.mjs` (10 cases) wired into `ci.yml` next to the existing `test:stage-aging` gate — it drives the truncation branch with a cap of 3, including the exactly-at-cap case from finding 2, which no e2e run could ever reach. A new e2e case stubs a `partial: true` response so the notice, the `cap` payload and the `{cap}` substitution are exercised in a browser at least once, rather than only ever asserted absent.

Static checks after the changes: `tsc --noEmit` clean, `next lint` clean for the touched files (only the pre-existing `exhaustive-deps` warnings elsewhere), `check:i18n` OK (4153 × 3, no new keys needed), `check:release-fragments` OK (ships `0.156.23-beta`), plus `check:auth-reads`, `check:query-scalars` and `check:openapi` OK, and the new unit suite 10/10. The release fragment's changelog prose was extended to describe the tiebreaker, the probe and the narrowed scan; the user-facing notes were already accurate and are unchanged. No schema change, so no `prisma validate` run needed.

---

PR #2206, opened 2026-09-06
fix: raise the muted-text token and size the inbox person-card trigger (#2131)

### LINKED ISSUE #2131: a11y: muted `text-gray-400` secondary text and the 14px person-card trigger fail WCAG AA
Split out of #2043 (PR #2115), which widened the axe scan to `/messages`, `/notifications`, `/mentor/board`, `/admin/board`, `/admin/settings` and `/apply/:mentorId`. Widening the scan surfaced nine *serious* violations that already existed on those shipped screens — none of them introduced by that PR. Three other classes it surfaced (`select-name` critical, `scrollable-region-focusable`, `aria-command-name`) were unambiguous one-liners and were fixed in #2115 itself; these nine are not, so they are frozen in `e2e/a11y-baseline.json` and recorded here instead.

They are two root causes, not nine bugs.

### 1. `text-gray-400` (#9ca3af) as the muted secondary-text token — 7 findings

Measured contrast is **2.38–2.53:1** against white; WCAG AA wants 4.5:1 for body text.

| Page | Element |
|---|---|
| `/messages` | `<p class="text-xs text-gray-400 mt-0.5">` under `[data-testid="support-entry"]` — 2.53:1 |
| `/messages` | `<p class="text-xs truncate mt-0.5 text-gray-400">No messages yet…</p>` — 2.53:1 |
| `/notifications` | `<p class="text-xs text-gray-400 mt-1">` timestamp on an unread row — 2.38:1 |
| `/notifications#dark` | `<p class="text-sm text-gray-500 dark:text-gray-400">` body text — 3.40:1 |
| `/mentor/board` | `<div class="flex items-center gap-1 text-xs text-gray-400 mt-2">` on a card — 2.53:1 |
| `/admin/board` | the person-card trigger, inheriting `text-gray-400` from its row — 2.53:1 |
| `/admin/settings` | `<span class="text-gray-400">○ Bulk channel off …</span>` — 2.53:1 |

Plus one that is the mirror image, a dark-mode retint rather than the token:

| `/admin/board#dark` | `<span class="text-sm font-semibold text-gray-800">` group header, #e5e7eb on #797f87 — 3.26:1 |

`text-gray-400` is used this way across the whole app, so this is a **token decision, not a per-screen patch**: raising it to `text-gray-500`/`text-gray-600` (and giving the dark board header its own override in `globals.css`, per the compound-override pattern in CLAUDE.md) will move other baseline keys too. Regenerate `e2e/a11y-baseline.json` with `A11Y_UPDATE_BASELINE=1 npx playwright test e2e/a11y-scan.spec.ts` in the same PR and check the printed "widens the baseline" warning is empty.

### 2. `target-size` on the person-card trigger in the inbox — 2 findings

`/messages` and `/messages#dark`: the `role="button"` person-card trigger in each inbox row wraps a `h-3.5 w-3.5` `<UserRound/>`, giving a **14×14px** target where WCAG 2.2 AA asks for 24×24. It now has an accessible name (fixed in #2115); it still needs a hit area — e.g. `inline-flex h-6 w-6 items-center justify-center` on the trigger, which is a visible spacing change to every inbox row and wants a look before it ships.

### Done when

- [ ] The nine entries above are gone from `e2e/a11y-baseline.json` (the six affected keys are back to `{}`).
- [ ] `docs/a11y-audit.md` regenerated in the same PR.
- [ ] The dark-mode half is checked too — the board header finding only exists in dark.

mersahin:
## Summary

The axe scan widened in #2115 froze nine *serious* violations across `/messages`,
`/notifications`, `/mentor/board`, `/admin/board` and `/admin/settings`. They were never
nine bugs — they were **two root causes**, and this PR fixes both.

**1. `text-gray-400` is the muted secondary-text token, and it fails AA.** #9ca3af measures
2.38–2.53:1 on white where WCAG AA asks 4.5:1 for body text. It is what this codebase reaches
for whenever text is "secondary": inbox previews, timestamps, board card meta rows, empty
states, the settings page's SMTP status lines. Six screens failing on the same class is the
shape of a *token* problem, so the token is raised **once**, in `src/app/globals.css`, to
gray-500 (`#6b7280`, 4.83:1 on white) — right beside the dark-mode and `prefers-contrast`
utility remaps that already live there and that CLAUDE.md documents as the house pattern.
Dark mode explicitly keeps `#9ca3af` (already 7.0:1 on gray-900); following the light-mode
raise there would be a regression, dark text on a dark surface.

**2. The icon-only person-card trigger is a 14×14 target.** WCAG 2.2 target-size (AA) asks
24×24. The trigger box is now sized rather than the glyph; the icon stays 14×14.

Closes #2131

## Changes

- **`src/app/globals.css`** — the token: `.text-gray-400 { color: #6b7280 }`, plus
  `html.dark .text-gray-400 { color: #9ca3af }` so dark mode is untouched. Placed *before*
  the `prefers-contrast: more` block, which pushes the same utility to gray-700 and keeps
  winning on source order at equal specificity.
- **`src/app/globals.css`** — `html.dark .bg-white\/40`. This is the dark-mode mirror image
  the issue calls out: unlike `/80` and `/70`, the admin board's stage-group panel was never
  retinted, so in dark it stayed a 40%-white wash (~#797f87) under a `text-gray-800` header
  remapped to `#e5e7eb` — 3.26:1. Retinted next to its siblings; the header lands at ~15:1.
- **`src/app/admin/board/page.tsx`** — the group's count chip pinned to `text-gray-700`. It
  carries its own `bg-gray-100`, where even the raised gray-500 only reaches 4.4:1 (4.1:1
  after the dark remap). The one place the token is deliberately *not* the answer.
- **`src/app/notifications/page.tsx`** — a read row loses `opacity-60`. Opacity fades the
  text **and** the card under it by the same amount (2.3:1 light, 3.4:1 dark), so no colour
  token can win that argument. Read rows are still distinct: gray border, muted text, no blue
  tint, no unread dot.
- **`src/app/messages/page.tsx`, `src/components/TargetedEmailComposer.tsx`** — the icon-only
  `PersonHoverCard` triggers get `inline-flex h-6 w-6 shrink-0 items-center justify-center`.
  Both, not just the scanned one: the composer's recipient list is literally the same pattern
  with the same 14×14 defect, it just is not on a page the scan visits.
- **`e2e/person-hover-card.spec.ts`** — asserts the trigger's bounding box is ≥ 24×24. That
  spec already exercises the composer trigger, so this needed no new fixture.
- **`e2e/a11y-baseline.json` / `docs/a11y-audit.md`** — all nine entries cleared; the six
  affected keys are back to `{}`. **See the caveat below.**
- **`e2e/a11y-scan.spec.ts`** — the header comment that explained why the nine were frozen
  now says how they were fixed. A frozen entry is a debt with an issue number; this one is
  paid.
- **`releases/unreleased/a11y-muted-text-contrast.json`** — `patch`, with EN/TR/DE notes
  (the change is visible on every screen).

### Why the token, and not ~590 class edits

`text-gray-400` appears **595 times in ~180 files** under `src/`. I grepped and categorised
before choosing:

| Category | Count | Treatment |
| --- | --- | --- |
| Body / secondary **text** (previews, timestamps, meta rows, empty states, table cells, status lines) | ~530 | **Raised** — this is the failing case. |
| **Icons** (`text-gray-400` on an `h-N w-N` lucide glyph) | ~61 | Raised **with** them, deliberately: 2.5:1 was below the 3:1 WCAG 1.4.11 asks of meaningful non-text content, so this is a fix there too, not collateral. |
| **Placeholders** | 3 | **Untouched** — they are `placeholder:text-gray-400` / `placeholder-gray-400`, different class names the rule does not match. |
| **Disabled** states | — | **Untouched** where they use `disabled:text-*`; where a disabled control is coloured with a bare `text-gray-400` it darkens, which is harmless (1.4.3 exempts them, it does not forbid contrast). |

Rewriting 530 call sites by hand across 180 files is a diff nobody can review and a sed
nobody can trust, and it would still miss the nodes axe reports but the audit doc does not
print. One rule, at the layer the issue says the decision belongs to, moves all of them.
Only two files needed a per-element decision on top (the gray-100 chip, the opacity row),
and both carry a comment saying why.

Nothing gets *lighter* anywhere, so this cannot widen the baseline on a page it touches by
accident — the worst case is a screen looking slightly darker than intended.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint's two `exhaustive-deps` warnings are
      pre-existing, in files this PR does not touch)
- [ ] `npm run test:e2e` passing (or CI green) — **cannot run here**, see below
- [x] Added an e2e test (`e2e/person-hover-card.spec.ts`, the 24px floor)

Also green: `npm run check:i18n` (no new strings) and `npm run check:release-fragments`
(ships as `0.156.23-beta`).

### ⚠️ The baseline and the audit doc were edited BY HAND

This work was done in a container with **no browser and no database**, so
`A11Y_UPDATE_BASELINE=1 npx playwright test e2e/a11y-scan.spec.ts` could not be run. I
removed exactly the nine entries the issue enumerates from `e2e/a11y-baseline.json` and
rewrote the generated half of `docs/a11y-audit.md` to match; the doc carries a temporary
note saying so, which the next real regenerate erases.

**A reviewer must run that command once against a real environment, commit whatever it
writes, and confirm the printed "⚠️ This regenerate WIDENS the accessibility baseline"
warning is empty.** Until then the six keys are an assertion, not a measurement — and if the
scan still finds a node on one of those pages, the scan is right and this PR is incomplete.

Preview: https://pr2206.interncrm.com

## Contributor terms


mersahin:
Review findings applied in 4ae59b0. All five were real; none rejected, but two are answered differently from the suggested fix and that needs stating plainly.

**1 (blocker) — `/admin/settings` still firing `color-contrast`.** Confirmed against the actual job log (run 34037736362, job 101498784164): one regression, `/admin/settings: color-contrast (serious) ×1 > baseline 0 — .text-green-600`, identical on attempt 1 and retry 1. Root cause is exactly as described — `scan()` maps `results.violations`, which axe already groups per rule, so the one baselined entry covered every node the rule matched. The offender renders on load: `loadEmailLog()` runs in a `useEffect` (settings/page.tsx:80-83), so the e-mail-log summary's `text-green-600 dark:text-green-400` (#16a34a, 3.30:1 on white) is on screen for the scan. Fixed by moving every status colour on that page to its -700 shade: green-700 and amber-700 are 5.02:1 on white and still 4.56:1 on gray-100, so the fix holds whatever surface the card sits on. Amber went too because `text-amber-600` is 3.19:1 — it only escaped the scan because the CSV-import result rows never render there, i.e. it was a latent red run, not a pass. That import row was also the one place with no `dark:` companion at all; it now has one, since green-700 on the dark ground is 4.02:1.

**2 and 3 (major) — the hand-written report and the spec header.** Both real, both fixed, but not by running the regenerate: this container has no browser and no database, so `A11Y_UPDATE_BASELINE=1` genuinely cannot run here. Rather than assert a measurement again, the report note now separates the two halves of the claim: **critical: 0 · serious: 0 is measured** — the smoke job runs this same spec on every push with the baseline at `{}` for all sixteen page/theme keys, so a green run *is* that assertion — while **moderate: 0 · minor: 0 and the empty table are carried over** from the last real regenerate and are explicitly labelled inherited rather than observed. The note also records the failed first attempt and the node it missed. The spec header keeps the per-entry accounting of which selector each frozen key covered (that deletion was a fair complaint) and adds the per-rule-not-per-node trap as the lesson. If you'd rather the doc carry no unobserved numbers at all, the honest alternative is to run the regenerate once on a real env and commit it — the note asks for exactly that.

**4 (minor) — fragment prose.** Both claims corrected. (a) now reads that the row *drops* `opacity-60` and that its existing gray muting plus its border carry the state on their own — the colour was already there on main. (b) is restated as what the gate proves rather than as a count of edited JSON keys.

**5 (minor) — the dark remap outranking `dark:text-*`.** Verified: `darkMode: 'class'` emits `.dark .dark\:text-gray-500` at (0,2,0), and `html.dark .text-gray-400` is (0,2,1). All ten occurrences in the tree were `text-gray-400 dark:text-gray-500`, so the `dark:` half was dead in every one of them. Took the second option and deleted them, so source and rendering agree, and documented the trap beside the rule for the next base-utility remap.

`npx tsc --noEmit`, `npm run lint`, `npm run check:i18n` and `npm run check:release-fragments` (0.156.23-beta) all pass; the schema is untouched. Preview env: https://pr2206.interncrm.com

---

Issue #2008, opened 2026-08-30
🌱 🩺 [Admin] Task · Integration health board — per-connector status for webhooks, Google, SSO and email


mersahin:
## 🎯 Goal
Silent integration failure is how trust in an integration layer dies. An admin today cannot tell that a webhook has been failing for a week, that a user's Google Calendar token expired, or that the SSO certificate is about to lapse — even though the product faithfully records all three. We already proved the pattern works for email; this task applies it to every other connector on one screen. It is mostly reading data that already exists and rendering it, which makes it a well-bounded first issue.

## 🔍 Current state
- **The pattern to copy**: `src/lib/emailHealth.ts:11-46` — an `EmailHealth` interface (`lastOkAt`, `lastErrorAt`, `lastError`, `failuresSinceOk`, `attempts24h`), derived from the `EmailLog` ledger rather than a separate marker *"so the two can never drift apart"*, with `sanitizeEmailError` at `:25-28` stripping anything address-shaped before it leaves the module.
- **Google Calendar**: `prisma/schema.prisma:2636-2639` stores `lastSyncAt` and `lastError` with the comment *"Operational state, so a broken connection is visible in the UI instead of failing silently"* — but it is only rendered on the user's own `/account` card (`src/components/GoogleCalendarCard.tsx`), never to an admin. `src/app/api/admin/integrations/google/status/route.ts` reports configured/enabled/connection-count only.
- **Webhooks**: nothing to read yet — the delivery log arrives with #1681/#1695. Until it lands, derive what is derivable (subscription count, active/paused, last created) and leave a clearly-marked slot for delivery health.
- **SSO**: certificate expiry is not surfaced anywhere; `docs/sso-saml.md` has the current picture.
- **Inbound receivers**: `src/app/api/webhooks/jaas/route.ts` and `src/app/api/inbound-email/route.ts` both answer 404 when unconfigured — worth showing as configured/not-configured.
- The page this lands on is `src/app/admin/integrations/page.tsx` (client component, loads `/api/admin/webhooks`, `/api/admin/api-keys` and `/api/admin/integrations/google/status` in one `Promise.all`).
- **#1607** is building `/admin/operations` (scheduled-job runs + error rate) and will call this task's `GET /api/admin/integrations/health` for a summary roll-up. That route is therefore a shared read surface, not a page-private helper — see § Scope boundary.

## 🛠 What to build
[...]

Issue #1991, opened 2026-08-30
🪟 [Calendar] Task · Microsoft 365 / Outlook — Graph OAuth, event write, and a local Graph stub


mersahin:
## 🎯 Goal
Our only calendar connector is Google, and it is off by default. The enterprise mentoring buyer is overwhelmingly a Microsoft shop — every serious competitor ships Outlook/Exchange sync, and a Teams-first HR buyer disqualifies us in the first demo. Guider is explicitly criticised for being Microsoft-only; we are the mirror-image failure. This is a straight procurement blocker. This task implements the second `CalendarProvider`: delegated Microsoft Graph OAuth, sealed tokens, and event create/update/delete — plus the local stub that makes it testable in CI, exactly the way the Google slice was made testable.

## 🔍 Current state
- No Microsoft Graph client exists anywhere in `src/` (`grep -rn "graph.microsoft" src/` is empty).
- The Google slice is the template, and it was deliberately built to be testable without the real provider: `src/lib/googleCalendar.ts:66-76` reads `GOOGLE_OAUTH_TOKEN_URL`, `GOOGLE_OAUTH_REVOKE_URL` and `GOOGLE_CALENDAR_API_BASE` from env with the real URLs as defaults, and the header comment at `:12-18` says why — "it is what makes the token exchange TESTABLE … the e2e now points these at a local stub and drives connect → callback → push → disconnect end to end". Do the same for Graph.
- Token handling to reuse verbatim: `src/lib/secretBox.ts` (AES-256-GCM, per-purpose HKDF — **use a new purpose string**, not `google-calendar`), the 60-second refresh skew and `noteError` in `src/lib/googleCalendarClient.ts`.
- The consent flow shape: `googleConsentUrl(state)` (`src/lib/googleCalendar.ts:47`) with an opaque state token the caller persists and re-checks on callback; routes under `src/app/api/integrations/google/`.
- `e2e/google-calendar.spec.ts` and `e2e/google-calendar-status.spec.ts` are the specs to mirror.
- This task assumes the provider-neutral layer from #1989 exists; if it does not, stop and land that first rather than writing a second parallel sync module.

## 🛠 What to build
1. **`src/lib/calendar/microsoft.ts`** implementing `CalendarProvider`:
[...]

PR #2207, opened 2026-09-06
feat: one "Connected calendars" card with connection health (#1993)

### LINKED ISSUE #1993: 🌱 🔌 [UX] Task · One &#34;Connected calendars&#34; card — both providers, health, and a reconnect prompt
## 🎯 Goal
Once there are two calendar providers, one card has to speak for both — and it has to tell the truth when a connection is broken. Today the connection stores `lastError` and `lastSyncAt` and nothing on screen reads them for what they are: a person whose token was revoked keeps seeing "connected" while their meetings quietly stop syncing. This is a small, self-contained UI + read-endpoint task, and a good first issue: the data is already there, the pattern already exists elsewhere in the codebase, and there is no schema change.

## 🔍 Current state
- `src/components/GoogleCalendarCard.tsx` is the whole surface today, hard-wired to Google.
- The connection row already carries everything the card needs: `googleEmail` (shown back "so they can tell WHICH calendar this is before they trust it or disconnect it"), `scope` ("stored so a scope that was reduced on re-consent is visible rather than discovered by a failing write"), `lastSyncAt` and `lastError` ("so a broken connection is visible in the UI instead of failing silently on every meeting") — `prisma/schema.prisma:2617-2646`.
- `noteError(connectionId, msg)` in `src/lib/googleCalendarClient.ts` is already called on every failed write (`src/lib/googleCalendarSync.ts:73`), so the error text exists and is fresh.
- The health-surface pattern to copy is already in the repo: `src/lib/emailHealth.ts` renders channel health for the admin integrations page.
- `e2e/google-calendar-status.spec.ts` already asserts the current card's states — extend it, do not replace it.

## 🛠 What to build
1. Rename/rewrite `GoogleCalendarCard.tsx` as `ConnectedCalendarsCard.tsx`, rendering **one row per provider** (Google, Microsoft) with three states: not configured by the operator / configured but not connected / connected.
2. For a connected row show: the account e-mail, "last synced <relative time>", and — when `lastError` is set — a warning strip with the error and a **Reconnect** button that restarts the consent flow for that provider only.
3. A small read endpoint (`GET /api/integrations/calendar/status`) returning, for the signed-in user only, `[{ provider, configured, enabled, connected, accountEmail, lastSyncAt, lastError }]`. **Never** return a token, a scope string that reveals another user's setup, or any other user's row — the route reads `session.user.id` and nothing else.
4. Mount it wherever the Google card is mounted today, and add it to the mentor and admin account/settings surface so all three roles can connect a calendar.
5. EN/TR/DE keys for every string: the three states, the relative-time label, the error strip and the reconnect button.

## ✅ Acceptance criteria
- [ ] Both providers appear on one card, each with its own connect/disconnect action
- [ ] A provider the operator has not configured says so, instead of offering a button that cannot work
- [ ] A connection with `lastError` set shows a warning and a working Reconnect button
- [ ] The status endpoint returns only the signed-in user's own connections and never a token
- [ ] The card is reachable by MENTOR, MENTEE and ADMIN
- [ ] `npm run build` passes
- [ ] `e2e/google-calendar-status.spec.ts` is extended to cover the error/reconnect state
- [ ] A release fragment `releases/unreleased/connected-calendars-card.json` is added

## 🇹🇷 Türkçe özet
İki takvim sağlayıcısı olunca tek bir kartın ikisi için de konuşması gerekiyor — ve bağlantı bozulduğunda bunu dürüstçe söylemesi. Bugün bağlantı satırında `lastError` ve `lastSyncAt` duruyor ama ekranda hiç kimse bunları okumuyor: token'ı iptal edilmiş bir kullanıcı "bağlı" yazısını görmeye devam ediyor, toplantıları ise sessizce senkronize olmuyor. Bu küçük ve kendi içinde kapalı bir iş: veri zaten var, desen zaten kod tabanında (`emailHealth.ts`), şema değişikliği yok — stajyer için uygun. Tek güvenlik kuralı: durum uç noktası yalnızca oturum sahibinin kendi bağlantılarını döndürür ve hiçbir koşulda token döndürmez.

## 🤖 Agent prompt
```
Objective: replace the Google-only calendar card with one "Connected calendars" card that
covers both providers and surfaces a broken connection honestly.

This is a small, well-bounded task - a good first issue. No schema change.

Open, in order:
  src/components/GoogleCalendarCard.tsx        (what exists today)
  prisma/schema.prisma:2617-2646               (read the doc comments - they say exactly
      why googleEmail, scope, lastSyncAt and lastError are stored: to be SHOWN)
  src/lib/googleCalendarClient.ts              (noteError - where lastError comes from)
  src/lib/emailHealth.ts                       (the health-surface pattern to copy)
  src/app/api/integrations/google/             (connect / disconnect routes)
  e2e/google-calendar-status.spec.ts           (extend, do not replace)
  src/i18n/dictionaries.ts                     (EN/TR/DE - all three or CI fails)

Constraints:
  - GET /api/integrations/calendar/status reads session.user.id and returns ONLY that
    user's connections. Never return accessTokenEnc/refreshTokenEnc or anyone else's row.
  - "Configured by the operator" and "connected by me" are two different states - do not
    collapse them. A provider with no credentials in env must say so rather than showing a
    dead button.
  - i18n: every new string needs EN/TR/DE keys or `npm run check:i18n` fails.
  - e2e pitfall: getByText does substring matching - scope locators to the card container
    or use { exact: true }.

Done when: both providers render with correct states, an errored connection shows a
reconnect prompt, the e2e covers it, build + check:i18n pass, a release fragment exists
and a PR is open with the preview URL.
```

**Zorluk** · S · **Tahmini** · 4-6 saat · **Bağımlılık** · #1991

mersahin:
## Summary

A calendar connection already stored `lastSyncAt` and `lastError`, and nothing on screen read them. Someone whose Google token was revoked kept seeing "connected" while their meetings quietly stopped syncing — the failure was only visible in the database. This replaces the Google-only card with a provider-shaped **Connected calendars** card that tells the truth about each connection, and adds the small per-user read endpoint behind it.

**Scope note — Microsoft (#1991) is deliberately not a row here.** Microsoft 365/Outlook is not implemented anywhere in the repo, so a literal "Microsoft" row today could only be a hardcoded placeholder permanently reading "not configured": dead UI that lies about a provider with no code behind it. What ships instead is the provider-**shaped** card and endpoint, with Google as the single registry entry. **#1991 adds an entry to the array in `src/lib/calendarProviders.ts` (plus its id in the `CalendarProviderId` union) and neither the card nor the endpoint changes** — the row testids are derived from the provider id too, so a new provider gets `microsoft-calendar-connect` and friends for free.

Closes #1993

## Changes

- **`src/lib/calendarProviders.ts` (new)** — an *array* registry (order is a product decision; a map's key order is not), one entry per provider: id, brand label, `isConfigured()` / `isEnabled()` reading the operator's env credentials, the connect/disconnect route paths, and a `readConnection(userId)` that fetches one person's row behind an **explicit column allowlist**.
- **`GET /api/integrations/calendar/status` (new)** — one row per registry entry, `{ provider, label, configured, enabled, connected, accountEmail, lastSyncAt, lastError, connectPath, disconnectPath }`, for the signed-in user only. Security, as required by the issue:
  - the only identity it reads is `session.user.id`; there is no `userId` parameter to tamper with;
  - `accessTokenEnc` / `refreshTokenEnc` are **never selected at all**, rather than selected and stripped afterwards — an allowlist can't be defeated by a forgotten edit;
  - `lastError` is provider text that can echo the account or a token, so it goes through the shared `sanitizeError` (#2008) on the way out.
- **`src/components/ConnectedCalendarsCard.tsx` (new, replaces `GoogleCalendarCard.tsx`)** — one row per provider with four states that are deliberately *not* collapsed:
  1. **not configured by the operator** — says so; no button that could only bounce back with `unavailable` (the old card just rendered nothing at all here);
  2. **configured but switched off for everyone on this installation** — a different sentence from (1): "never set up" and "turned off" are different facts, and only one of them is likely to change back;
  3. **available but not connected by me** — offers Connect;
  4. **connected** — the account e-mail (so you can tell *which* calendar you are about to disconnect), a relative "last synced …" label, Disconnect, and an amber warning strip whenever meetings are **not** reaching that calendar: either because `lastError` is set (the provider's own message — `invalid_grant` is the difference between "reconnect" and "call support" — plus a **Reconnect** button that restarts consent for that provider only), or because the provider is switched off here (no Reconnect: the connect route checks the same switch and can only bounce; Disconnect stays, because revoking must always be possible).

  The card is driven entirely by the status endpoint, so nothing in the browser names Google.
- **Mounting** — swapped in `AccountSettings`, which is what `/account` (MENTOR + MENTEE, linked from both shells' account menus) and `/admin/account` (ADMIN) render, so all three roles reach it through the one mount. No second copy was added: a duplicated card would be a second thing to keep in sync.
- **i18n** — a new `connectedCalendars` namespace in EN/TR/DE: the four states, the relative-time label, the "never synced" case, both strips, the Reconnect button and a failed-disconnect message. `googleCalendar`'s flash messages stay as they are (the OAuth callback still redirects with `?google=…`).
- **e2e** — `e2e/google-calendar-status.spec.ts` **extended, not replaced** (the two existing admin-only tests are untouched, as is `/api/admin/integrations/google/status`): the endpoint returns only my own row and never a token even while a second user's connection exists, it 401s a signed-out visitor, a broken connection renders the warning strip (asserted on the whole rendered sentence, and that it carries no `<redacted>`) + a Reconnect button pointing at `/api/integrations/google/connect`, and a healthy connection renders neither strip. One line in `e2e/google-calendar.spec.ts` follows the card container's testid rename; the per-provider testids keep the old Google names.
- **Release fragment** — `releases/unreleased/connected-calendars-card.json` (`minor`, with EN/TR/DE notes).
- **Retrospective** — a dated entry in `docs/agent-experience.md`, plus an addendum covering the review findings below.

The existing OAuth routes under `/api/integrations/google/*` (connect / callback / connection) are untouched — this only adds an aggregate read. No schema change.

## Review follow-up (second commit)

All three review findings were verified against the code and fixed:

1. **(major) A connected row ignored `enabled`.** `available = configured && enabled` was computed but only consulted in the two *not connected* branches. With the provider switched off — `GOOGLE_CALENDAR_ENABLED` unset is what production ships — the card affirmatively read "Connected as … / Last synced 3 hours ago" while `pushMeeting()` returns at its first line and `lastError` is never written, because the error path sits behind that same early return. State (2) and the connected-but-unavailable strip above are the fix; the old `GoogleCalendarCard` at least rendered nothing here, so this was a regression, not a pre-existing gap.
2. **(minor) `sanitizeError` ate the word after any literal "token".** That is three of the four `lastError` strings this app writes: `Token refresh failed` → `<redacted> failed`, `invalid_grant: Token has been expired or revoked.` → `invalid_grant: <redacted> been expired or revoked.` — mangling the one informative line in the strip that promises the provider's own words. The rule is now split: an explicit `:`/`=` still redacts whatever follows, however short, while a whitespace-separated value is redacted only when it looks like a credential (12+ characters carrying non-alphabetic material). Checked against real `Authorization: Bearer …`, `token=1//0g…`, `password: hunter2`, `api_key=icrm_…` and `client_secret: "…"` samples — all still redacted — and against the three provider sentences, which now survive intact. Every surface that renders connector errors (integration health, e-mail health) gets the same improvement.
3. **(minor) `disconnect()` never checked the DELETE response.** A refusal still flipped the row to "not connected" and flashed "disconnected and access revoked" while the sealed tokens stayed in the database. This is reachable today, not hypothetical: `middleware.ts` 403s every write method for an unverified account, and `/api/integrations/google/connect` is a GET, so an unverified user can connect and then fail to disconnect. It now branches on `res.ok` (new `connectedCalendars.disconnectFailed` in EN/TR/DE) and, on success, **re-reads** the status endpoint rather than patching local state — what the card claims comes from the server that did the work.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint reports only the two pre-existing `exhaustive-deps` warnings on unrelated admin pages)
- [x] `npm run check:i18n` — OK, 4166 keys × 3 locales
- [x] `npm run check:release-fragments` — OK, `0.156.21-beta -> 0.157.0-beta`
- [ ] `npm run test:e2e` — **not run here**: this container has no database and no browser, so the new specs are written but were not executed. CI is the check.
- [x] Added e2e coverage for the error/reconnect state, asserted on the rendered strip text

Preview: https://pr2207.interncrm.com

## Contributor terms


mersahin:
Review findings addressed in e7dff70 — all three were real, none rejected.

**1. (major) Connected row ignored `enabled`.** Confirmed: `available` was computed at `ConnectedCalendarsCard.tsx:123` and consulted only in the two *not connected* branches; the connected branch was a bare `{row.connected && …}`. With `GOOGLE_CALENDAR_ENABLED` unset the status route still returns the row (it gates on `configured`, by design), so the card read "Connected as … / Last synced 3 hours ago" while `pushMeeting()` and `pushMeetingInBackground()` both return at their first line — and `lastError` could never appear either, since `noteError()` sits behind the same early return. The connected branch now renders a fourth state: an amber strip (`connectedCalendars.offTitle` / `offHint`, `{provider}` interpolated from the row label) with **no** Reconnect, because `/api/integrations/google/connect` checks the same switch and can only bounce to `?google=unavailable`; Disconnect stays, since revoking must always be possible. The not-connected branch also now separates "the operator never set this up" (`notConfigured`) from "the operator switched it off" (`notEnabled`) — the second inaccuracy you flagged in the same finding.

**2. (minor) `sanitizeError` mangled provider prose.** Reproduced exactly as reported by running the real module:

```
'Token refresh failed'                               -> '<redacted> failed'
'No usable refresh token — the user must reconnect.' -> 'No usable refresh <redacted> the user must reconnect.'
'invalid_grant: Token has been expired or revoked.'  -> 'invalid_grant: <redacted> been expired or revoked.'
```

Took option (a), tightened rather than remapped: keeping the provider's own words is the point of the strip, and remapping would have to be redone for every future connector. The rule is now two rules — an explicit `:`/`=` redacts whatever follows however short (`password: hunter2`, `api_key=icrm_…`, `client_secret: "…"`), while a whitespace-separated value is redacted only when it looks like a credential (`(?:bearer|token|…)["'\s]+(?=\S*[\d_\-.])[^\s"',;]{12,}`). Verified both directions: `Authorization: Bearer eyJ…`, `token=1//0gLm3x…` and the samples above still redact; the three sentences above now pass through untouched. Known trade, written into the comment: an unquoted six-character space-separated secret would escape this rule — but a sentence beginning "Token" is real and a six-character bearer token is not, and the 32+ opaque-blob rule below still catches real token material. The change benefits every surface that renders connector errors, not just this card.

Tightened the spec as you suggested, on the *rendered* strip: `toContainText('invalid_grant: Token has been expired or revoked.')` plus `not.toContainText('<redacted>')`, so the mangling would now fail the test instead of surviving the `toContain('invalid_grant')` substring.

**3. (minor) `disconnect()` ignored the DELETE response.** Confirmed, including your reachability argument — `middleware.ts:8` puts DELETE in `WRITE_METHODS`, `/api/integrations/google/connection` is not allowlisted, and the connect route is a GET, so an unverified account genuinely can reach "connected" and then be refused the disconnect. It now branches on `res.ok` and flashes a new `connectedCalendars.disconnectFailed` (EN/TR/DE) instead of the success message; on success it **re-fetches** `/api/integrations/calendar/status` rather than patching local state, with the old patch kept only as a fallback if that re-read itself fails. As you put it: the card's claim should come from the server that did the work.

Also folded into this commit: the release fragment's prose updated to describe four states, the disconnect fix and the sanitizer change (still one fragment); the `docs/agent-experience.md` entry gained a follow-up section, including the correction that adding a provider is two lines (the `CalendarProviderId` union as well as the array), not one.

Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` 0 errors (the same two pre-existing `exhaustive-deps` warnings on unrelated admin pages), `npm run check:i18n` OK at 4166 keys × 3 locales, `npm run check:release-fragments` OK (`0.156.21-beta -> 0.157.0-beta`). Schema untouched, so no `prisma validate`. The e2e suite still cannot run in this container (no database, no browser) — CI is the check.

---

PR #2210, opened 2026-09-06
fix: stop reporting a failed delete as a success (#1355)

### LINKED ISSUE #1355: 🐛 [Mentor] Task · Etkileşim silme başarısız olsa da &quot;silindi&quot; bildirimi gösteriliyor
## Amaç

Etkileşim silme isteği 403/500 dönse bile mentöre yeşil "silindi" bildirimi gösteriliyor.
Kayıt ekranda duruyor ama kullanıcı silindiğini sanıyor — sessiz veri kaybı değil, sessiz
**veri kaybı sanrısı**, ki bir CRM'de daha kötüsü.

## Mevcut durum

`src/app/mentor/mentees/[id]/page.tsx:152-160`:

```ts
const confirmDeleteInteraction = async () => {
  if (!deleteInteractionId || deletingInteraction) return;
  setDeletingInteraction(true);
  try {
    await fetch(`/api/interactions/${deleteInteractionId}`, { method: 'DELETE' });
    await fetchRelation();
    toast(t.mentor.interactionDeleted);      // ← res.ok kontrol edilmiyor
  } finally { … }
};
```

`res.ok` hiç okunmuyor, `catch` yok. Karşılaştırma: aynı dosyadaki `handleAddInteraction`
(`:126-133`) ve `handlePipelineChange` (`:166-172`) doğru deseni kullanıyor
(`if (!res.ok) throw new Error(body.error)`).

## Ne yapılacak

1. `res.ok` kontrolü ekle; başarısızlıkta hata bildirimi (`toast(..., 'error')`) göster.
2. Ağ hatası için `catch` ekle.
3. Aynı desen `src/app/mentor/availability/page.tsx:55` (slot silme) için de geçerli — orada da düzelt.
4. Hata metinleri sözlükten (`t.common.error`) gelsin.

## Kabul kriterleri

- [ ] Başarısız silmede kırmızı hata bildirimi görünüyor, "silindi" görünmüyor
- [ ] Başarılı silme davranışı değişmedi
- [ ] `npm run build` yeşil

**Zorluk** · kolay (stajyer dostu) · **Tahmini** · 0.5 gün · Üst: #1349

mersahin:
## Summary

`confirmDeleteInteraction` on the mentee detail page never read the DELETE response. A 403/404/500 therefore refetched the list and fired the green "Interaction deleted" toast while the record was still sitting on the page. That is not silent data loss — it is a silent *illusion* of data loss, which in a CRM is the worse failure: the mentor walks away believing a note is gone.

The handler now checks `res.ok` and shows a red toast carrying a **localized** reason. The list is still refreshed afterwards — but that refresh is deliberately *outside* the guarded request, and swallows its own error, so a refresh that fails can never be narrated as a failed delete (see "After review" below; the first draft got this wrong in the opposite direction).

I then grepped for the same shape across `src/` — a `fetch()` whose response is never read before a success signal — and fixed every instance. Scan method: every statement-position `fetch(`/`await fetch(` with a `toast(` within the following lines, plus every `const res = await fetch(...)` where no `res.ok`/`res.status` check precedes the toast (that second variant turned up nothing).

Closes #1355

## Changes

**Fixed — unchecked response before a success signal**

| Site | What it lied about |
|---|---|
| `src/app/mentor/mentees/[id]/page.tsx` `confirmDeleteInteraction` | the issue's case: green "Interaction deleted" on a rejected DELETE |
| `src/components/NotesPanel.tsx` `confirmRemove` | green "Note deleted" on a rejected DELETE of a portal note |
| `src/components/ContactActions.tsx` `log` | "Contact logged" on a rejected POST — and `onLogged()` refetched a history that never gained the row |
| `src/app/mentor/availability/page.tsx` `confirmRemove` | named in the issue. No toast here, but the silent reload made the deleted slot reappear with no explanation. Now an error box in the *slots* card (not the add-slot form's box, which belongs to a form the user didn't touch) |

All four have the same shape now: guard **only** the request, then refresh outside it.

```ts
let deleted = false, failure = '';
try {
  const res = await fetch(url, { method: 'DELETE' });
  if (res.ok) deleted = true;
  else failure = apiErrorMessage(res, t.common, t.common.deleteFailed);
} catch { failure = t.common.deleteFailed; } finally { /* clear busy state */ }
await refresh().catch(() => {});
if (deleted) toast(success); else toast(failure, 'error');
```

**New**

- `src/lib/apiErrorMessage.ts` — one place for the "what do we tell the user" rule, since four call sites needed it. It maps the **status** to dictionary copy: 401 → `common.sessionExpired`, 403 → `common.forbidden`, 404 → `common.alreadyGone`, anything else (5xx, a dead connection, an unparseable body) → the caller's own localized fallback. The server's `error` string is never rendered — same stance as `lib/authErrors.ts`. A status that carries detail the client genuinely cannot reconstruct is still read at its own call site (the 409 slot overlap that names the interval it collided with).
- i18n `common.deleteFailed`, `common.sessionExpired`, `common.forbidden`, `common.alreadyGone`, `contact.logFailed` (EN/TR/DE).
- `e2e/interaction-delete-failure.spec.ts` — routes a 403 over the DELETE, then asserts the *localized* error toast appears, that the route's English `'Forbidden'` is **not** shown, that the "Interaction deleted" toast does **not** appear, that the row is still on the page and still in the DB. Then it deletes the row out from under the page and clicks delete again, so the real 404 path is covered: the toast says the record is already gone and the row disappears. Finally, a real delete on a second interaction still behaves exactly as before. Untagged (not `@smoke`): this is a P2 correctness fix, and the smoke set is deliberately kept to ~15-20 critical-path tests.
- `data-testid="interaction-list"` / `interaction-delete-<id>` and an `aria-label` on the row's trash button, which previously had no accessible name at all (needed to target it from the spec).
- Release fragment `releases/unreleased/interaction-delete-honest-toast.json`, `patch`, with EN/TR/DE notes (the change is user-visible).

## After review

Self-review of the first commit turned up two substantive problems, both fixed in `79a1a45`:

1. **The refresh sat inside the guarded `try`.** `fetchRelation()` / `load()` have no guard of their own, so they throw on a dropped connection or a non-JSON body (a proxy error page during a container swap). A refresh that threw was therefore reported as *"Could not delete. Nothing was removed."* — after a delete that had already removed the row. The exact lie this PR is about, pointed the other way, and worse than the pre-change behaviour (which produced an unhandled rejection and no toast at all). The refresh now runs after the `try` and swallows its own error.
2. **The reason was the server's English literal.** Every 4xx body on these four routes is a hardcoded English string — `'Unauthorized'`, `'Forbidden'`, `'Not found'`, `'Interaction not found'`, `'Validation failed'` — so passing `body.error` through meant a Turkish mentor whose 12h JWT expired saw **"Unauthorized"** in an otherwise Turkish UI, and the localized keys this PR added were reachable only on 5xx. Step 4 of the issue asks for dictionary text. `apiErrorMessage` now maps status → dictionary key and never renders server-authored text.

Two smaller review points, also fixed:

3. **"Never refetch on failure" left a phantom row.** A 404 means the server does not have the record — deliberately keeping it on screen is the same class of mismatch the fix claims to remove, and every retry 404s again until a manual reload. The refresh now runs on the failure path too, so the list always converges on what the server kept while the toast still says this delete did not succeed.
4. **The availability page's inline delete-error box was sticky** — cleared only at the start of the next delete attempt, with no dismiss control, so it survived a successful add and sat there claiming a delete failed next to a list that had just changed. `load()` now clears it.

**Left alone deliberately**

- **#683, the candidate-history delete** — the task brief flagged it as the identical bug, but it is **already fixed on `main`**: `deleteHistory` in `src/app/admin/candidates/[id]/page.tsx:240` checks `res.ok`, toasts `t.common.error` as an error and has a `catch`. Nothing to do; #683 looks closeable on inspection, which I've left to the maintainer rather than closing an issue assigned to someone else.
- **Unchecked deletes with no success signal** — `RelationNotesPanel`, `ProjectsManager`, `GoogleCalendarCard`, `InterviewPanelView`, `MeetingRequestsPanel`, `DocumentRequirementsAdmin`. These reload from the server and never claim success, so a failure leaves the row visibly present: unhelpful, but not deceptive. That is a different (lower-severity) class than "the app told me it worked", the fix is not uniform across them, and folding six more files in would bury the actual bug. Worth its own issue.
- **Fire-and-forget preference saves** (`AccountSettings` theme/density, `DensityControl`, `FontSizeControl`, `ActivityTracker`) — no user-visible claim is made, out of scope.
- **`/api/notes/[id]` answering 403 for a note that no longer exists.** `ownNote()` returns null both for "not yours" and "gone", and the route reports 403 for both — deliberately fail-closed, so a note id cannot be probed for existence. The consequence is that a note deleted in another tab produces "You are not allowed to do that" rather than "already gone"; the reload now at least makes the row disappear instead of 403-ing forever. Changing the route's status would be an information-disclosure decision, not a toast fix.
- **The rest of the repo still surfaces `body.error` verbatim** (`MyTodos`, `CertificateGenerator` and others). `apiErrorMessage` is now the localized alternative for anyone converting them, but sweeping them is separate work.

## Verification

- [x] `npm run lint` clean (only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`, untouched files)
- [x] `npm run check:i18n` — OK, 4156 keys × 3 locales
- [x] `npm run check:release-fragments` — OK, ships as `0.156.23-beta`
- [ ] `npm run test:e2e` — **not run**: this container has no database and no Docker daemon, so the Playwright suite cannot execute. The new spec is written but unrun; CI is the first execution.
- Dark mode: the new availability error box reuses the exact `bg-red-50 … text-red-700` combination already covered by the `html.dark .bg-red-50.text-red-700` override in `globals.css`, so no new override is needed. Error toasts go through the existing `Toast` component, which is already theme-aware.

Preview: https://pr2210.interncrm.com

## Contributor terms

- [x] I accept the [contributor terms](https://github.com/21072026/internship/blob/main/CONTRIBUTING.md#contributor-terms-ip)


mersahin:
Review findings applied in `79a1a45`. All four were real; none rejected. Verified each against the code before changing anything:

**1 (major) — refresh inside the guarded `try`. Confirmed, fixed.** `fetchRelation` (`src/app/mentor/mentees/[id]/page.tsx:112`) is an unguarded `fetch` + `res.json()`, `NotesPanel.load` and `AvailabilityPage.load` likewise, so all three throw on a dropped connection or a non-JSON body and the `catch` narrated that as "Could not delete. Nothing was removed." after a delete that had already committed. All four call sites (including `ContactActions`, where `onLogged?.()` sat in the same place) now guard only the request and run the refresh afterwards with `.catch(() => {})`.

**2 (major) — `apiErrorMessage` echoing English server text. Confirmed, fixed.** Checked every 4xx the four routes can emit: `/api/interactions/[id]` DELETE → `'Unauthorized'`/`'Interaction not found'`/`'Forbidden'`; `/api/notes/[id]` DELETE → `'Unauthorized'`/`'Forbidden'`; `/api/availability` DELETE → `'Unauthorized'`/`'Not found'`/`'Forbidden'`; `/api/interactions` POST → `'Validation failed'`/`'Mentorship relation not found'`/`'Forbidden'`. All hardcoded English literals, so the helper's passthrough was the only reachable path for the common 401/403/404 cases and the new localized keys only fired on 5xx. The helper is now synchronous and maps status → dictionary copy — 401 `common.sessionExpired`, 403 `common.forbidden`, 404 `common.alreadyGone`, else the caller's fallback — and never reads the body. 401 got its own key rather than being folded into "forbidden": with a 12h JWT an expired session is the likeliest 401, and "sign in again" is actionable where "you are not allowed" is not. The spec now asserts the localized string and additionally asserts the route's `'Forbidden'` is *not* rendered.

**3 (minor) — 404 leaving a phantom row. Confirmed, fixed, and the rule generalised.** Rather than special-casing 404, the refresh now runs on every failure path. With a distinct red toast there is no "the delete went through" ambiguity to protect against, and the list always ends up showing what the server kept. The spec covers it: it deletes the row via Prisma out from under the open page, clicks delete, and asserts the "no longer there" toast plus the row disappearing. The notes route's 403-for-a-missing-row is left as-is — see the PR body for why that status is deliberate; the reload at least stops it 403-ing forever.

**4 (minor) — sticky availability banner. Confirmed, fixed.** `load()` now clears `deleteError`; `confirmRemove` re-sets it after its own reload, so a genuine failure survives while a successful add or reload clears the stale claim. Kept it as an inline box rather than converting to a toast — the page has no `ToastProvider` usage and the add-form already communicates through an inline box directly above.

`npx tsc --noEmit`, `npm run lint`, `npm run check:i18n` (4156 keys × 3 locales) and `npm run check:release-fragments` (`0.156.23-beta`) all green. The release fragment prose and the PR body were rewritten to match the changed approach. Playwright still unrun here — no database in this container.


---

Issue #1560, opened 2026-08-30
🌱 [Infra] Task · `npm run check:tenant-models` — fail CI when an orgId model is not registered


mersahin:
## 🎯 Goal
`TENANT_MODELS` drifted from the schema once and nobody noticed for months — eight models with a tenant key were quietly unprotected. The rule already exists in `CLAUDE.md` ("any new model that holds tenant data needs `orgId` AND an entry in `TENANT_MODELS`"), but a rule that only lives in a doc is a rule that gets skipped in a hurry. Make the repo enforce it, the same way `check:i18n`, `check:demo-blocklist` and `check:query-scalars` already enforce their invariants.

## 🔍 Current state
- `src/lib/orgContext.ts:43-55` — the hand-maintained `ReadonlySet<Prisma.ModelName>` with 11 entries.
- `prisma/schema.prisma` — 19 models declare an `orgId` field.
- There is no check anywhere that compares the two. `.github/workflows/ci.yml:35-60` already runs a stack of small guard scripts: `check:i18n`, `check:release-fragments`, `test:release`, `check:auth-reads`, `check:query-scalars`, `check:demo-blocklist`, `check:k6`.
- Existing scripts to copy the shape from: `scripts/check-demo-blocklist.mjs` and `scripts/check-query-scalars.mjs` (plain Node `.mjs`, read files, print offenders, `process.exit(1)`).

## 🛠 What to build
1. `scripts/check-tenant-models.mjs`:
   - parse `prisma/schema.prisma` for `model X { … }` blocks containing a field named `orgId` (a regex over the block is fine; this repo's schema is conventionally formatted and `prisma format` runs in CI);
   - parse the `TENANT_MODELS` set out of `src/lib/orgContext.ts` (read the quoted names between `new Set([` and `])`);
   - fail with a clear message listing (a) models with `orgId` that are **not** registered and (b) registered names that do not exist in the schema (a typo would otherwise be a silent no-op, since the middleware just never matches).
2. An **allowlist with a reason** for deliberate exceptions — `Organization` itself is not a tenant-scoped row, and `Setting`'s global fallback layer means it must be registered but behaves specially. Keep the allowlist in the script with an inline comment per entry, so an exception is a code review conversation.
3. `package.json`: `"check:tenant-models": "node scripts/check-tenant-models.mjs"`.
[...]

Issue #1551, opened 2026-08-30
🔑 [Infra] Story · Complete the tenant key — orgId on Setting, Webhook, AiUsage and the telemetry tables


mersahin:
## 🎯 Goal
Six tables that hold tenant data have no tenant key at all, so they cannot be isolated by any amount of middleware. One global `Setting` row decides the 2FA policy, the retention window and the AI quota **for every customer**; one global `AiUsage` pool means tenant A can exhaust tenant B's AI budget; one `Webhook` table means every tenant's stage changes are POSTed to every tenant's endpoint; and `ActivityLog`/`PageView`/`EmailLog` mean audit trails and behavioural telemetry cross tenants even after the flag is on. Until each of these carries an `orgId`, "your data is isolated" is not a sentence we can say.

## 🔍 Current state
- `prisma/schema.prisma:1397-1401` — `Setting { key String @id, value String @db.Text, updatedAt }`. One row per key, globally.
- `src/lib/settings.ts:4-62` — 18 global setting keys including `require2fa`, `retentionMonths`, `aiMonthlyQuota`, `premiumAnalytics`, `blindReview`, `selfRegistration`, `newsletterSchedule`.
- `prisma/schema.prisma:1331-1341` `AiUsage`, `:1377-1385` `Webhook`, `:1387-1395` `ApiKey`, `:407-419` `PageView`, `:2412-2432` `ActivityLog`, `:2453-2478` `EmailLog` — none has `orgId`.
- `src/lib/aiGate.ts:24-30` — `getAiQuota()` counts `prisma.aiUsage.count({ where: { createdAt: { gte: monthStart } } })` across the whole database.
- `src/lib/webhooks.ts:29` — `prisma.webhook.findMany({ where: { active: true } })`, no tenant filter.
- `src/app/api/admin/activity/route.ts:12-37` and `src/app/api/admin/email-log/route.ts:16-33` — admin-only, but neither is wrapped in `withTenantScope`.
- `src/app/api/track/pageview/route.ts:30-50` *is* wrapped in `withTenantScope`, but `PageView` is not in `TENANT_MODELS`, so the wrapper does nothing for it.
- `prisma/backfill-organization.mjs:23` — backfills only 6 models (`user, source, cohort, company, project, mentorshipRelation`) of the 19 that carry `orgId`. Run from `infra/deploy-prod.sh:529`.
[...]

Issue #1540, opened 2026-08-30
🧹 [Infra] Task · `prisma/backfills/` registry + one runner, replacing the nine `|| true` lines in deploy-prod.sh


mersahin:
## 🎯 Goal
A schema migration usually needs a data migration next to it, and right now those live as loose scripts hand-appended to a bash file, each suffixed `|| true` so a failure is invisible, with no record of which one has run where. One of them was never added at all and has therefore never run in any environment (#1385). Give data migrations the same discipline the schema is about to get.

## 🔍 Current state
- Hand-listed in the deploy script, all with `|| true`: `infra/deploy-prod.sh:521-541` runs `prisma/seed-templates.mjs`, `prisma/seed-goal-templates.mjs`, `prisma/seed-contributor-terms.mjs`, `prisma/backfill-project-members.mjs`, `prisma/backfill-organization.mjs`, `prisma/backfill-cron-baseline.mjs`, `prisma/backfill-relation-completed-at.mjs`, `prisma/backfill-series-meetings.mjs`. Plus `prisma/backfill-company-interest-scope.mjs` at `:490` and `prisma/backfill-json-columns.mjs --repair` at `:501` and `:518`.
- `scripts/backfill-requisitions.mjs` exists and is wired as `npm run backfill:requisitions` in `package.json`, but appears in **no deploy step** — that is open bug **#1385** ("`backfill:requisitions` hiçbir deploy adımında koşmuyor — eski pozisyonlar hiç dönüşmedi"), and the gap entry for retiring `CompanyNeed` cites the same fact.
- One backfill self-skips by writing a marker row (`prisma/backfill-cron-baseline.mjs`, marker `Setting 'cronBaselineAt'`, noted at `infra/deploy-prod.sh:530-532`) — the only existing "has this run?" record, and it is bespoke.
- The scripts are genuinely idempotent by design (the comments at `infra/deploy-prod.sh:522-541` say so per script), so re-running is safe; what is missing is visibility, ordering and failure surfacing.
- Every deploy runs all of them, on every environment, forever — including one-shot backfills whose work finished months ago.

## 🛠 What to build
1. `prisma/backfills/` with one file per data migration, named `NNNN-<kebab-slug>.mjs`, each exporting `{ id, description, once, run(prisma) }`. Move the existing scripts in unchanged apart from the wrapper; keep their file-top comments, they carry the rationale.
[...]

PR #2234, opened 2026-09-06
fix: backfill every nullable orgId column on deploy (#1557)

### LINKED ISSUE #1557: 🌱 [Infra] Task · Backfill every orgId column on deploy — the backfill fills 6 of 19 models
## 🎯 Goal
The deploy-time backfill that is supposed to give every legacy row a tenant only touches six models. Thirteen others carry an `orgId` and are silently left `NULL`. Under `MT_ENFORCE_ISOLATION=true` the middleware injects `where: { orgId }` into every query on a tenant model — a row with `orgId = NULL` matches nobody and simply **disappears from the product**. A safe rollout needs the backfill to be complete *before* the flag is flipped, not after somebody notices their tags are gone.

## 🔍 Current state
- `prisma/backfill-organization.mjs:23`:
  ```js
  const models = ['user', 'source', 'cohort', 'company', 'project', 'mentorshipRelation'];
  ```
  That is 6 of the 19 models that carry `orgId`.
- Not covered today: `Tag` (`prisma/schema.prisma:143`), `StageSla:176`, `PipelineStage:192`, `DocumentRequirement`, `CompanyInquiry:1200`, `MentorApplication`, `Offer:1278`, `InterviewPanel:1822`, `EvaluationTemplate:1865`, `InvitationToken:2243`, `WeeklyReport`, `Requisition`, `InterviewRequest`.
- The script is idempotent and upserts the `default` org (`prisma/backfill-organization.mjs:16-21`), grandfathered to `ENTERPRISE`.
- It runs on every production deploy from `infra/deploy-prod.sh:529` (`run_tool node prisma/backfill-organization.mjs || true`) — note the `|| true`: a failure is currently invisible. **#1540** is replacing that pattern with a `prisma/backfills/` registry; align with it if it has landed.
- Five of the thirteen (`Tag`, `StageSla`, `PipelineStage`, `EvaluationTemplate`, `DocumentRequirement`, `WeeklyReport`, `Requisition`, `InterviewRequest`) declare `orgId` as **required**, so they cannot hold NULLs and need no backfill — verify this per model rather than assuming.

## 🛠 What to build
1. Derive the model list instead of hard-coding it: read the `orgId`-carrying models from the generated Prisma DMMF (`Prisma.dmmf.datamodel.models`, filtering for a field named `orgId`) so the list can never drift from the schema again. If that proves awkward in a plain `.mjs` script, hard-code the list but add it to the parity check in the sibling task (`npm run check:tenant-models`).
2. Skip models whose `orgId` is non-nullable (nothing to backfill) and log which ones were skipped and why.
3. Print a per-model row count and a final total, and **exit non-zero** when a nullable `orgId` column still has rows left over — so a partial backfill cannot pass silently.
4. Include the newly keyed models from the sibling tasks (`Webhook`, `AiUsage`, `ActivityLog`, `PageView`, `EmailLog`) — but **not** `Setting`, whose legacy rows deliberately stay `orgId = NULL` as the global fallback layer (see the `Setting` task).
5. Add a short note to `docs/tenant-isolation.md` that a complete backfill is step 1 of the rollout checklist.

## ✅ Acceptance criteria
- [ ] Running the script against a DB with NULL `orgId` rows in any nullable tenant column assigns them all to the `default` org
- [ ] Re-running it is a no-op (idempotent) and prints `0 row(s) assigned`
- [ ] The script exits non-zero if any nullable `orgId` column still holds NULLs after the pass
- [ ] `Setting` is explicitly excluded, with the reason in a comment
- [ ] The script's model list cannot drift from `prisma/schema.prisma`
- [ ] `npm run build` passes
- [ ] a release fragment `releases/unreleased/backfill-all-org-columns.json` is added

## 🇹🇷 Türkçe özet
Deploy sırasında çalışan `backfill-organization.mjs` yalnızca 6 modeli dolduruyor; `orgId` taşıyan diğer 13 model boş kalıyor. Bayrak açıldığında middleware her sorguya `where: { orgId }` eklediği için `orgId = NULL` olan satırlar hiçbir kiracıya eşleşmez ve ürün içinde **kaybolur**. Bu görev backfill'i şemadan türetip eksiksiz hale getiriyor, eksik satır kalırsa sıfırdan farklı çıkışla hata veriyor. `Setting` bilerek dışarıda: onun eski satırları global varsayılan katmanı olarak `NULL` kalacak. Küçük, izole ve iyi tanımlı — stajyer için uygun.

## 🤖 Agent prompt
```
Objective: make the deploy-time org backfill cover every orgId column instead of
six of them.

Open first:
  prisma/backfill-organization.mjs   (the 6-model list is on line 23)
  prisma/schema.prisma               (grep -n "  orgId" -> 19 models)
  infra/deploy-prod.sh:529           (how it is invoked, with `|| true`)
  docs/tenant-isolation.md           (rollout checklist step 1)

Constraints:
- Node .mjs script, plain PrismaClient, no framework. Keep it idempotent.
- Skip models whose orgId is NOT nullable — they cannot hold NULLs.
- EXCLUDE Setting: its legacy rows must stay orgId = NULL, that is the global
  fallback layer. Put the reason in a comment.
- Exit non-zero when a nullable orgId column still has rows after the pass, so a
  partial backfill cannot pass silently (today the deploy calls it with
  `|| true` and swallows failures; #1540 is fixing that pattern — align if it
  has landed).
- Prefer deriving the model list from Prisma.dmmf.datamodel.models so it cannot
  drift from the schema.
- Do not flip MT_ENFORCE_ISOLATION anywhere in this task.

Test locally: npx prisma db push && npx prisma db seed && npm run seed:demo,
null out some orgIds by hand, run the script, confirm counts and idempotency.

Done when: every nullable orgId column is filled, a re-run reports 0, and a
leftover NULL fails the script.
```

**Zorluk** · S · **Tahmini** · 0.5 day · **Bağımlılık** · #1540 (backfill registry), #1551

mersahin:
## Summary

The deploy-time tenant backfill (`prisma/backfill-organization.mjs`) hardcoded six models — `user`, `source`, `cohort`, `company`, `project`, `mentorshipRelation` — while **23 models in `prisma/schema.prisma` carry an `orgId`**. Every nullable column outside that list was left `NULL` on every deploy. That is not cosmetic: once `MT_ENFORCE_ISOLATION=true`, the Prisma middleware in `src/lib/orgContext.ts` injects `where: { orgId }` into every query on a tenant model, so a row with `orgId = NULL` matches nobody and **disappears from the product**. A complete backfill has to be green *before* that flag is flipped, not after somebody notices their API keys or invitations are gone.

The root cause is the hardcoded list, so the fix is to stop maintaining one: the models are now derived from the generated Prisma DMMF, and the script fails loudly instead of reporting a partial pass as success.

Derived from the schema as it stands today (verified against the generated client, not the issue's list):

- **15 nullable → backfilled**: `AccountLockout`, `ApiKey`, `Cohort`, `Company`, `CompanyInquiry`, `InterviewPanel`, `InvitationToken`, `MatchFeedback`, `MentorApplication`, `MentorshipRelation`, `MessageTemplate`, `Offer`, `Project`, `Source`, `User` — nine of these were never touched before.
- **8 `NOT NULL` → skipped and logged**: `DocumentRequirement`, `EvaluationTemplate`, `InterviewRequest`, `PipelineStage`, `Requisition`, `StageSla`, `Tag`, `WeeklyReport`. They cannot hold NULLs, so there is nothing to fill. (The issue listed five of these; the schema says eight — the script reads the schema, so the count follows it automatically.)
- **`Setting` → excluded on purpose, pre-emptively.** It has **no `orgId` column yet** — #1551 adds one — and the exclusion is keyed by model name so it takes effect the moment that column lands. Its legacy rows must then stay `orgId = NULL`: NULL *is* the global fallback layer, a setting with no org applies to every tenant and a per-org row overrides it. Assigning those rows to the default org would turn platform-wide defaults into one tenant's private settings and leave every other tenant unconfigured. The reason sits next to the entry so it does not get "fixed" later.

Closes #1557

## Changes

- **`prisma/backfill-organization.mjs`**
  - Model list derived from `Prisma.dmmf.datamodel.models`, filtering for a scalar field named `orgId`, so it can never drift from `prisma/schema.prisma` again. Nullability comes from the same place (`field.isRequired === false`), not from an assumption. Newly keyed models from the sibling tasks (#1551: `Webhook`, `AiUsage`, `ActivityLog`, `PageView`, `EmailLog`) are picked up automatically the moment their `orgId` lands — no follow-up edit here.
  - Non-nullable models are skipped with a one-line reason in the log; excluded models are logged with their reason.
  - **The pass is raw SQL, not `updateMany`.** Prisma stamps `@updatedAt` on `updateMany`, and five of the newly covered models declare one (`AccountLockout`, `InterviewPanel`, `MatchFeedback`, `MessageTemplate`, `Offer`) — `Offer.updatedAt` is returned by both `/api/offers` handlers. A bookkeeping backfill must not rewrite the last-modified time of every legacy row, irreversibly, and make it look like an edit. Table and column names come from the DMMF (`model.dbName ?? model.name`) and are asserted against `/^[A-Za-z0-9_]+$/` before being backticked, the same shape `prisma/backfill-json-columns.mjs` already uses.
  - **Verification is a retry loop, not a single re-count.** "Are there NULLs right now?" is not a signal about the backfill: during a prod deploy the old container is still serving (backfills are section 4 of `infra/deploy-prod.sh`, the container swap section 5), and three ordinary paths insert `orgId = NULL` with no tenant context — a failed sign-in against an unknown e-mail (`src/lib/accountLockout.ts`), the public company-inquiry form and the public mentor-application form. One of those landing between the `UPDATE` and the `COUNT` would have failed the check on a perfectly healthy database. The script now runs up to three `UPDATE`+`COUNT` cycles and **exits non-zero only when a row survives all of them** — a row the backfill genuinely cannot fill — while the concurrent inserts get filled by the retry instead of being left behind.
  - The pass itself is exported as `assignDefaultOrg(prisma, orgId)` behind a direct-run guard, so importing the module does not execute it.
  - Still plain Node + `PrismaClient`, still idempotent: a second run assigns 0 rows.
- **`prisma/seed.mjs` / `prisma/seed-demo.mjs`** — both carried their own copy of the same six-model loop, which is the drift #1557 is about. Both now call `assignDefaultOrg`, so a freshly seeded database is complete on all 15 models.
- **`infra/server/topic-deploy.sh` / `infra/server/demo-refresh.sh`** — neither ran the backfill at all, and a per-PR topic environment is the only place the doc's own "verify in a preview environment first" instruction can actually be carried out. Both now run it after the seeds (which create fresh org-less rows), non-fatally.
- **`infra/deploy-prod.sh`** — the invocation keeps `|| true` (a deploy must not abort on a backfill step, and #1540 is replacing that whole pattern with a `prisma/backfills/` registry — it has **not** landed, so nothing to align with yet), but the failure is no longer swallowed. The warning describes the **exit code**, not the database: the script also exits non-zero when the DB is unreachable (`deploy-preview.yml` documents exactly that case) or the run never got as far as counting anything, so asserting "rows still have `orgId = NULL`" would be a tenancy alarm for a networking problem. The legacy `deploy.yml` pipeline is superseded and deliberately left alone.
- **`docs/tenant-isolation.md`** — rollout step 1 rewritten: a complete backfill is the precondition for flipping the flag. It names the environments that run the script rather than claiming "every deploy" in the abstract, records the two omissions (NOT NULL columns, `Setting`), and notes that non-NULL is not the same as *correct* once real tenants exist.
- **`releases/unreleased/backfill-all-org-columns.json`** — `patch`, no user-facing notes (infra/deploy-time change; nothing visible in the UI today, since the flag is off and the app is effectively single-tenant).

## Verification

- [x] `npm run lint` clean (only the pre-existing `react-hooks/exhaustive-deps` warnings)
- [x] `npm run check:i18n` — OK, 4231 keys × 3 locales (no new strings; the script's output is developer-facing log text)
- [x] `npm run check:release-fragments` — fragment valid
- [x] `node --check` on all three touched `.mjs` files, `bash -n` on all three touched shell scripts
- [x] DMMF derivation verified against the **generated client** in this container (throwaway probe, removed before committing): 23 models carry `orgId`, 15 nullable / 8 required, every derived table/column resolves, and importing the module runs nothing.
- [ ] **Not executed against a database.** This container has no MySQL and no Docker daemon, so the script itself was never run — the model partition is reasoned from the schema and confirmed via the DMMF probe above, but the actual `UPDATE`/`COUNT` pass is unverified. No e2e spec is added: this is a deploy-time script with no UI surface, and exercising it needs a real DB.

**A reviewer should run, on a local/topic DB (the sequence from the issue):**

```bash
npx prisma db push && npx prisma db seed && npm run seed:demo
# null out some orgIds by hand, e.g.:
#   UPDATE ApiKey SET orgId = NULL; UPDATE User SET orgId = NULL;
node prisma/backfill-organization.mjs          # assigns them, prints per-model counts
node prisma/backfill-organization.mjs          # idempotent: "0 row(s) assigned."
echo $?                                        # 0
```

Worth checking specifically, since it is what the review changed: pick a row on `Offer` or `MessageTemplate`, note its `updatedAt`, `UPDATE … SET orgId = NULL`, run the script, and confirm `updatedAt` is unchanged.

## Notes / out of scope

- `MT_ENFORCE_ISOLATION` is **not** touched anywhere in this PR.
- `npm run check:tenant-models` (#1560) is being built in parallel by another agent; this PR does not create it. There is no list left for it to police in this script — the list *is* the schema — but the exclusion set is a single named `EXCLUDED` map at the top of the file if that guard wants to read it.
- Adjacent, not fixed here: `docs/tenant-isolation.md` still describes the middleware as scoping six models (`User`, `Source`, `Company`, `Project`, `Cohort`, `MentorshipRelation`) while `TENANT_MODELS` in `src/lib/orgContext.ts` now lists fourteen. Stale doc, unrelated to the backfill.

Preview: https://pr2234.interncrm.com

## Contributor terms


mersahin:
Review follow-ups applied in 29da441. All five findings were reproduced against the code and fixed; none was rejected, but one fix deviates from the suggested approach and one extra inaccuracy turned up while verifying.

**1 (major) — verification races with live traffic. Fixed, differently than suggested.**
Confirmed all three NULL-writing paths: `src/lib/auth.ts` → `recordFailedAttempt({ orgId: user?.orgId ?? null })` → `src/lib/accountLockout.ts` upsert; `POST /api/company-inquiry`; `POST /api/mentor-applications`. And the backfills are section 4 of `infra/deploy-prod.sh` while the container swap is section 5, so the old container really is serving throughout.

The suggested fix (`createdAt: { lt: startedAt }`) does not work for the full target set: **`MentorshipRelation` has no `createdAt` column** — it is the one nullable-`orgId` model without one, so that filter would either throw or silently drop it from verification. It also leaves the concurrently-inserted NULL rows behind unfixed until the next deploy. Instead the pass now runs up to three `UPDATE`+`COUNT` cycles: a row that survives all of them is one the backfill genuinely cannot fill, and rows inserted mid-window get filled by the retry rather than failing the run. Same guarantee, no timestamp dependency, and the data ends up complete.

**2 (minor) — warning asserts an unmeasured state. Fixed** — reworded to describe the exit code, with an explicit note in the surrounding comment that the script also exits non-zero when the DB is unreachable or the generated client is stale.

**3 (minor) — topic/demo envs never run this. Fixed at both ends.** `infra/server/topic-deploy.sh` (after the seed block, so it covers both the fresh-DB and existing-DB branches) and `infra/server/demo-refresh.sh` now run the script, and the duplicated six-model loops in `prisma/seed.mjs:56` and `prisma/seed-demo.mjs:344` are gone — both now call the same DMMF-derived pass (`assignDefaultOrg`), which is exported from `prisma/backfill-organization.mjs` behind a direct-run guard so importing it does not execute the backfill. The doc now names the environments instead of claiming "every deploy" in the abstract.

**4 (minor) — release fragment prose. Fixed.** Rewritten: six hardcoded models, 23 carry `orgId`, the **nine** other nullable ones stayed NULL. Header comment `6 of 19` → `6 of 23`.

**5 (minor) — `@updatedAt` rewrite. Fixed.** Verified from the DMMF that `AccountLockout`, `InterviewPanel`, `MatchFeedback`, `MessageTemplate` and `Offer` all declare `@updatedAt` among the newly covered models, and that `Offer.updatedAt` is selected by both `/api/offers` handlers. The pass is now `$executeRawUnsafe('UPDATE \`<table>\` SET \`<column>\` = ? WHERE \`<column>\` IS NULL')` with table/column taken from `model.dbName ?? model.name` and asserted against `/^[A-Za-z0-9_]+$/` — same pattern `prisma/backfill-json-columns.mjs` already uses. Prisma no longer touches `updatedAt`. The delegate-exists guard went away with it: raw SQL does not go through a client delegate, and a missing table now surfaces as a plain MySQL error.

**Extra, found while verifying:** `Setting` has **no `orgId` column at all**, so the `EXCLUDED` entry for it matched nothing and the doc's "two deliberate omissions" was one. The exclusion mechanism stays (with the reason-per-entry rule documented) but the map is empty, and the doc no longer claims `Setting` is skipped.

Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings), `npm run check:i18n` OK (4231 keys × 3 locales), `npm run check:release-fragments` exit 0. Schema untouched, so no `prisma validate` needed. No database or Docker daemon in this container, so the pass itself is still unexecuted here — it wants a look on the PR environment.

---

mersahin:
Correction to the previous comment — the `Setting` exclusion is back in acb5f41.

I removed it as dead code after finding `Setting` has no `orgId` column, then read the PR body and #1551: the entry is **deliberately pre-emptive**, keyed by model name so the backfill does not assign the default org to the global fallback rows the moment #1551 adds the column. Removing it would have made a future schema change silently destructive, which is exactly the failure mode the entry exists to prevent. Restored, with the comment now saying it is pre-emptive rather than implying the rows already exist, and `docs/tenant-isolation.md` says the same. Everything else from the previous comment stands.

Static checks green again after the correction: `tsc --noEmit`, `lint`, `check:i18n` (4231 × 3), `check:release-fragments`.

---

Issue #1559, opened 2026-08-30
🧱 [Trust] Task · Register the 8 uncovered orgId models in TENANT_MODELS and audit their create paths


mersahin:
## 🎯 Goal
Eight models carry a tenant key but are not registered with the auto-scoping middleware, so today they are protected only by whatever `where` clause a developer remembered to write. Two of them are the ones a buyer will actually notice: `Offer` (a candidate's compensation note) and `InvitationToken` (a live registration link with a role attached). Registering them turns "we hope every query filters" into "the database cannot return another tenant's row".

## 🔍 Current state
- `src/lib/orgContext.ts:43-55` — `TENANT_MODELS` = `User, Source, Company, Project, Cohort, MentorshipRelation, MentorApplication, DocumentRequirement, WeeklyReport, Requisition, InterviewRequest` (11).
- Missing, all carrying `orgId`:
  | Model | schema line | `orgId` nullability | notes |
  |---|---|---|---|
  | `Tag` | `prisma/schema.prisma:143` | **required** | `@@unique([orgId, name])` already |
  | `StageSla` | `:176` | **required** | `@@unique([orgId, stageKey])` |
  | `PipelineStage` | `:192` | **required** | `@@unique([orgId, key])` |
  | `CompanyInquiry` | `:1200` | nullable | written by the **public** `/for-companies` form |
  | `Offer` | `:1278` | nullable | holds `compensationNote` |
  | `InterviewPanel` | `:1822` | nullable | `@@index([orgId, subjectId])` |
  | `EvaluationTemplate` | `:1865` | **required** | per-tenant competency framework |
  | `InvitationToken` | `:2243` | nullable | `orgId` captured at invite time (#1272) |
- `src/lib/orgContext.ts:96` — an unregistered model is a **silent** pass-through.
- `src/lib/orgContext.ts:99-125` — registration also affects writes: `create`/`createMany`/`upsert` get `orgId` injected from the bound context when the caller did not set it.
- Routes that write these models without a session context: `src/app/api/register/route.ts:82,164,178` and `src/app/api/invite/opened/route.ts:19-23` (token-based, no session, deliberately unwrapped per `docs/tenant-isolation.md`), plus the public company-inquiry form.

## 🛠 What to build
1. Add all eight names to `TENANT_MODELS` in `src/lib/orgContext.ts:43`.
2. For each, walk its call sites (`grep -rn "prisma\.<model>\." src/`) and confirm:
[...]

PR #2235, opened 2026-09-06
ci: fail the build when an orgId model is not registered with the tenant middleware (#1560)

### LINKED ISSUE #1560: 🌱 [Infra] Task · `npm run check:tenant-models` — fail CI when an orgId model is not registered

mersahin:
## Summary

`TENANT_MODELS` in `src/lib/orgContext.ts` is the list the Prisma `$use` middleware consults before it injects the current request's org into a query. A model that is **missing** from that set is a silent pass-through: nothing throws, nothing logs, and a row carrying an `orgId` column only *looks* isolated while it is protected by whatever `where` clause the last developer remembered to write. That is exactly how the set drifted from the schema once and eight tenant-keyed models stayed unprotected for months. The rule already lived in `CLAUDE.md`; this PR makes the repo enforce it, the same way `check:i18n`, `check:demo-blocklist` and `check:query-scalars` enforce theirs.

`scripts/check-tenant-models.mjs` reads the `orgId`-bearing models out of `prisma/schema.prisma` and the registered names out of `orgContext.ts` **as text** — no import of the server-only TS module (it pulls in the Prisma client and `node:async_hooks`), no new dependency — and fails in **both** directions:

* a model that declares `orgId` and is not registered → unenforced tenant data;
* a name in `TENANT_MODELS` that is not a model, or is a model that no longer has the column → a typo or a rename, which the middleware also ignores in silence.

Closes #1560

### What this check does *not* claim yet

The eight models of #1559 (`Tag`, `StageSla`, `PipelineStage`, `CompanyInquiry`, `Offer`, `InterviewPanel`, `EvaluationTemplate`, `InvitationToken`) carry `orgId` and are unprotected **today**. Registering them changes runtime behaviour — the middleware starts injecting `orgId` into their *creates* too, and the sessionless write paths (`/api/register`, invite acceptance, the public company-inquiry form) need an audit, plus a cross-tenant spec run against a real DB. That is #1559's reviewed piece of work, and this PR deliberately does not do it.

The issue notes the check "will fail on main until #1559 registers the eight models". Shipping a hard failure would turn main red for every unrelated PR until then, so the eight sit in a **`PENDING_REGISTRATION` ratchet** instead. A first revision of this PR let that ratchet print `tenant models OK`, which is precisely the wrong answer for the drift #1560 was filed about; it no longer can:

* every run prints each pending model **on its own line with its own reason** (the previous version joined the names and appended only the *first* entry's reason, so a second entry with a different blocker would have been misattributed);
* in CI the same list is emitted as a `::warning` **annotation** on `src/lib/orgContext.ts`, so the unprotected set shows on the PR page instead of only inside a green step's log;
* while anything is pending the summary line **drops the word "OK"** — it reads `tenant models: no NEW drift, but 8 model(s) remain unprotected (#1559)`;
* the list length is pinned by an `EXPECTED_PENDING` literal, so a ninth unprotected model **cannot** be waved through by appending a line — the number has to move in the same diff, in front of a reviewer;
* an entry that has since been registered, or whose model lost the column, **fails** the check until it is deleted, so #1559 cannot land without clearing its own entries.

So while #1559 is open, this guard's job is to stop the unprotected set from *growing* and to name it on every run.

### `EXEMPT`: one entry, and no pre-granted exemption for `Setting`

`EXEMPT` holds **`Organization`** only — it *is* the tenant, keyed by its own `id`, and can never grow an `orgId`, so the entry is permanent documentation rather than a standing decision. Because it carries no column, the summary reports it as *declared but not applicable* rather than counting it as an active exemption.

`Setting` is **not** exempted, though an earlier revision of this PR exempted it. `model Setting` has no `orgId` field at all today (`key`, `value`, `updatedAt`) — #1551 adds one — so the entry asserted a schema shape that does not exist, and since `EXEMPT` is consulted before anything else it would have kept this guard green on the day the column landed, silently skipping the register-vs-exempt decision. #1560 itself says `Setting` "must be registered but behaves specially", while #1557 says its legacy rows will stay `orgId = NULL` as the global fallback layer; those two have to be reconciled against the real column, not pre-decided here. Verified: with `orgId String?` added to `model Setting`, the check now exits 1 and asks for that decision.

## Changes

- `scripts/check-tenant-models.mjs` — new guard (plain Node `.mjs`, shaped after `check-demo-blocklist.mjs` / `check-query-scalars.mjs`: read files, print offenders, `process.exit(1)`).
- `package.json` — `"check:tenant-models": "node scripts/check-tenant-models.mjs"`.
- `.github/workflows/ci.yml` — the step next to the sibling guards, with a comment naming the failure mode it prevents.
- `docs/tenant-isolation.md` — new "Keeping the registry honest" section, covering both escape hatches and the `Setting` decision.
- `CLAUDE.md` — a multi-tenancy bullet stating the rule, and saying plainly that the check is not green-means-clean while the #1559 ratchet is populated.
- **No release fragment.** AC 7 of #1560 says a pure CI change needs none, and `CLAUDE.md` agrees ("trivial non-user-facing changes (pure docs, CI config) still need no fragment"). The diff touches nothing under `src/`, so under the one-fragment-one-release rule (#1457) a fragment would mint a version and a CHANGELOG entry for something neither a user nor the maintainer can observe in the running app. An earlier revision included one; it has been removed.

## Proof that it fails on drift

Current output (`npm run check:tenant-models`, exit 0 — note the absent "OK"):

```
tenant models: 8 model(s) carry orgId and are NOT auto-scoped yet — their rows are protected only by hand-written `where` clauses:
  • Tag — awaiting registration + cross-tenant spec in #1559
  • StageSla — awaiting registration + cross-tenant spec in #1559
  … (six more)
tenant models: no NEW drift, but 8 model(s) remain unprotected (#1559) — 23 model(s) declare orgId; 15 registered in TENANT_MODELS, 8 pending (#1559), 0 exempt by declared exception (1 further exemption(s) declared but not applicable — Organization carries no orgId column); every registered name resolves to a model that still has the column.
```

Under `GITHUB_ACTIONS=1` the same list is additionally emitted as `::warning file=src/lib/orgContext.ts,title=Unprotected tenant models::…`.

**A: a new model with `orgId`, registry untouched** — exit 1:

```
  • Widget declares `orgId String` in prisma/schema.prisma but is not in TENANT_MODELS (src/lib/orgContext.ts). The middleware ignores unregistered models in silence, so its rows are scoped only by hand-written `where` clauses. Add 'Widget' to TENANT_MODELS — or, if it must stay unscoped, add it to EXEMPT in this script with the reason.
```

**B: `Setting` gains `orgId` (simulating #1551)** — exit 1, the decision is forced:

```
  • Setting declares `orgId String?` in prisma/schema.prisma but is not in TENANT_MODELS (src/lib/orgContext.ts). …
```

**C: a typo in the set** (`'MentorApplication'` → `'MentorAplication'`) — exit 1, both directions reported:

```
  • MentorApplication declares `orgId String?` in prisma/schema.prisma but is not in TENANT_MODELS …
  • TENANT_MODELS lists 'MentorAplication', which is not a model in prisma/schema.prisma. The middleware matches on the Prisma model name, so a name that does not exist never matches anything — check the spelling, or drop the entry if the model was removed.
```

**D: a ninth name appended to the pending ratchet** — exit 1, the pin holds:

```
  • PENDING_REGISTRATION holds 9 entries, but EXPECTED_PENDING in this script says 8. The set of knowingly unprotected models is pinned to a literal so it cannot change quietly: update EXPECTED_PENDING in the same diff and say in the PR why a model was added to — or removed from — the unprotected list.
```

**E: a pending entry that has since been registered** (`'Offer'` added to `TENANT_MODELS`) — exit 1:

```
  • 'Offer' is in PENDING_REGISTRATION in this script AND in TENANT_MODELS. It is registered, so remove it from PENDING_REGISTRATION — a stale entry hides the next real omission.
```

Every mutation above was made in a throwaway copy of the tree; the working tree contains none of them.

## Verification

- [x] `npx tsc --noEmit` clean; `npm run lint` exit 0 (only the pre-existing `react-hooks/exhaustive-deps` warnings in `src/app/admin/companies/page.tsx` and `src/app/admin/mentorship/page.tsx`)
- [x] `npm run check:i18n` (4231 keys × 3 locales) and `npm run check:release-fragments` pass — the latter still emits its pre-existing "compaction looks stuck" notice, which is unrelated to this PR and one fragment lighter after the removal
- [x] `npm run check:tenant-models` green on this branch, and red in each of the five drift scenarios above
- [ ] `npm run test:e2e` — not run: no database, no Docker daemon in this environment. No e2e spec is added; the issue scopes the cross-tenant spec to #1559.

No schema change, no i18n keys, no runtime code touched — nothing in `src/` changes behaviour.

Preview: https://pr2235.interncrm.com

## Contributor terms

- [x] I accept the [contributor terms](https://github.com/21072026/internship/blob/main/CONTRIBUTING.md#contributor-terms-ip)


mersahin:
## Review follow-up — all four findings addressed (10b7003)

I reproduced each finding against the code before touching anything. All four were real; none were rejected. One was fixed via the second of the two remedies the review itself offered, and I want to be explicit about why.

**1. (major) The ratchet shipped green for exactly the drift #1560 was filed about.** Confirmed — the old run ended in `tenant models OK` with exit 0 while eight `orgId` models sat unregistered, and the only signal was a `console.warn` in a step GitHub renders green.

The review offered two remedies: land #1559 here, or make the ratchet visible and self-limiting. **I took the second**, deliberately. #1559 is not a list edit: registering a model changes *create* behaviour (the middleware injects `orgId` where the caller left it `undefined`), which touches the sessionless write paths — `/api/register`'s token lookup, invite acceptance, the public company-inquiry submit — and four of the eight have a **required** `orgId`, so an unwrapped create that stops passing one explicitly breaks at runtime, not at compile time. #1559's own acceptance criteria demand a direct-Prisma spec proving cross-tenant reads return nothing for each of the eight, plus an e2e spec. This environment has no database and no Docker daemon, so I can verify none of that. Landing an unverifiable runtime change to tenant isolation inside a CI-tooling PR is a worse outcome than a loud ratchet, and it would also make this PR unreviewable as the "does the guard work" change it is.

What changed instead, so the ratchet cannot be mistaken for a clean tree:

* per-model lines, each with **its own** reason (this also fixes finding 4);
* a `::warning file=src/lib/orgContext.ts,title=Unprotected tenant models::…` annotation under `GITHUB_ACTIONS`, so the set appears on the PR page rather than only in the step log — same pattern as `check-release-fragments.mjs`;
* the summary drops **"OK"** while anything is pending: `tenant models: no NEW drift, but 8 model(s) remain unprotected (#1559)`;
* `EXPECTED_PENDING = 8` pins the list length, so appending a ninth name fails until the literal moves in the same diff (verified: exit 1 with a message saying so). The review was right that "adding a new name here is not a way to pass CI" was, until now, an unenforced comment.

The net effect: while #1559 is open this guard's job is to stop the unprotected set from *growing* and to name it every run — and the docs now say that in those words rather than implying the registry is clean.

**2. (minor) The `Setting` exemption asserted a column that does not exist.** Confirmed — `model Setting` is `key` / `value` / `updatedAt`, no `orgId`, which is why the script's own summary read `0 exempt by declared exception`. Worse than the wrong tense in the doc was the mechanism: `EXEMPT` is consulted first, so the entry would have kept the guard green the day #1551 adds the column, and the register-vs-exempt decision #1560 explicitly wants made would never have been asked for. **`Setting` is removed from `EXEMPT`.** With `orgId String?` spliced into the model, the check now exits 1 and demands the decision. `Organization` stays — it is the tenant and can never grow an `orgId`, so it is documentation, not a standing decision about an unseen shape — but the summary now reports it as *declared but not applicable* instead of counting it, and `docs/tenant-isolation.md` states the `Setting` position in the future tense with both #1551 and #1557 named.

**3. (minor) Release fragment contradicts AC 7.** Confirmed and **deleted**. AC 7 says "no release fragment needed (pure CI change) — state that in the PR"; `CLAUDE.md` says the same for CI config. The diff touches only `scripts/`, `.github/`, `docs/`, `CLAUDE.md` and `package.json`'s scripts block, so under #1457 the fragment would have minted its own version and a CHANGELOG paragraph for something nothing in the running app reflects. Stated in the PR body under Changes.

**4. (minor) Wrong reason attributed to the whole pending list.** Confirmed — line 200 interpolated `PENDING_REGISTRATION.get(pending[0])` after joining every name. Fixed by printing one line per entry. I verified the failure mode the review predicted by adding a ninth entry with a different blocker: before, its reason was reported as "#1559"; now each prints its own.

**Static checks after the fixes:** `npx tsc --noEmit` clean, `npm run lint` exit 0 (only the two pre-existing `react-hooks/exhaustive-deps` warnings), `npm run check:i18n` OK (4231 keys × 3 locales), `npm run check:release-fragments` exit 0, `npm run check:tenant-models` exit 0 with the honest summary. `prisma validate` not re-run — the schema is untouched. No e2e run: no DB, no Docker daemon here.

---

Issue #1624, opened 2026-08-30
🪜 [Pipeline] Task · Re-key PipelineStage and StageSla on programId, with the org row as default


mersahin:
## 🎯 Goal
Make the stage set and the stage SLAs belong to a programme rather than to the whole tenant, so an internship pipeline and a leadership cohort can have different stages without one rewriting the other. The org-level set stays as the fallback, so nothing changes for an organisation that never configures a programme.

## 🔍 Current state
- `model StageSla { id, orgId, stageKey, days, ... } @@unique([orgId, stageKey]) @@index([orgId])` — `prisma/schema.prisma:176-190`.
- `model PipelineStage { id, orgId, key, label, order, isTerminal, isOffPath, color, ... } @@unique([orgId, key]) @@index([orgId, order])` — `prisma/schema.prisma:192-207`.
- `resolvePipelineStages(orgId, locale)` → org rows, else `defaultPipelineStages(locale)` — `src/lib/pipelineStages.ts:18-37`; `resolveCustomStages(orgId)` at `:41-56`.
- `resolveStageSlas(orgId)` → `Map<stageKey, days>` — `src/lib/stageSla.ts:24-30`; consumed by `stageDeadlineUpdate` via `src/lib/stageChangeEffects.ts:7`.
- Write API replaces the whole set per org: `src/app/api/admin/organizations/[id]/pipeline-stages/route.ts` (PUT deletes then createMany, DELETE resets); SLA write at `src/app/api/admin/stage-sla/route.ts:84-87`.
- Neither `PipelineStage` nor `StageSla` is in `TENANT_MODELS` — they rely on hand-written `orgId` filters (`src/lib/orgContext.ts:43-55`); #1559 tracks registering them.

## 🛠 What to build
1. Add `programId String @default("")` to **both** models — **not** a nullable column. Empty string is the sentinel meaning "the org-level default set".
   > **Why not `String?`:** MySQL treats NULLs as distinct inside a unique index, so `@@unique([orgId, programId, key])` with a nullable `programId` would stop preventing duplicate org-level rows — the exact constraint that keeps the stage editor's replace-the-set write correct. A non-null sentinel keeps the uniqueness real.
2. Replace the uniques: `@@unique([orgId, programId, key])` on `PipelineStage`, `@@unique([orgId, programId, stageKey])` on `StageSla`. Update the indexes to `@@index([orgId, programId, order])` / `@@index([orgId, programId])`.
[...]

Issue #51, opened 2026-06-27
admin panel mobile uyumluluk 


mersahin:
![Screenshot_2026-06-27-08-46-13-775_com.android.chrome.jpg](https://github.com/user-attachments/assets/755f2de6-9775-4524-a4a1-e5849171995d)

Telefondan girince güzel kullanilabilmeli menü gizlenmesi vs best practice uygula

PR #2236, opened 2026-09-06
fix: start new relations on the tenant's first on-path stage (#1634)

### LINKED ISSUE #1634: 🐛🌱 [Pipeline] Task · New relations start on the programme's first on-path stage, not APPLICATION_100
## 🎯 Goal
On a tenant that customised its pipeline, **every relation ever created is invisible**. The schema default is the canonical key `APPLICATION_100`; a tenant whose stage set does not contain that key gets a relation parked on a stage that appears in no column of its own board and in no row of its own funnel. Small fix, but it makes the flagship white-label feature unusable on day one.

## 🔍 Current state
- `pipelineStatus String @default("APPLICATION_100")` — `prisma/schema.prisma:1696-1700`.
- `POST /api/mentorship` takes `{ mentorId, menteeId, companyId?, projectId?, startDate? }` and no stage at all — `src/app/api/mentorship/route.ts:17-23`; the create happens at `:124-190` and simply lets the DB default apply.
- The tenant's real set is already resolvable: `resolvePipelineStages(orgId, locale)` — `src/lib/pipelineStages.ts:18-37` — and `onPathKeys(stages)` returns the happy-path keys in order — `src/lib/pipeline.ts:236-238`.
- The board renders the resolved set, so a relation on an unresolved key has no column to live in — `groupResolvedStages`, `src/lib/pipeline.ts:135-147`.

## 🛠 What to build
1. Add `startStageKey(stages: ResolvedStage[]): string` to `src/lib/pipeline.ts` — the first entry of `onPathKeys(stages)`, falling back to `'APPLICATION_100'` when the set is somehow empty. Pure function, no Prisma, so it stays client-safe.
2. In `POST /api/mentorship` (`src/app/api/mentorship/route.ts`), resolve the stages for the relation's programme/org (`resolvePipelineStages`, three-level after #1624) and pass `pipelineStatus: startStageKey(stages)` explicitly on create.
3. Do the same in the other two paths that create a relation: the mentorship-request acceptance in `src/lib/mentorshipDecision.ts`, and the CSV import under `src/app/api/admin/import/`.
4. Leave the schema default in place — it is the right value for the canonical set and removing it would be a schema change for no gain.
5. Add a one-off idempotent repair in `prisma/backfills/` (registry from #1540): relations sitting on a key that does not exist in their resolved set move to that set's first on-path stage, writing a `StatusChange` row so the move is auditable rather than silent.

## ✅ Acceptance criteria
- [ ] On an org whose custom stage set starts with e.g. `SOURCED`, a relation created through `POST /api/mentorship` has `pipelineStatus = 'SOURCED'`
- [ ] On an org with no custom stages, the created value is still `APPLICATION_100` — byte-identical behaviour
- [ ] A relation created by accepting a mentorship request and one created by CSV import both start on the same resolved stage
- [ ] The repair backfill is idempotent and leaves a `StatusChange` row per moved relation
- [ ] The new relation appears in a board column and in the funnel count immediately
- [ ] `npm run build` passes
- [ ] an e2e spec creates a custom stage set, then a relation, and asserts it renders in the first column — tagged `@smoke`
- [ ] a release fragment `releases/unreleased/start-stage-from-resolved-set.json` is added (`bump: "patch"`)

## 🇹🇷 Türkçe özet
Hattını özelleştirmiş bir kiracıda oluşturulan her ilişki görünmez oluyor: şemadaki varsayılan `APPLICATION_100` ve o anahtar kiracının kendi aşama kümesinde yoksa ilişki hiçbir sütunda görünmüyor, hiçbir huniye sayılmıyor. Düzeltme küçük — ilişki oluşturulurken kümenin ilk "yol üstü" aşaması hesaplanıp açıkça yazılıyor — ama beyaz etiket vaadini ilk günden kullanılamaz hâle getiren şey tam olarak bu. Mevcut yanlış konumdaki kayıtlar için de bir kez çalışan, denetlenebilir bir onarım betiği ekleniyor.

## 🤖 Agent prompt
```
Task (junior-friendly bug fix): new mentorship relations must start on the tenant's own
first on-path stage instead of the hardcoded APPLICATION_100 default.

Open first:
  prisma/schema.prisma:1696-1700     (pipelineStatus String @default("APPLICATION_100"))
  src/app/api/mentorship/route.ts    (createRelationSchema :17, POST :124)
  src/lib/pipeline.ts                (onPathKeys :236, ResolvedStage :215)
  src/lib/pipelineStages.ts          (resolvePipelineStages :18)
  src/lib/mentorshipDecision.ts
  src/app/api/admin/import/

Do:
1. Add a pure helper startStageKey(stages) to src/lib/pipeline.ts: first of onPathKeys(),
   falling back to 'APPLICATION_100'. Keep it Prisma-free — that file is client-safe.
2. POST /api/mentorship resolves the stages and passes pipelineStatus explicitly.
3. Same in mentorshipDecision (accepting a request) and in the CSV import path.
4. Leave the schema default alone — no schema change in this PR.
5. Idempotent repair script under prisma/backfills/ for relations already on an unresolved
   key; write a StatusChange row for each move so it is auditable, never silent.

Constraints:
- This must work on the FREE plan. Custom stages are premium to EDIT, but a relation landing
  on a nonexistent stage is a bug, not a tier.
- Do not use z.enum for stage keys anywhere (repo convention: z.string() + whitelist).
- No schema change means no prisma validate requirement, but run npm run build.
- Add releases/unreleased/start-stage-from-resolved-set.json with bump "patch".

Done when: a custom-stage org sees new relations in its first column, a default org is
unchanged, and the @smoke spec proves both.
```

**Zorluk** · S · **Tahmini** · 0.5-1 day · **Bağımlılık** · yok (better after #1624, but shippable today against the org-level resolver)

mersahin:
## Summary

A tenant that customised its pipeline got every new relation parked on the hardcoded canonical first key. `MentorshipRelation.pipelineStatus` defaults to `APPLICATION_100` in the schema and **every** application create path simply let that default apply — so in an org whose `PipelineStage` rows are `SOURCED / SCREENING / PLACED`, a freshly assigned mentee landed on a key that exists in none of them: no board column to render in, no funnel row to count in, and no stage select to move out of. The mentee was invisible from the moment they were assigned, which makes the flagship white-label feature unusable on day one.

The fix resolves the tenant's own stages at create time and writes the **first on-path** one explicitly. On-path is read from the schema (`PipelineStage.isOffPath`), not guessed from a name, so an org that ordered its "Withdrew" bucket at `order: 0` still starts on its first real stage. When the org has no `PipelineStage` rows at all, `resolvePipelineStages` returns the canonical set and the start stage is `APPLICATION_100` — byte-identical behaviour for anyone who never touched the stage editor.

Closes #1634

Preview: https://pr2236.interncrm.com

## Changes

**New helpers**

- `src/lib/pipeline.ts` — `startStageKey(stages)`: the first entry of `onPathKeys(stages)`. A set with no on-path stage at all falls back to **its own first stage by `order`**; only a genuinely *empty* set falls back to the exported `DEFAULT_START_STAGE` (`'APPLICATION_100'`, matching the schema default), which is exactly the org that resolves to the built-in catalogue. Pure and Prisma-free, so the file stays client-safe.
- `src/lib/pipelineStages.ts` — `resolveStartStage(orgId, locale?)`: the server-side wrapper, `startStageKey(await resolvePipelineStages(orgId))`, plus `createStartStageResolver()`, the same lookup memoized per call for a caller that walks a whole roster. One shared helper so the create paths — and the readers of "the first stage" — cannot drift apart.

**Creation sites changed** (all four application paths that create a `MentorshipRelation`)

| Site | Org resolved from |
|---|---|
| `src/app/api/mentorship/route.ts` — admin assignment (`POST /api/mentorship`) | `mentee.orgId` |
| `src/lib/mentorshipDecision.ts` — approving a mentorship request (admin queue *and* the mentor's own accept) | `gate.orgId` |
| `src/app/api/mentor/mentees/route.ts` — a mentor adding a mentee directly | `orgId` (the mentor's) |
| `src/app/api/register/route.ts` — the invitation auto-link ("click the link and you are connected", #51) | `user.orgId` |

In `mentorshipDecision.ts` the resolve happens *before* the `$transaction([...])` array, since an interactive read cannot live inside one.

**Readers of "the first stage" unified**

- `src/lib/menteeOnboarding.ts` resolves the start stage per org instead of comparing against a hardcoded `APPLICATION_100`. Its "move them off the first pipeline stage" step is *derived*, so with the constant it would have gone green the second a custom-stage relation was created — a tick no mentor could undo, one fewer `remaining`, and a mentee filtered out of `pendingOnboardings()` altogether.
- `src/lib/dormantFirstContact.ts` drops its own three-line copy of the resolver in favour of the shared one, so "parked at first contact" and "created at first contact" can never mean two different stages.

**Creation sites deliberately left alone** — each already carries a real, intended stage per row, and overriding it would be a regression:

- `prisma/seed-demo.mjs` and `scripts/seed-dummy.mjs` — write a chosen `pipelineStatus` so the demo data has a spread of journeys.
- `scripts/import-csv.mjs` — maps the legacy spreadsheet's numeric status column (`100 → APPLICATION_100`, …) onto the stage; that value *is* the import.
- `e2e/**` fixtures (incl. `e2e/helpers/db.ts`) — specs pin the stage they are testing.
- `src/app/api/admin/import/` — the issue lists this as a create path, but it only creates `User` rows; no relation is created there, so there was nothing to change.

**Repair backfill**

`prisma/backfill-relation-start-stage.mjs`, wired into `infra/deploy-prod.sh` next to the other idempotent backfills. It walks only the orgs that *have* custom stages and moves the stranded relations onto the set's first on-path stage (the same destination `startStageKey` picks, including its all-off-path fallback, so the two halves cannot disagree), writing a `StatusChange` row per move so the correction shows up in the relation's own history instead of a value silently changing under an admin. Attribution is the org's oldest `ADMIN`, falling back to the relation's mentor so the audit row is never dropped.

Its filter is **deliberately narrow**: still on the schema default `APPLICATION_100`, that key absent from the org's set, still `ACTIVE`, and never moved (zero `StatusChange` rows). "Any key outside the org's *current* set" would have been catastrophic — the stage editor replaces a set by deleting and recreating every row and remaps no relation, so a tenant that renames one stage briefly has its whole in-flight pipeline in that state (see the review follow-up below). The script is **dry-run by default** and prints a per-org count of what it would move; `deploy-prod.sh` passes `--apply`. Idempotent twice over: after a move the key is inside the set *and* the relation now has a status-change row.

**Not changed, on purpose**

- The Prisma `@default("APPLICATION_100")` stays. It is the last-resort floor for a row created outside the application; changing a column default would not fix this and would hide the bug. No schema change in this PR.
- The `PipelineStatus` enum and the canonical stage order are untouched.

## Review follow-up (second commit)

Three defects found in self-review, all real, all fixed:

1. **Blocker — the backfill was far broader than its stated scope.** It matched every relation whose key was outside its org's *current* stage set. `PUT /api/admin/organizations/[id]/pipeline-stages` `deleteMany`s the stage rows and `createMany`s the new ones and remaps nothing, so a tenant that renames one stage — or customises its pipeline after months on the built-in one — has its entire in-flight pipeline outside the current set for a moment. The next prod deploy would have swept all of it to stage 1 unattended: someone at `EMPLOYED_700` back to `SOURCED`, a `COMPLETED` relation reading as a fresh lead, and the daily dormancy sweep then mailing an already-placed mentee "hâlâ ilgileniyor musun?". Narrowed to the rows that can only be this bug, made dry-run by default, and `deploy-prod.sh`'s comment now describes what the code does rather than what it was meant to do.
2. **Major — `menteeOnboarding.ts` kept a hardcoded `FIRST_STAGE`**, so this PR would have regressed the onboarding checklist for exactly the custom-stage tenants it targets. Now resolved per org through the shared resolver.
3. **Minor — `startStageKey` fell back to `APPLICATION_100` for a non-empty, all-off-path set** (the editor accepts `isOffPath` on every row), writing a key in none of the tenant's rows and re-creating the very bug being fixed — while the backfill refused to guess in the same situation. A non-empty set now falls back to its own first stage by `order`.

## Deviations from the issue

- The issue's step 2 says to resolve "three-level after #1624" (programme → org → default). There is **no programme/`Program` model in the repo yet** — #1624 is not merged — so `resolveStartStage` resolves at the org level, which is exactly what `resolvePipelineStages` supports today. When the programme level lands it goes inside `resolvePipelineStages`, and this helper picks it up with no further change.
- The issue names the backfill's home as `prisma/backfills/` (registry from #1540). **That directory does not exist**; every backfill in this repo lives as `prisma/backfill-*.mjs` and is invoked from `infra/deploy-prod.sh`. This one follows the existing convention rather than inventing half a registry.
- Release fragment is named `relations-start-on-first-programme-stage.json` (per the task brief) rather than the issue's `start-stage-from-resolved-set.json`. Same content, `bump: "patch"`, with EN/TR/DE notes since the change is user-visible.

## Verification

- [x] `npm run lint` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings on `admin/companies` and `admin/mentorship`)
- [x] `npm run check:i18n` — 4231 keys × 3 locales, OK (no new user-facing strings; the release notes carry the EN/TR/DE copy)
- [x] `npm run check:release-fragments` — fragment validates, ships as `0.159.1-beta` (the "compaction looks stuck: 18 fragments pending" warning is pre-existing and unrelated)
- [ ] `npm run test:e2e` — **not run**: this container has no database, no Docker daemon and cannot boot the app. Verified statically only; the new specs are written for CI.
- [x] Added `e2e/relation-start-stage.spec.ts`, both tests tagged `@smoke`:
  - an org with `WITHDRAWN` (off-path, `order: 0`) / `SOURCED` / `SCREENING` / `PLACED` — a relation created through `POST /api/mentorship` (which takes no stage at all) lands on `SOURCED`, the card is asserted visible inside `board-column-SOURCED` on `/admin/board`, and the mentor's onboarding checklist still reports the "move them off the first stage" step as **pending**;
  - an org on the built-in stages still lands on `APPLICATION_100`, so the first assertion is not vacuous.
- [x] `e2e/pipeline-stages.spec.ts` gains a pure `startStageKey` case: built-in catalogue, off-path-first, all-off-path, and empty.

No schema change, so `npx prisma validate` was not required.

## Contributor terms


mersahin:
Review findings applied in 1b38713. All three reproduced against the code — **none rejected as a false positive**.

**1. Blocker — backfill scope (`prisma/backfill-relation-start-stage.mjs`).** Confirmed: `PUT /api/admin/organizations/[id]/pipeline-stages` replaces the stage set with `deleteMany` + `createMany` and remaps no relation, so `pipelineStatus: { notIn: [...known] }` describes a renaming tenant's *entire* in-flight pipeline, not just the rows this bug produced — and `deploy-prod.sh` ran it unattended behind `|| true`. Narrowed to rows that can only be the bug: still on the schema default `APPLICATION_100`, that key absent from the org's set, `status: 'ACTIVE'`, and `statusChanges: { none: {} }` (a relation that was ever moved was placed deliberately). It prints a per-org count and is **dry-run unless `--apply`** is passed; `deploy-prod.sh` passes it as a deliberate, commented choice, and that comment now describes the real scope. Idempotent twice over now — after a move the key is in the set *and* the relation has a status-change row.

**2. Major — `src/lib/menteeOnboarding.ts`.** Confirmed: `pipeline: relation.pipelineStatus !== FIRST_STAGE` with `FIRST_STAGE = 'APPLICATION_100'` would have gone true the moment a custom-stage relation was created, i.e. this PR itself regressing the checklist for the exact tenants it fixes. Now resolved per org. Rather than inline the lookup, the memoized resolver in `dormantFirstContact.ts` moved to `pipelineStages.ts` as `createStartStageResolver()` and both files use it, so `startStageKey` is the single definition of "the first stage" for writers and readers alike. `pendingOnboardings()` shares one resolver across the roster, so it stays a single stage query.

**3. Minor — `startStageKey` all-off-path fallback.** Confirmed: the PUT schema accepts `isOffPath` on every row and only requires `>= 1` stage, so `onPathKeys(stages)` can legitimately be `[]` on a non-empty set, and returning the canonical key there wrote a key the tenant does not have — the bug being fixed. A non-empty set now falls back to its own first stage by `order`; `DEFAULT_START_STAGE` is reserved for the genuinely empty set, so an org on the built-in catalogue is unchanged. The backfill's destination picker mirrors this exactly, so the "no on-path stage → skip forever" disagreement between the two halves is gone.

Coverage added: a pure `startStageKey` case in `e2e/pipeline-stages.spec.ts` (built-in / off-path-first / all-off-path / empty), and the custom-org spec now also asserts the onboarding step is still pending. Release fragment prose updated to match — the user-facing note now says explicitly that anyone already moved through the pipeline stays where they were put.

Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` clean (two pre-existing `exhaustive-deps` warnings only), `npm run check:i18n` OK (4231 × 3), `npm run check:release-fragments` exit 0 at `0.159.1-beta`. No schema change, so no `prisma validate`. The e2e suite still cannot run here (no DB, no Docker) — CI is the first execution.


---

Issue #1697, opened 2026-08-30
🌱 🔁 [Infra] Task · Move the 10 dispatchWebhook call sites onto emit(), plus a check:events CI guard


mersahin:
## 🎯 Goal
Mechanical but load-bearing: every place in the product that currently shouts directly at the webhook dispatcher should raise a domain event instead, and CI should stop the next contributor from adding an eleventh direct call. Once this lands, the epic's other three stories (audit, timeline, automation) get their input for free — they subscribe instead of asking each route to remember them.

This is a good first task: the target list is fixed and enumerable, the replacement is one-for-one, and the CI guard makes a mistake impossible to miss.

## 🔍 Current state
`grep -rn "dispatchWebhook(" src/ --include=*.ts` — ten call sites, each passing its own ad-hoc object:
| File:line | Event | Payload today |
|---|---|---|
| `src/app/api/apply/route.ts:122` | `application.created` | `{ mentorId, menteeName, email }` |
| `src/lib/stageChangeEffects.ts:90` | `pipeline.stage_change` | `{ relationId, from, to }` |
| `src/app/api/mentorship/route.ts:224` | `mentorship.created` | `{ relationId, mentorId, menteeId, companyId }` |
| `src/app/api/interactions/route.ts:116` | `interaction.logged` | `{ relationId, type, date }` |
| `src/lib/meetingAutoLog.ts:106` | `interaction.logged` | (auto-logged variant) |
| `src/app/api/evaluations/route.ts:134` | `evaluation.added` | `{ relationId, type, authorId }` |
| `src/app/api/meetings/route.ts:229` | `meeting.scheduled` | `{ title, scheduledAt, count }` |
| `src/app/api/meetings/instant/route.ts:110` | `meeting.scheduled` | (instant variant) |
| `src/app/api/meeting-series/route.ts:150` | `meeting.scheduled` | (series variant) |
| `src/app/api/interview-panels/[id]/score/route.ts:90` | `interview_panel.completed` | `{ panelId, subjectId }` |

- The three `meeting.scheduled` sites pass **different shapes** for the same event name — the catalogue (#1691) is where that gets reconciled, and this task is where the call sites are made to match it.
[...]

Issue #1693, opened 2026-08-30
🔌 [Infra] Task · emit() and the subscriber registry, delivered through the durable job queue


mersahin:
## 🎯 Goal
One function to raise an event, one place to register something that reacts to it, and a delivery path that survives a process restart. This is what makes the difference between "we log events" and "an event reliably causes work to happen" — the difference a webhook customer, an automation rule and a scheduled report all depend on.

## 🔍 Current state
- The nearest existing pattern is `src/lib/stageChangeEffects.ts:20-91`: one exported function that fans out inline, awaiting each side effect in the request handler. It works for three effects; it does not survive a crash and it makes the request as slow as its slowest effect.
- `src/lib/webhooks.ts:36-59` runs deliveries under `Promise.all` with a 5s `AbortSignal.timeout` (`:53`) precisely because an unresponsive endpoint used to hold the handler open (#895). Fire-and-forget, no retry, no record.
- `src/lib/activity.ts:32-58` establishes the contract this must copy: **never throws** — `logActivity` swallows its own failure so it cannot break the request it describes.
- `src/lib/realtimeBus.ts` is already an in-process bus and one of the three SPOFs blocking a second container (the others are the in-memory rate limiter and node-cron). Do not add a fourth.
- The durable queue lands in **#1671** (`Job`/`JobRun`, `enqueue()`/`claim()` with `SKIP LOCKED` and an idempotency key) and **#1673** (worker loop, exponential backoff, DLQ). The transactional write path and the relay land in **#1680**.

## 🛠 What to build
1. `src/lib/events/emit.ts` — `emit(name, payload, ctx)` where `ctx` carries `orgId`, actor, an optional `correlationId` and an optional `tx`:
   - look the entry up in the catalogue (#1691); `parse` the payload with its zod schema
   - write the `DomainEvent` row **through #1680's `publishEvent(tx, …)`** — do not open your own `prisma.domainEvent.create`, and do not enqueue subscriber jobs from inside the caller's request when a transaction is in play; that is the relay's job
   - the relay (#1680) fans a committed row out to subscribers, one `enqueue()` per subscriber (idempotency key = `eventId:subscriberName`, so a retry never double-delivers)
   - **never throws**, same contract as `logActivity`; on a validation failure throw in dev/CI (`NODE_ENV !== 'production'`) and log-and-drop in production
[...]

Issue #1691, opened 2026-08-30
🗃 [Infra] Task · DomainEvent model + the versioned event catalogue in src/lib/events/catalog.ts


mersahin:
## 🎯 Goal
Define the vocabulary once. Today "an event" is a string in a 7-element array with no payload type and no version, and no record is kept that it ever happened. This task adds the append-only `DomainEvent` table and the versioned catalogue that describes every event the product can raise — the single artefact that webhooks, the audit trail, the timeline and the automation rules will all read from.

> 🔗 This issue owns the **one** event table and the **one** catalogue for the whole programme, including the transactional-outbox epic #1668. See "Scope boundary" below before writing schema.

## 🔍 Current state
- `src/lib/webhooks.ts:11-22` — `WEBHOOK_EVENTS`, seven strings (`application.created`, `pipeline.stage_change`, `mentorship.created`, `interaction.logged`, `evaluation.added`, `meeting.scheduled`, `interview_panel.completed`) exported as a `const` tuple. No payload type: `dispatchWebhook(event, data: Record)` at `:26`.
- These names are already public: they are published unauthenticated in the `x-webhooks` block of `src/app/api/v1/openapi.json/route.ts`.
- There is no event table. The closest shapes are `ActivityLog` (`prisma/schema.prisma:2412-2439`: level/action/actorId/actorEmail/targetType/targetId/detail/ip/userAgent) and `AuditLog` (`:2477-2490`: actorId/action/targetId/detail).
- `src/lib/orgContext.ts:43-55` — `TENANT_MODELS` currently lists 11 models. A tenant-data model missing from it escapes the isolation middleware entirely.
- 88 models in `prisma/schema.prisma` today.

## 🛠 What to build
1. `prisma/schema.prisma` — new model, **purely additive** (a brand-new table, so `db push` resolves it as a plain CREATE; no re-key, nothing dropped):
   - `id`, `orgId String?` + `org Organization?` relation, `name String @db.VarChar(96)`, `version Int @default(1)`, `payload Json`, `actorId String?`, `actorType String?` (`user` | `system` | `apikey`), `subjectType String?`, `subjectId String?`, `correlationId String?`, `createdAt DateTime @default(now())`
   - indexes: `@@index([orgId, createdAt])`, `@@index([name, createdAt])`, `@@index([subjectType, subjectId, createdAt])`
[...]

PR #2237, opened 2026-09-06
ci: guard the webhook dispatcher behind one door (#1697)

mersahin:
## Summary

`Refs #1697` — **deliberately not `Closes`.** The half of the issue that is unblocked ships here; the other half is blocked on dependencies that do not exist in the tree. Details below.

Ten call sites reach the webhook dispatcher directly from inside whichever request handler happened to notice the change. Each direct caller is a place where the next cross-cutting concern — a durable queue, a retry, tenant scoping, a `DomainEvent` row, an audit entry, a timeline card — has to be remembered and wired by hand, and silently does not apply where it was not. #1697's answer is that producers raise a domain event through `emit()` and subscribers react.

**That migration cannot happen in this PR.** `emit()` (#1693), the versioned catalogue (#1691) and the `DomainEvent` model are all still open, with no branch on the remote and nothing under `src/lib/events/` — `emit()` has no implementation to move ten call sites onto. Writing a local `emit()` and a local payload catalogue here would fork exactly the two artefacts those issues explicitly own ("Do not create a second event model, a second catalogue, or a second insert path" — #1693's scope boundary), and #1693 would then have to unpick it.

So this lands the part that stands on its own and that the migration will consume: **the ratchet**. The problem is capped at ten and given a burndown list that the follow-up PR deletes line by line.

## Changes

- **`scripts/check-events.mjs`** (new) — fails when any file under `src/` calls or imports the webhook dispatcher outside three places: `src/lib/webhooks.ts` (which defines it), `src/lib/events/subscribers/webhook.ts` (the single legitimate caller once #1693 lands; not required to exist yet) and the admin test-ping route (allowlisted — a `'ping'` is deliverable but deliberately not subscribable, so it has no domain event to emit).
  - **Both exits are scanned, not one.** `src/lib/webhooks.ts` exports `dispatchWebhook(event, data)` *and* `deliverToWebhook(hook, event, data)` — the latter posts the same signed body the former sends, one receiver at a time. Six lines copied out of the dispatcher reach every subscriber without ever naming `dispatchWebhook`, so the scan matches both identifiers.
  - **The cap is counted in call sites, not files.** Each `PENDING_MIGRATION` entry records how many direct calls its file is on record for, and the recorded numbers must add up to a `PENDING_CALL_SITES` constant that may only ever be lowered. A second `dispatchWebhook()` added to a route that already has one is red — which matters, because six of the ten listed files are `route.ts` modules with several HTTP handlers, so "add a webhook fire to the PUT handler" is the next plausible direct call.
  - The ten current producers each carry the event they raise and what is specific about them — the three *different* payload shapes hiding behind one `meeting.scheduled` name, the second `interaction.logged` shape in `meetingAutoLog.ts`, and the `from === to` no-op guard in `stageChangeEffects.ts` that the emit must stay behind or a bulk advance floods the stream.
  - **A stale entry fails too.** If a producer stops reaching the dispatcher, the guard tells you to delete its line and lower the constant. The list can only shrink, and the failure text does not invite the next offender to add themselves to it: an eleventh producer is not accepted, and growing the list means editing a second, obviously load-bearing number.
- **`npm run check:events`** in `package.json`, wired into `.github/workflows/ci.yml` next to the other `check:*` guards, with the same comment-the-reasoning style.
- **`docs/testing.md`** — an "Architecture guards" row in the test-type table, stating the rule that is enforced *today* rather than #1693's end state, and the prose above it stops miscounting the rows it summarises.
- **`releases/unreleased/webhook-call-sites-on-emit.json`** — `bump: patch`, no `notes` (nothing user-visible).

No production code is touched. Event payloads are byte-identical because no call site moved.

## Review round 2 (commit `aafcd51`)

Review of the first commit found three ways the guard fell short of what its own header, the `ci.yml` step comment and the release fragment claimed. All three are fixed; each was reproduced first.

| # | Gap | Before | After |
|---|-----|--------|-------|
| 1 | The ratchet counted **files**, not call sites — the scan `continue`d on the first hit in a listed file, so everything after it was invisible | a second `dispatchWebhook()` in `src/app/api/mentorship/route.ts` → **exit 0** | recorded per-file counts, compared → **exit 1** |
| 2 | `deliverToWebhook` was a second, unguarded door the scan never looked at | a new route fanning out via `deliverToWebhook` → **exit 0** | both identifiers scanned → **exit 1** |
| 3 | The failure text invited the offender to append themselves to the list the header calls "not an exemption list" | wording only | wording closed, plus a `PENDING_CALL_SITES` constant the counts must sum to |

Plus a docs accuracy fix: the table row described `emit()` as today's rule and sent a reader looking for a `src/lib/events/` that does not exist. The script header also gained a short **WHAT IT CANNOT SEE** note — a text scan counts a commented-out call and misses an aliased import; it is a ratchet on the honest path, not a sandbox.

## Left out, and why

Two pieces of #1697 are not here:

1. **The ten call-site migrations.** Blocked on #1691 + #1693, as above. This PR is what makes that follow-up mechanical: the list of what to move, and what to watch for at each site, is now in the guard.
2. **The `prisma.activityLog.create` / `prisma.auditLog.create` rule** that #1697 also asks the guard to enforce "for an action name the catalogue owns". There is no catalogue to read owned names from. The rule would either pass vacuously or hardcode a second, forkable copy of the vocabulary #1691 exists to own — so it is left out, with a `NOT CHECKED YET` block in the script header naming #1691.

Consequence for the acceptance criteria: `grep -rn "dispatchWebhook(" src/` still returns ten producers, and the three `meeting.scheduled` shapes are still three shapes. Those tick when the migration lands, which is why this is `Refs` and not `Closes`.

## Proof the guard works

Every case below was run in the worktree and reverted.

**A second call in a file already on the list** (`src/app/api/mentorship/route.ts` — green before this round):

```
$ node scripts/check-events.mjs; echo "exit=$?"
event emission FAILED — the webhook dispatcher has more than one door:

  • src/app/api/mentorship/route.ts:225  is on record for 1 direct call site(s) but now has 2.
    A new direct call to the webhook dispatcher is not accepted — raise the event through
    emit() (src/lib/events/emit.ts, #1693), or land the change after #1693. The recorded
    number is a burndown figure and may only go down.
exit=1
```

**A new route fanning out through `deliverToWebhook`** — the six lines copied out of `dispatchWebhook`'s own body (green before this round):

```
$ node scripts/check-events.mjs; echo "exit=$?"
event emission FAILED — the webhook dispatcher has more than one door:

  • src/app/api/fake-door/route.ts:5  reaches the webhook dispatcher directly. A domain event
    must leave through emit() … PENDING_MIGRATION lists the ten call sites that predate this
    guard — it is a burndown list, not somewhere to add yourself for a green build.
exit=1
```

**A brand-new eleventh caller**, **a second `deliverToWebhook` in the allowlisted test-ping route**, and **a line deleted from `PENDING_MIGRATION` without lowering `PENDING_CALL_SITES`** all exit 1 as well.

**A producer genuinely migrating off the dispatcher** (call + import deleted from `meetingAutoLog.ts`, list entry left behind):

```
  • src/lib/meetingAutoLog.ts is on PENDING_MIGRATION but no longer reaches the webhook
    dispatcher — it has been moved onto emit(). Delete its line from PENDING_MIGRATION and
    lower PENDING_CALL_SITES in scripts/check-events.mjs (event: interaction.logged).
exit=1
```

**On this branch as it stands:**

```
$ npm run check:events; echo "exit=$?"
events OK — the dispatcher is reached from 1 dispatcher module(s), 1 allowlisted direct
deliverer(s) and 10 call site(s) across 10 producer(s) still awaiting emit() (#1693);
no new direct call sites (796 source files scanned).
exit=0
```

## Verification

- [x] `npm run lint` clean (only the pre-existing `react-hooks/exhaustive-deps` warnings on `admin/companies` and `admin/mentorship`)
- [x] `npm run check:i18n` (no new strings in this PR)
- [x] `npm run check:release-fragments` — fragment valid
- [x] `npm run check:events` — green, and red in every failure direction above
- [x] `.github/workflows/ci.yml` still parses as YAML after the comment edit
- [ ] `npm run test:e2e` — not runnable in this container (no database, no Docker daemon). No production code changed, so `e2e/webhooks-openapi.spec.ts` and `e2e/integrations.spec.ts` assert an unchanged surface; CI will confirm.
- [x] `npx prisma validate` — schema untouched

Preview: https://pr2237.interncrm.com

## Note for the maintainer

`check:release-fragments` warns that compaction looks stuck — 18 fragments pending, over the threshold of 12. That is unrelated to this PR, but `release-compact.yml` has apparently not folded fragments recently, so `CHANGELOG.md` and `src/lib/releaseNotes.ts` are behind.

## Contributor terms


mersahin:
Review addressed in `aafcd51`. I reproduced all four findings against the branch before changing anything, and all four were real — **none rejected**.

1. **(major) The ratchet counted files, not call sites.** Confirmed: a second `dispatchWebhook()` in `src/app/api/mentorship/route.ts` exited 0. `PENDING_MIGRATION` entries now carry the number of direct calls each file is on record for, and the scan compares counts with `matchAll` instead of stopping at the first `search()` hit. More than recorded is a new call site (red); fewer is a partial migration (lower the number) or a completed one (delete the line). The "an eleventh direct call is a red build" wording in `ci.yml`, the script header and the release fragment is now true rather than trimmed.
2. **(minor) `deliverToWebhook` was a second door.** Confirmed: a new route fanning out with the six lines copied from `dispatchWebhook`'s own body exited 0. Both identifiers are scanned now. `src/app/api/admin/webhooks/test/route.ts` is allowlisted in its own `DIRECT_DELIVERY_ALLOWED` list rather than in `DISPATCHER_HOMES` — it is not a dispatcher home, it is the one legitimate direct deliverer, and it is count-guarded too (a second call there is red).
3. **(minor) The failure text invited the next offender onto the list.** Removed, and closed mechanically as well: the recorded counts must sum to `PENDING_CALL_SITES`, a constant whose comment says it may only ever be lowered, so growing the list requires editing a second obviously load-bearing number rather than appending a line.
4. **(minor) `docs/testing.md` stated the end state as today's rule.** The row now says what is enforced now — no file under `src/` may reach the dispatcher beyond the ten listed call sites, which move onto `emit()` when #1693 lands — and the sentence under the table covers the last two rows instead of classifying 12 of 14.

One thing the review did not raise that is worth stating plainly, so it is in the record: this is a text scan. A commented-out call still counts as a call, and an aliased import (`import { dispatchWebhook as fire }`) slips past. The script header now says so under **WHAT IT CANNOT SEE** — it is a ratchet on the honest path, not a sandbox against a determined bypass.

`tsc --noEmit`, `lint`, `check:i18n` and `check:release-fragments` all clean; the guard is green on the branch and red in six separate failure directions (listed in the PR body).


---

Issue #1673, opened 2026-08-30
🏃 [Infra] Task · Worker loop with a handler registry, exponential backoff, DLQ and graceful shutdown


mersahin:
## 🎯 Goal
The half that actually does the work: a loop that claims jobs, looks up a typed handler by name, runs it under a timeout, records the attempt, and on failure schedules a retry with exponential backoff — until the attempt budget runs out and the job lands in the dead-letter queue where an operator can see it. This is what turns "the reminder cron threw and printed to stdout" into "attempt 3 of 5, next at 09:14, error: ECONNREFUSED smtp".

## 🔍 Current state
- Today a failing scheduled job is caught by one `try/catch` per tick and `console.error`'d — `src/services/emailService.ts:3075-3081`, `:3106-3108`, `:3117-3119`. There is no retry, and the tick's remaining steps are skipped because the six calls at `:3063-3074` are sequential inside the same `try`.
- The process-boot path is deliberately `fetch`-only: `src/instrumentation.ts:10-16` documents that `middleware.ts` makes Next compile instrumentation for the **edge** runtime, where Prisma and nodemailer cannot resolve — so anything Node-only must be started from a route handler, not imported into `instrumentation.ts`.
- `src/app/api/cron/start/route.ts:22-35` is the existing "start background work in this process" endpoint (`CRON_SECRET`-gated, idempotent).
- `src/lib/logger.ts` is the structured logger; `src/lib/activity.ts:32-37` mirrors into `ActivityLog`.

## 🛠 What to build
1. **`src/lib/jobs/registry.ts`** — `registerJob(name, handler)` and `getHandler(name)`. A `Job` row whose `name` has no handler must be surfaced as a startup error, not discovered at 03:40; add a `assertAllHandlersRegistered()` called from the worker start path.
2. **`src/lib/jobs/worker.ts`**:
   - `startWorker({ concurrency, pollMs, leaseMs })` — poll → `claim()` → run each job under `Promise.race` with a per-job timeout → write the `JobRun` row → mark `SUCCEEDED`, or compute the next `runAt`.
   - Backoff: `runAt = now + min(baseMs * 2^attempt * jitter, capMs)`, defaults base 30s / cap 1h, overridable per job name.
   - `attempts >= maxAttempts` → `status = DEAD_LETTER`, `lastError` set, one `logger.error` and one `ActivityLog` row (`job.dead_letter`).
[...]

Issue #1671, opened 2026-08-30
🗃 [Infra] Task · Job + JobRun schema and enqueue()/claim() with SKIP LOCKED and an idempotency key


mersahin:
## 🎯 Goal
The storage layer of the queue. One table holds work that must happen, one table holds what happened on each attempt, and a claim query hands a job to exactly one worker even when two containers are racing. Nothing runs yet — this task is the durable substrate the worker loop plugs into.

> 🔗 Absorbs #1606 (closed as duplicate) — its run-ledger fields (`trigger` provenance, PII-free `summary` counters), its 90-day retention sweep and its rule that a ledger write must never kill the job it records are merged into the `JobRun` model below. #1606 wrapped the node-cron callbacks in `emailService.ts`, which #1677 deletes; there is exactly **one** `JobRun` model in this repo and it is this one.

## 🔍 Current state
- No queue storage exists. `prisma/schema.prisma` has 118 models; the closest analogue is `EmailLog` (`prisma/schema.prisma:2453`), which is a *log*, not a queue — it is written after the fact and has no claim, attempt or retry semantics.
- The house style for a log/ledger table is visible at `prisma/schema.prisma:2453-2476` (`EmailLog`: enum status, nullable `error String @db.Text`, indexes on the columns the admin view filters by) and `prisma/schema.prisma:2412-2432` (`ActivityLog`).
- Tenant anchoring convention: `src/lib/orgContext.ts:43` lists the 11 models carrying a nullable `orgId` that the isolation middleware scopes. #1559 and #1560 are extending that set and adding a `check:tenant-models` CI guard.
- `src/lib/prisma.ts` is the client singleton; raw SQL is available through `prisma.$queryRaw`.
- **Today there are 12 scheduled jobs** — 10 `cron.schedule` calls in `src/services/emailService.ts:3055-3195` and 2 in `src/lib/newsletterDispatch.ts:449-461`. (Grep returns 11 and 3 hits respectively; the extra hit in each file is the `Map<string, ReturnType<typeof cron.schedule>>` type annotation at `emailService.ts:3049` / `newsletterDispatch.ts:437`, not a schedule.) Every one of them becomes a `Job` row once #1677 lands, so this table is what makes their outcomes visible for the first time.

## 🛠 What to build
1. **Schema** (`prisma/schema.prisma`, additive only):
   - `enum JobStatus { PENDING RUNNING SUCCEEDED FAILED DEAD_LETTER CANCELLED }`
[...]

PR #2239, opened 2026-09-06
feat: queue depth and DLQ size on /api/health, daily dead-letter alert (#1674)

### LINKED ISSUE #1674: 🌱 [Admin] Task · Queue depth and DLQ size on /api/health, plus a daily dead-letter alert e-mail
## 🎯 Goal
A dead-letter queue nobody looks at is a folder of lost work. This adds the two cheapest possible eyes on it: queue counters in the existing health endpoint (so an uptime monitor can alert on a backlog) and one daily e-mail when the DLQ is not empty. Small, isolated, and it uses two patterns the repo already has.

## 🔍 Current state
- `/api/health` already reports DB and SMTP probes, version/sha, uptime, `responseMs` and e-mail delivery health, with detail gated behind `HEALTH_TOKEN` or an ADMIN session — `src/app/api/health/route.ts:20-108`.
- The e-mail-health derivation it uses is a good template for a small "compute a health object from a table" module — `src/lib/emailHealth.ts:1-52` (PII-scrubbed, returns `lastOk` / `lastError` / `failuresSinceOk` / `attempts24h`).
- There is already an hourly cron that alerts when sends keep failing — `src/services/emailService.ts:3139-3148` (`runEmailHealthCheck`) — and a reusable alert-mail script for workflows at `scripts/send-alert-email.mjs`.
- `ALERT_EMAIL_TO` is the established operator alert address (used by `.github/workflows/backup-verify.yml` and `scripts/e2e-report-email.mjs`).
- Health is on the k6 anonymous-GET mix, and `docs/testing.md:288-296` warns that `/api/health` already costs four `EmailLog` queries per anonymous request — do not make it more expensive for anonymous callers.

## 🛠 What to build
1. `src/lib/jobs/health.ts` — `jobQueueHealth()` returning `{ pending, running, deadLetter, oldestPendingAgeSec, failedLast24h }` from grouped counts over `Job`.
2. Wire it into `src/app/api/health/route.ts` **inside the token/ADMIN-gated detail branch only**, next to the existing e-mail health block. The anonymous response must not gain a query.
3. A `jobs.dlq-alert` job (registered on the queue, scheduled once a day by the scheduler story) that sends one e-mail to `ALERT_EMAIL_TO` listing dead-letter jobs by name with counts and the newest error — and sends **nothing** when the DLQ is empty. Follow the "green is silent" convention the k6 and e2e reports use.
4. `sendEmail({ category: 'ops-alert' })` so the send lands in `EmailLog` with a category. This is an operator alert to a configured address, not a user — pass no `userId` and leave the required `// no-user-row: operator alert address, not a User` comment that `e2e/email-groups-footer.unit.spec.ts` scans for.

## ✅ Acceptance criteria
- [ ] `GET /api/health?jobs=1` with a valid `HEALTH_TOKEN` (or an ADMIN session) returns the queue counters
- [ ] An anonymous `GET /api/health` issues **no additional** queries — verified by reading the route, and the k6 `ep:health` budget in `k6/nightly-load.js` still passes under `K6_SMOKE=1`
- [ ] With an empty DLQ the daily alert job sends nothing and writes no `EmailLog` row
- [ ] With a non-empty DLQ exactly one e-mail is sent, naming each job type and count
- [ ] `npm run build` passes
- [ ] an e2e spec asserts the gated counters and the silent-when-green behaviour
- [ ] a release fragment `releases/unreleased/job-queue-health.json` is added

## 🇹🇷 Türkçe özet
Kimsenin bakmadığı bir ölü mektup kuyruğu, kaybolmuş işlerin durduğu bir klasörden ibarettir. Bu görev en ucuz iki gözü ekliyor: mevcut `/api/health` uç noktasına kuyruk sayaçları (jeton/ADMIN ile korunan detay kısmına, anonim yanıt ağırlaşmasın) ve DLQ boş değilse günde bir kez operatöre giden tek bir uyarı e-postası. Kuyruk boşsa e-posta gitmiyor — depodaki "yeşilse sessiz" geleneğine uyuyor. Stajyer için uygun: iki hazır kalıbı (emailHealth + alert e-postası) taklit ediyor.

## 🤖 Agent prompt
```
Objective: expose job-queue counters on /api/health (gated detail only) and add a daily
dead-letter alert e-mail that is silent when the DLQ is empty.

Open first:
  src/app/api/health/route.ts:20-108  — the existing gated-detail structure. HEALTH_TOKEN or
                                        an ADMIN session unlocks detail; anonymous stays thin.
  src/lib/emailHealth.ts:1-52         — copy this module's shape for jobQueueHealth()
  src/services/emailService.ts:416    — sendEmail signature. Read the comment at :445-455
                                        about `userId` and the required
                                        `// no-user-row: <why>` marker; e2e/email-groups-
                                        footer.unit.spec.ts scans source for it.
  scripts/send-alert-email.mjs        — the "green is silent" alert convention
  docs/testing.md:288-296             — why /api/health must not get more expensive

Constraints:
  - Do NOT add a query to the anonymous /api/health path; that endpoint is in the nightly k6
    mix with a latency budget.
  - The alert address is ALERT_EMAIL_TO (an operator address, not a User row).
  - No new user-facing UI strings, so no i18n keys needed. If you add any, EN/TR/DE parity is
    enforced by `npm run check:i18n`.
  - Register the daily job on the queue; the scheduler task in the sibling story owns the
    cron expression.
Done when: gated counters work, empty DLQ sends nothing, non-empty sends one mail, e2e green,
release fragment added, PR open with preview URL.
```

**Zorluk** · S · **Tahmini** · 0.5-1 day · **Bağımlılık** · the Job/JobRun schema task and the worker task in this story

mersahin:
## Summary

A job queue with a growing backlog, or a dead-letter table that is quietly filling up, fails **silently** — nothing reports either number, so the first symptom is a user noticing that an e-mail never arrived. This adds the two cheapest possible eyes on both: queue counters on the existing health endpoint, and one daily e-mail when the dead-letter queue is not empty.

`/api/health` gains a `jobs` block — `pending`, `running`, `deadLetter`, `oldestPendingAgeSec`, `failedLast24h` — read **only** for a caller who opts in with `?jobs=1` **and** has proved who they are: a matching `HEALTH_TOKEN` or an ADMIN session. That is deliberately a *stricter* gate than the detail block the counters sit in. The detail block is fail-open when `HEALTH_TOKEN` is unset (the deploy drift gate parses `sha` out of it, and closing it by default would blind every gate), but nothing automated parses `jobs` — so on the un-tokened servers we actually run today, an anonymous `?jobs=1` gets no counters and pays no query. That keeps the endpoint inside its nightly k6 `ep:health` budget (`docs/testing.md` warns it already pays four `EmailLog` queries) and makes the issue's "token or ADMIN" acceptance criterion true in the deployed configuration, not only in a configured one. No existing field moved or changed, so the deploy gate's `sha` parse is untouched.

The daily alert runs at **06:45 UTC** and mails `ALERT_EMAIL_TO` in Turkish when the dead-letter queue is not empty: how many, since when, the job types with counts, and the **distinct failure reasons with counts** — each reason sanitised through the shared `sanitizeError()` scrubber, first line only, truncated, lists capped at 10 with a "+N more" tail. Never a payload, never a row-by-row dump: the payload is the one part of a job that can carry a user's data and this mail goes to a configured address, not to a session. **An empty queue sends nothing at all** — no message, no `EmailLog` row, no activity row — the same "green is silent" discipline `scripts/e2e-report-email.mjs` and `scripts/k6-report-email.mjs` follow, because a daily "all clear" trains the reader to archive the alert unread.

### The alert reports what actually happened to the mail

The alert travels by the very channel it may have to report on, so it borrows both halves of `alertEmailHealth`'s (#1190) discipline:

- a **durable record first** — one `jobs.dlq_alert` `ActivityLog` row (total, oldest, worst job type; capped for the VARCHAR(191) column, the #1268 lesson) written *before* the send, so a dead-letter queue filling up while SMTP is broken is visible somewhere that does not depend on SMTP;
- and **`sent` means SENT**. `sendEmail()` answers `'SKIPPED'` without throwing on three paths (unsubscribed group, demo mode, `SMTP_USER` unset), and #1431 exists because four routes read that silence as success. `runDeadLetterAlert()` returns `{ sent, total, delivery, reason }`: `delivery` is the `EmailDeliveryResult` verbatim, `reason` is `no_alert_address` / `not_delivered` / `send_failed`. The e2e asserts the SKIPPED shape, because that is what the blanked-SMTP test config really produces.

### Which cron convention, and why

Registered from **`/api/cron/start`**, not from `initCronJobs()` — the same call the newsletter cron makes, for the same reason (#1469): `src/lib/jobs/dlqAlert.ts` imports `emailService` to send its mail, so registering it inside `emailService` would close a one-way import into a cycle. It is also exposed as `GET /api/cron?job=dlq-alert` for an ADMIN, deliberately *outside* the "run everything now" batch: the batch gets clicked repeatedly while somebody is fixing something, and this job's contract is one mail a day.

node-cron is only the carrier. When #1673's worker and the scheduler story land, `runDeadLetterAlert()` becomes the `jobs.dlq-alert` handler registered on the queue and this timer goes away with the other twelve — the function is already the handler either way.

### What this PR carries from a sibling task

#1671 (the `Job`/`JobRun` schema and `enqueue()`/`claim()`) **has not landed and has no PR in flight** — there is no queue table in `main` to count. Rather than ship nothing, this PR carries the `JobStatus` enum and the `Job` model, with field names and types **copied verbatim from #1671's specification**, so the two land as one table rather than as two competing ones. **Whoever merges second deletes their copy of that schema block rather than resolving a conflict by hand** — cross-linked on #1671 so it is not a surprise. `Job` is registered in `TENANT_MODELS` (`src/lib/orgContext.ts`) exactly as #1671 asks; `JobRun` is **not** here, and neither are `enqueue()`, the `SKIP LOCKED` `claim()`, `reapExpiredLeases()`, `recordRun()`, `pruneJobRuns()` or the worker loop.

One deliberate departure from #1671's index list: `@@index([orgId, status, runAt])` is added. `Job` would otherwise be the only `TENANT_MODELS` entry with no `orgId`-leading index, and under `MT_ENFORCE_ISOLATION` every auto-scoped query — including the future hot claim query — would filter `orgId` row-by-row on top of the status index. `idempotencyKey` stays a *global* `@unique` rather than `@@unique([orgId, idempotencyKey])`, and the schema now says why: `orgId` is nullable, MySQL treats NULLs in a unique index as distinct, so a composite key would silently stop de-duplicating anything while every row still has a null org (i.e. today).

## Changes

- `prisma/schema.prisma` — `enum JobStatus` + `model Job` (additive; `db push`, no migration authored).
- `src/lib/jobs/health.ts` — `jobQueueHealth()`: three queries, one grouped count plus an oldest-due lookup and a 24h failure count. Shaped after `src/lib/emailHealth.ts` (derived from the rows, never a separate counter that can drift).
- `src/lib/jobs/dlqAlertMessage.ts` — the pure message builder and reason normaliser: no Prisma, no nodemailer, no node-cron, so the "silent when green" property is unit-testable. Split out for the same reason `lib/newsletter.ts` is kept client-safe.
- `src/lib/jobs/dlqAlert.ts` — `getDeadLetterSummary()`, `runDeadLetterAlert()` (the `jobs.dlq_alert` activity row, then `sendEmail({ category: 'ops-alert' })` with the required `// no-user-row:` marker — `ALERT_EMAIL_TO` is an operator address, not a `User`), and `initDeadLetterAlertCron()`.
- `src/app/api/health/route.ts` — `?jobs=1`; the gate now answers with both `detail` (fail-open, what the deploy gate needs) and `verified` (fail-closed, what the counters need). The session is only read when it can change the answer, so an anonymous liveness probe on an un-tokened server behaves exactly as before.
- `src/app/api/cron/start/route.ts`, `src/app/api/cron/route.ts` — registration and the named manual run.
- `src/lib/orgContext.ts` — `Job` added to `TENANT_MODELS`. The operational readers run outside any request scope, where the isolation middleware does not engage, so they still see the whole queue — which is what an operator alert has to do.
- `e2e/job-queue-health.spec.ts` — seeds `Job` rows and asserts: anonymous `?jobs=1` gets nothing, the gated response without `?jobs=1` still gets nothing, both proof paths (token header, ADMIN session) get the counters (and no name, payload or tenant), an empty DLQ produces `{ sent: false, total: 0 }` and **no** `EmailLog` row, and three dead-lettered jobs produce exactly one `ops-alert` row plus the activity row — reported as `sent: false, delivery: 'SKIPPED'`, since e2e blanks SMTP. Cleanup removes only the ledger row this run created, not every dead-letter alert a long-lived dev database ever recorded.
- `e2e/job-queue-dlq-alert.unit.spec.ts` — pure: silence on empty, the message contents, HTML escaping, the truncation tail, reason sanitising, plus a source-shape assertion that `jobQueueHealth()` is called exactly once, past the gate, and behind `access.verified` rather than the fail-open branch.
- `playwright.config.ts` — pins `ALERT_EMAIL_TO` to a synthetic `ops-alert@e2e.local` so the alert's contract is testable and a real address can never be picked up by a test run (`SMTP_USER` is already blanked, so the send stops at a `SKIPPED` `EmailLog` row).
- `docs/testing.md`, `.env.example` — the `ep:health` cost note now records why the counters are opt-in *and* fail-closed; the env docs stop claiming `ALERT_EMAIL_TO` is read only by CI workflows (the app's e-mail-health alert already read it).
- `releases/unreleased/queue-depth-and-dlq-alert.json` — `minor` (a new API capability, a new scheduled mail, a new table), no `notes`: nothing here is user-visible — it is an operator endpoint and an operator mailbox.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings)
- [x] `npm run build` passes; `npx prisma validate` / `format` / `generate` clean
- [x] `npm run check:i18n`, `check:release-fragments`, `check:openapi`, `check:query-scalars`, `check:auth-reads`, `check:locale-dates`, `check:stage-keys` all green
- [ ] `npm run test:e2e` — **not run here**: this environment has no database and no browser, so the two new specs were written but not executed. CI is the first run.
- [x] No new user-facing strings, so no i18n keys (the operator mail is Turkish by convention, like the other alert mails)

Preview: https://pr2239.interncrm.com

Closes #1674

## Contributor terms

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,12 +60,16 @@ MAIL_FROM_NAME="Internship CRM"
 # See docs/EMAIL_DELIVERABILITY.md §2d.
 #UNSUBSCRIBE_MAILTO="unsubscribe@crm.ersah.in"
 
-# Automated-test alerts — these are consumed ONLY by the nightly stress GitHub
-# Actions workflow (.github/workflows/stress.yml), NOT by the running app. Set
-# them as GitHub repo secrets, not in a runtime .env. Listed here for discovery.
-#   ALERT_EMAIL_TO     comma-separated recipient(s) emailed when a test fails
+# Operator alerts.
+#   ALERT_EMAIL_TO     comma-separated recipient(s) for automated alerts. Read
+#                      by the CI workflows (stress, k6, e2e-full, backup-verify)
+#                      as a GitHub repo secret AND by the running app: the
+#                      hourly e-mail-delivery alert (#1190) and the daily
+#                      dead-letter alert (#1674) mail it. Both are silent while
+#                      everything is healthy; unset, the app only logs.
 #   STRESS_TARGET_URL  env the nightly stress test hits (defaults to production)
 # For a local stress run use BASE_URL instead: BASE_URL=... npm run test:stress
+#ALERT_EMAIL_TO=""
 
 # Inbound email (mentee/mentor replies routed back into a message thread)
 # INBOUND_EMAIL_DOMAIN: domain used in the generated reply-to address
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -278,6 +278,13 @@ As the table grows, that query — not the app — is the likeliest cause of an 
 breach. Setting `HEALTH_TOKEN` on the server removes the whole detail path from the anonymous
 probe and is the cheaper fix.
 
+The job-queue counters added in #1674 deliberately did **not** join that bill: they are read
+only for a caller who is past the detail gate *and* asks for them with `?jobs=1`, so the
+anonymous mix (`ep:health`, `ep:health_db`) issues exactly the queries it issued before.
+Anything added to `/api/health` from here on should follow the same shape — opt-in, inside
+the gated branch — and `e2e/job-queue-dlq-alert.unit.spec.ts` asserts it for the counters by
+reading the route's source.
+
 ## OS accessibility media preferences
 
 Three user preferences the browser exposes as media features are **supported and
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -279,10 +279,15 @@ breach. Setting `HEALTH_TOKEN` on the server removes the whole detail path from 
 probe and is the cheaper fix.
 
 The job-queue counters added in #1674 deliberately did **not** join that bill: they are read
-only for a caller who is past the detail gate *and* asks for them with `?jobs=1`, so the
-anonymous mix (`ep:health`, `ep:health_db`) issues exactly the queries it issued before.
-Anything added to `/api/health` from here on should follow the same shape — opt-in, inside
-the gated branch — and `e2e/job-queue-dlq-alert.unit.spec.ts` asserts it for the counters by
+only for a caller who asks for them with `?jobs=1` **and** has proved who they are — a
+matching `HEALTH_TOKEN` or an ADMIN session. Note that this is a *stricter* gate than the
+detail block above it: the detail block is fail-open when `HEALTH_TOKEN` is unset (the deploy
+drift gate parses `sha` out of it), while the counters fail closed, because nothing automated
+parses them and an un-tokened server would otherwise answer any anonymous `?jobs=1` with the
+queue's depth and three extra queries. So the anonymous mix (`ep:health`, `ep:health_db`)
+issues exactly the queries it issued before. Anything added to `/api/health` from here on
+should follow the same shape — opt-in, and gated on `access.verified` unless a deploy gate
+genuinely needs it — and `e2e/job-queue-dlq-alert.unit.spec.ts` asserts it for the counters by
 reading the route's source.
 
 ## OS accessibility media preferences
--- a/e2e/job-queue-dlq-alert.unit.spec.ts
+++ b/e2e/job-queue-dlq-alert.unit.spec.ts
@@ -19,7 +19,8 @@ import {
 //      row-by-row dump.
 //
 // The route-shape test at the bottom guards the *other* half of the issue: the
-// counters must cost an anonymous caller nothing.
+// counters must cost an anonymous caller nothing, and must not fall out of the
+// endpoint's fail-open detail branch on a server with no HEALTH_TOKEN.
 
 const empty: DeadLetterSummary = {
   total: 0,
@@ -118,18 +119,24 @@ test('a failure reason is sanitised, single-line and bounded', () => {
 // is read out of the source — the same trick e2e/email-groups-footer.unit.spec.ts
 // uses. /api/health is in the nightly k6 anonymous-GET mix with a latency
 // budget; a stray call up here would be paid by every uptime probe.
-test('the queue counters are read only behind the detail gate and ?jobs=1', () => {
+test('the queue counters are read only for a verified caller who asks for them', () => {
   const src = fs.readFileSync(path.join(process.cwd(), 'src/app/api/health/route.ts'), 'utf8');
 
-  // Exactly one call site, and it is opt-in.
+  // Exactly one call site, and it is opt-in AND behind proof of identity.
   const calls = src.match(/jobQueueHealth\(\)/g) ?? [];
   expect(calls).toHaveLength(1);
   expect(src).toContain("params.get('jobs') === '1'");
-  expect(src).toContain('...(wantsJobs ? { jobs: await jobQueueHealth() } : {})');
+  expect(src).toContain('...(wantsJobs && access.verified ? { jobs: await jobQueueHealth() } : {})');
 
-  // …and it sits after the anonymous early return, so an unauthorised caller
-  // never reaches it.
-  const anonymousReturn = src.indexOf('if (!(await maySeeDetail(request)))');
+  // `detail` is fail-open when HEALTH_TOKEN is unset — that is the bargain the
+  // deploy drift gate needs. `verified` is not, and the counters ride on
+  // `verified`, so an un-tokened server (every environment today) still refuses
+  // them to an anonymous ?jobs=1.
+  expect(src).toContain('return { detail: !expected, verified: false };');
+
+  // …and the read sits after the anonymous early return, so a caller the gate
+  // turned away never reaches it.
+  const anonymousReturn = src.indexOf('if (!access.detail)');
   expect(anonymousReturn).toBeGreaterThan(0);
   expect(src.indexOf('await jobQueueHealth()')).toBeGreaterThan(anonymousReturn);
 
--- a/e2e/job-queue-dlq-alert.unit.spec.ts
+++ b/e2e/job-queue-dlq-alert.unit.spec.ts
@@ -1,0 +1,140 @@
+import { test, expect } from '@playwright/test';
+import fs from 'fs';
+import path from 'path';
+import {
+  buildDeadLetterAlert,
+  normalizeReason,
+  MAX_ROWS_SHOWN,
+  type DeadLetterSummary,
+} from '@/lib/jobs/dlqAlertMessage';
+
+// The dead-letter alert's two promises (#1674), tested without a database, a
+// mail transport or a browser:
+//
+//   1. It is SILENT when the queue is clean. That is the property the whole
+//      convention rests on, and it is one `null` return — easy to break by
+//      accident, invisible until somebody starts ignoring the alert.
+//   2. What it says is bounded and safe: counts, an age, job type names and
+//      grouped failure reasons — never a payload, never an address, never a
+//      row-by-row dump.
+//
+// The route-shape test at the bottom guards the *other* half of the issue: the
+// counters must cost an anonymous caller nothing.
+
+const empty: DeadLetterSummary = {
+  total: 0,
+  oldestAt: null,
+  byName: [],
+  reasons: [],
+  reasonsSampledFrom: 0,
+};
+
+function summary(over: Partial<DeadLetterSummary> = {}): DeadLetterSummary {
+  return {
+    total: 3,
+    oldestAt: '2026-09-01T06:15:00.000Z',
+    byName: [
+      { name: 'email.send', count: 2, lastFailedAt: '2026-09-05T21:00:00.000Z' },
+      { name: 'report.weekly', count: 1, lastFailedAt: '2026-09-04T09:30:00.000Z' },
+    ],
+    reasons: [
+      { reason: 'Connection refused', count: 2 },
+      { reason: 'ETIMEDOUT', count: 1 },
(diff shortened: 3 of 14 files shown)


mersahin:
## Review follow-ups (692a87d)

Verified all six findings against the code. Four fixed as written, two fixed differently from the suggestion — reasoning below so the departures are on the record.

**1. `runDeadLetterAlert()` reported `sent: true` for a mail that was never delivered — fixed (this was the real one).** `sendEmail()` returns `EmailDeliveryResult` and answers `'SKIPPED'` without throwing on three paths (group/preference gate, demo mode, `SMTP_USER` unset); the function `await`ed it and threw the answer away. On a box with SMTP unconfigured, a filling dead-letter queue produced `{"dlqAlert":{"sent":true,"total":12}}` and nothing left the machine — exactly what #1431 fixed elsewhere. Now: the outcome is captured and returned as `{ sent, total, delivery, reason }` (`sent` means SENT; `reason` ∈ `no_alert_address` | `not_delivered` | `send_failed`), a warning is logged when the mail did not leave, and — following `alertEmailHealth`, which writes its durable record *first* precisely because the alert rides the channel it reports on — a `jobs.dlq_alert` `ActivityLog` row is written before the send. The e2e now asserts the truth under the blanked-SMTP config (`sent: false`, `delivery: 'SKIPPED'`) plus the `EmailLog` row and the new activity row.

**2. The counters inherited the fail-open detail gate — fixed by gating them, not by documenting the gap.** The finding is right that `HEALTH_TOKEN` is unset in every deployed environment (the k6 script, `docs/testing.md` and `e2e/health.spec.ts` all say so), so `?jobs=1` really did answer anyone. Documenting it would have left AC #1 false. Nothing automated parses `jobs` — only `sha` has a deploy gate depending on it — so the counters owe the fail-open bargain nothing. `maySeeDetail()` is now `resolveAccess()` returning `{ detail, verified }`: `detail` stays fail-open (the drift gate needs that), `verified` requires a real token match or an ADMIN session, and the counters ride `verified`. The session read only happens when it can change the answer (token gate closed, or `?jobs=1` asked), so an anonymous liveness probe on an un-tokened server issues exactly what it did before. The e2e now also exercises the ADMIN-session path with no token header, so both proof paths are covered rather than only the pinned-token one.

**3. `patch` bump on a feature — fixed.** Now `minor`; the fragment prose is updated to match the new behaviour (`0.160.0-beta`).

**4. `Job` was the only `TENANT_MODELS` entry with no `orgId`-leading index — fixed, in part.** Added `@@index([orgId, status, runAt])`, and said in the schema comment that this is a deliberate departure from #1671's index list. I did **not** add `org Organization? @relation(...)`: two other `TENANT_MODELS` entries carry a bare `orgId` with no relation (`AccountLockout`, `InterviewPanel`), so a relation is not the convention the finding claims it is, and adding one means a back-relation on `Organization` — a second place for #1671's copy of this model to conflict. The index is the half that has a measurable cost; the FK is latent (there is no org-delete path today) and belongs with whatever makes `orgId` non-null.

**5. Shipping #1671's table early — accepted as a hazard, rejected as a change.** Removing the model removes the feature: there is nothing to count without it, and #1671 has no PR in flight to wait for. Mitigated the way the finding suggests instead: cross-linked on #1671 (see the comment there), and the schema comment plus the PR body now say plainly that whoever merges second deletes their copy of the block rather than hand-resolving a conflict. The field list is byte-identical to #1671's spec, so a reconciliation cannot produce a column drop under `db push --accept-data-loss`.

  I did **not** take `@@unique([orgId, idempotencyKey])`. `orgId` is nullable and MySQL treats NULLs in a unique index as distinct, so a composite key would silently stop de-duplicating *anything* while every row still has a null org — which is today, and for the whole single-tenant present. That is a worse failure than the cross-tenant key collision it would prevent, so the global `@unique` stays with a doc comment explaining when to revisit it.

**6. e2e cleanup deleted ledger rows it did not create — fixed.** The `finally` block deleted every `EmailLog` row matching the dead-letter subject prefix, including the pre-existing rows the test had just counted and tolerated. It now captures the id of the one row this run produced and deletes only that.

Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings), `check:i18n` OK (4231 keys × 3 locales), `check:release-fragments` OK (`0.160.0-beta`), `check:openapi` OK (358 operations, health still classified `public` as before), `prisma format` / `validate` clean. The e2e specs still cannot run here (no database, no browser) — CI is their first run.


---

PR #2242, opened 2026-09-07
test: two-tenant fixture and an `isolation` Playwright project with the flag on (#1566)

### LINKED ISSUE #1566: 🌱 [Infra] Task · Two-tenant seed fixture and a Playwright project that boots with the flag on
## 🎯 Goal
Every isolation test today either flips the flag inside its own process or scopes queries by hand — neither proves anything about the *running application*. Give the suite two real tenants and a Playwright project whose web server actually has `MT_ENFORCE_ISOLATION=true`, so the next task can write assertions against the real HTTP surface instead of against helper functions.

## 🔍 Current state
- `e2e/tenant-isolation.spec.ts:49-64` — `process.env.MT_ENFORCE_ISOLATION = 'true'` set and restored inside the test process; the app under test is unaffected.
- `e2e/org-isolation.spec.ts:26-53` — creates two orgs inline (`Iso A` / `Iso B`) and three users, then asserts on `orgScoped()` directly. Each isolation spec re-invents its own two-tenant setup and its own cleanup.
- `e2e/helpers/db.ts` provides `prisma`, `seedUser`, `cleanupByEmail`, `uniqueEmail` — the building blocks exist, there is just no shared tenant fixture.
- `prisma/seed-demo.mjs` seeds a rich demo set but for **one** org, and refuses non-local `DATABASE_URL`s (`docs/DATA_ACCESS_POLICY.md`).
- `.github/workflows/e2e.yml` starts a MySQL service, runs `npx prisma db push` + `npx prisma db seed`, builds, then runs the `@smoke` subset.

## 🛠 What to build
1. `e2e/helpers/tenants.ts` — a fixture that creates two organizations (`iso-a-<stamp>`, `iso-b-<stamp>`) and, in each, an ADMIN, a MENTOR, a MENTEE, a COMPANY user, one `MentorshipRelation`, one `Tag`, one `PipelineStage`, one `InvitationToken` and one `Offer` — i.e. at least one row in every model the leak matrix will probe. Return typed handles (`{ orgA: { org, admin, mentor, mentee, relation, … }, orgB: {…} }`) and a `cleanup()` that removes users before orgs (FK order — see `e2e/org-isolation.spec.ts:16-23`).
2. Log in helpers for each tenant's actors, reusing whatever `e2e/helpers` already provides for authenticated storage state.
3. A Playwright **project** named `isolation` in `playwright.config.ts` that starts its own `webServer` with `MT_ENFORCE_ISOLATION: 'true'` in `env`, running only `e2e/isolation/**`. The default project keeps the flag off so the other ~150 specs are unaffected.
4. `package.json`: `"test:e2e:isolation": "playwright test --project=isolation"`.
5. A short section in `docs/testing.md` on how to run it locally and what it proves.

## ✅ Acceptance criteria
- [ ] `npm run test:e2e:isolation` boots a server whose `process.env.MT_ENFORCE_ISOLATION === 'true'` (assert it via `/api/health` or a trivial spec)
- [ ] The fixture creates two fully populated tenants and cleans up after itself, leaving no rows behind
- [ ] The default Playwright project still runs with the flag **off** and its results are unchanged
- [ ] Both tenants' actors can sign in
- [ ] `npm run build` passes
- [ ] no release fragment needed (test infrastructure) — state that in the PR

## 🇹🇷 Türkçe özet
Bugünkü izolasyon testleri ya bayrağı kendi süreçlerinin içinde çeviriyor ya da sorguları elle kapsamlıyor; ikisi de **çalışan uygulama** hakkında hiçbir şey kanıtlamıyor. Bu görev iki gerçek kiracı kuran paylaşılan bir fixture ve web sunucusu gerçekten `MT_ENFORCE_ISOLATION=true` ile açılan ayrı bir Playwright projesi ekliyor; böylece bir sonraki görev iddialarını yardımcı fonksiyonlara değil gerçek HTTP yüzeyine yazabilecek. Varsayılan proje bayrağı kapalı tutmaya devam ediyor, yani mevcut ~150 spec etkilenmiyor. İzole ve iyi tanımlı bir test altyapısı işi.

## 🤖 Agent prompt
```
Objective: build a shared two-tenant fixture and a Playwright project that runs
the app with MT_ENFORCE_ISOLATION=true.

Open first:
  e2e/helpers/db.ts                    prisma, seedUser, cleanupByEmail
  e2e/org-isolation.spec.ts:14-53      the inline two-org setup + FK-ordered
                                       cleanup to generalise
  e2e/tenant-isolation.spec.ts:49-64   the in-process flag flip to replace
  playwright.config.ts                 projects + webServer
  package.json                         the test:e2e:* scripts

Constraints:
- The DEFAULT project must keep MT_ENFORCE_ISOLATION off. ~150 specs assume the
  single-tenant behaviour; do not change it globally.
- Cleanup order matters: delete users before organizations (FK), and delete
  relation/tag/offer rows before their parents.
- Seed at least one row in every model the leak matrix will probe: User,
  MentorshipRelation, Tag, PipelineStage, InvitationToken, Offer, Company.
- Do not touch prisma/seed-demo.mjs — it is single-org by design and refuses
  non-local DATABASE_URLs (docs/DATA_ACCESS_POLICY.md). Synthetic data only.
- Claude Code web containers: run `npm install` first; if Playwright's pinned
  browser build is missing under /opt/pw-browsers, symlink the installed build
  into the expected version directory instead of running `playwright install`.
- Test infrastructure only: no release fragment, no i18n keys, no schema change.

Done when: `npm run test:e2e:isolation` boots with the flag on, the fixture
builds and tears down two tenants cleanly, and the default suite is unchanged.
```

**Zorluk** · M · **Tahmini** · 1 day · **Bağımlılık** · #1564

mersahin:
## Summary

Every isolation test we have proves something about a **helper function**, not about the running application. `e2e/tenant-isolation.spec.ts` sets `process.env.MT_ENFORCE_ISOLATION = 'true'` inside the Playwright process — which the app under test never reads — and `e2e/org-isolation.spec.ts` calls `orgScoped()` by hand and checks the rows Prisma hands back. Both pass green against a server that leaks every row across tenants.

This PR gives the suite two real tenants and a Playwright project whose web server genuinely boots with `MT_ENFORCE_ISOLATION=true`, so the next task can write its leak matrix against the real HTTP surface instead of against library calls.

The default project keeps the flag **off**: ~150 specs assume the single-tenant behaviour the flag switches off, and the flag is still un-flipped in production. `npx playwright test --list` reports the same 1058 tests in 398 files as on `main`.

Closes #1566

## Changes

- **`e2e/helpers/tenants.ts`** — `seedTwoTenants()` creates two organizations (`iso-a-<stamp>` / `iso-b-<stamp>`), each with an ADMIN, MENTOR, MENTEE and COMPANY actor plus a row in every model the leak matrix will probe: `Company`, `MentorshipRelation`, `Tag`, `PipelineStage`, `InvitationToken`, `Offer`. Returns typed handles (`tenants.orgA.company.id`, `tenants.orgB.admin`, …), a `signInAsTenantActor(page, actor)` that reuses the existing `signInAsFreshUser` guards, `remainingTenantRows()` for asserting teardown, and a `cleanup()` that deletes in foreign-key order (child rows → users → companies → organizations) and is idempotent.
  The doc comment states **which** of those models the middleware actually enforces today (`User`, `Company`, `MentorshipRelation`) and which are still in `check-tenant-models`' `PENDING_REGISTRATION` ratchet from #1560 (`Tag`, `PipelineStage`, `InvitationToken`, `Offer` → registered in #1559), so a leak found on one of the latter reads as a documented gap rather than a fresh regression.
  `PipelineStage` is the one model seeded as a **set** rather than a single row: each tenant gets the whole canonical catalogue, relabelled with its own prefix. `resolvePipelineStages()` returns a tenant's rows *instead of* the built-in catalogue as soon as one exists, so a lone custom stage would make both tenants custom-pipeline orgs whose catalogue does not contain the stage their own relation sits on — every board, funnel and pipeline bar would render empty and a leak assertion would pass in both directions without reading a row. The keys stay canonical and therefore identical across the two tenants (the shape a scoping bug surfaces in — `@@unique([orgId, key])` allows it); only the labels differ, so B's wording on A's board is a visible leak.
  Seeding is not transactional and the specs guard teardown with `tenants?.cleanup()`, so `seedTwoTenants()` records what it writes as it writes it and removes that again before rethrowing — a create that throws partway through tenant B cannot leave an orphan organization and eight `iso-*@e2e.local` users behind.
- **`playwright.config.ts`** — an `isolation` project with its own `webServer` on port **3010**, carrying `MT_ENFORCE_ISOLATION: 'true'` (and `NEXTAUTH_URL` pointed at that origin, or the post-sign-in redirect walks off to the other server). It runs only `e2e/isolation/**`, which the default `chromium` project now `testIgnore`s. Playwright's `webServer` is config-level rather than per-project, so each run assembles only the servers it needs: an ordinary `npm run test:e2e` and the CI smoke gate never pay for a second `next start`, and `--project=isolation` never boots the default one. Selecting the project with `BASE_URL` set throws with an explanation instead of quietly testing an unenforced deployment.
  The two projects are **mutually exclusive in one run**. Next has no per-port `distDir`, so two servers started from this working directory compile into the same `.next/` and overwrite each other's build manifests and webpack cache — intermittent `/_next/static/chunks/*` 404s that read as a flaky app. `E2E_ISOLATION=1` therefore means *isolation only* (the default project is dropped from the config), and naming both projects explicitly throws with an explanation.
- **`e2e/isolation/flag-on.spec.ts`** — proves the flag is really on the only way a test can: by observing behaviour that exists *only* then. `GET /api/companies` runs `prisma.company.findMany()` with no `where` inside `withTenantScope`, and `Company` is in `TENANT_MODELS` — so tenant A's admin sees tenant B's company **iff** the middleware is dormant. Asserted in both directions, plus a liveness probe so a server that failed to boot fails with a clear message.
- **`e2e/isolation/two-tenant-fixture.spec.ts`** — the fixture's own tests: both tenants fully populated and separate, each tenant's relation sits on a stage that tenant's own catalogue contains (asserted through the app's `resolvePipelineStages()`, so fixture and resolver cannot drift apart), all eight actors sign in through the real form and land on their role home, and `cleanup()` leaves exactly zero rows (exact counts, not "more than zero").
- **`.github/workflows/e2e-full.yml`** — an `isolation` job beside the four shards, so the project is actually executed by something. It cannot join the sharded job (it needs its own server, and the config refuses to assemble both projects in one run) and it is deliberately not in the PR gate (that gate is the `@smoke` subset). Same 4×/day cadence and drift gate; its JSON report is named `e2e-json-shard-isolation` so a red run reaches the same Turkish summary email, and `E2E_EXPECTED_REPORTS` is 5 so a job that crashes before writing one is caught too.
- **`package.json`** — `"test:e2e:isolation": "playwright test --project=isolation"`.
- **`e2e/global-setup.ts`** — seeds the cookie-consent `localStorage` for the second origin too; it is per-origin, so otherwise every isolation spec would start with the consent banner over the form.
- **Docs** — a section in `docs/testing.md` (how to run it, what it proves, where it runs in CI, why the default project is untouched, how to write a new isolation spec) plus a new table row, and a pointer from `docs/tenant-isolation.md`.

**No release fragment.** This is test infrastructure only — nothing user-visible ships, so per `releases/README.md` there is nothing to version. For the same reason there are no i18n keys and no schema change.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint's two `react-hooks/exhaustive-deps` warnings are pre-existing on `main`)
- [x] `npm run check:i18n`, `npm run check:release-fragments` clean
- [x] `npx playwright test --list` → 1058 tests in 398 files, same as `main`, with no `[chromium] › isolation/…` entries; `npx playwright test --project=isolation --list` → 7 tests in 2 files
- [x] Resolved config asserted directly: default run boots the google mock + `:3000`; `--project=isolation` and `E2E_ISOLATION=1` each boot only `:3010` with `MT_ENFORCE_ISOLATION=true`; `BASE_URL=… --project=isolation` throws; naming both projects throws
- [ ] The isolation suite itself has **not been executed** in this container — no database, no browser, so it was verified statically only. Its first real run is the `isolation` job in `e2e-full.yml` (or `npm run test:e2e:isolation` against a local MySQL).

## Contributor terms

Preview: https://pr2242.interncrm.com

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -2,7 +2,9 @@ name: E2E Full Suite (scheduled)
 
 # The PR gate (e2e.yml) runs only the @smoke subset; this workflow is the safety
 # net behind it (#621/#624): the FULL Playwright suite, 4× a day, 4-way sharded,
-# with a Turkish summary email when a run goes red — the failing tests with
+# plus the `isolation` project in a job of its own (#1566 — it needs a second
+# Next server with MT_ENFORCE_ISOLATION=true and cannot share a run with the
+# default project), with a Turkish summary email when a run goes red — the failing tests with
 # error snippets (scripts/e2e-report-email.mjs). Red-only is the default since
 # 2026-08-23 (maintainer request: green heartbeats were noise); set the repo
 # variable E2E_REPORT_MODE=always to bring the "✅ N/N test geçti" heartbeat
@@ -155,6 +157,102 @@ jobs:
           # detects the gap via E2E_EXPECTED_REPORTS instead.
           if-no-files-found: warn
 
+  # The `isolation` Playwright project (#1566), which no other gate runs.
+  #
+  # It cannot join the sharded job above: it needs its own Next server, booted
+  # with MT_ENFORCE_ISOLATION=true, and playwright.config.ts deliberately
+  # refuses to assemble both projects in one run (two servers compiling into the
+  # same .next/). The PR gate can't run it either — that gate is the @smoke
+  # subset and these specs are excluded from the default project. So it rides
+  # along here: same 4×/day cadence, same drift gate, and its JSON report is
+  # named `e2e-json-shard-isolation` so the summary email below picks it up like
+  # any other shard (a red isolation run therefore mails, and a crash that
+  # writes no report is caught by E2E_EXPECTED_REPORTS).
+  isolation:
+    name: Playwright tenant isolation
+    runs-on: ubuntu-latest
+    needs: [gate]
+    if: needs.gate.outputs.run == 'true'
+
+    services:
+      mysql:
+        image: mysql:8.0
+        env:
+          MYSQL_ROOT_PASSWORD: root
+          MYSQL_DATABASE: internship_test
+        ports:
+          - 3306:3306
+        options: >-
+          --health-cmd="mysqladmin ping -h 127.0.0.1 -uroot -proot"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=15
+
+    env:
+      DATABASE_URL: "mysql://root:root@127.0.0.1:3306/internship_test"
+      NEXTAUTH_SECRET: "ci-test-secret-not-for-production"
+      # The isolation webServer overrides this to its own origin (port 3010);
+      # left at the default, NextAuth's post-sign-in redirect walks off it.
+      NEXTAUTH_URL: "http://localhost:3010"
+      SEED_ADMIN_EMAIL: "admin@example.com"
+      SEED_ADMIN_PASSWORD: "ChangeMe123!"
+      ADMIN_EMAIL: "admin@example.com"
+      ADMIN_PASSWORD: "ChangeMe123!"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Apply DB schema
+        run: npx prisma db push
+
+      - name: Seed admin user
+        run: npx prisma db seed
+
+      - name: Build app
+        run: npm run build
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
+          restore-keys: playwright-${{ runner.os }}-
+
+      - name: Install Playwright browser
+        run: npx playwright install --with-deps chromium
+
+      - name: Run the isolation project
+        env:
+          PLAYWRIGHT_JSON_OUTPUT_NAME: e2e-results-shard-isolation.json
+        run: npx playwright test --project=isolation --reporter=list,html,json
+
+      - name: Upload Playwright report
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: playwright-report-isolation
+          path: playwright-report/
+          retention-days: 7
+
+      - name: Upload JSON result (for the summary email)
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: e2e-json-shard-isolation
+          path: e2e-results-shard-isolation.json
+          retention-days: 7
+          if-no-files-found: warn
+
   # Single Turkish summary email for the whole run — red-only by default since
   # 2026-08-23 (maintainer request: the earlier every-run heartbeat became
   # noise). Aggregates the per-shard JSON reports; a shard that crashed before
@@ -164,7 +262,7 @@ jobs:
   report:
     name: Summary email (red runs)
     runs-on: ubuntu-latest
-    needs: [gate, e2e-full]
+    needs: [gate, e2e-full, isolation]
     if: always() && needs.gate.outputs.run == 'true'
     steps:
       - uses: actions/checkout@v7
@@ -189,7 +287,8 @@ jobs:
           SMTP_FROM: ${{ secrets.SMTP_FROM }}
           ALERT_EMAIL_TO: ${{ secrets.ALERT_EMAIL_TO || 'mehmetersahin@gmail.com' }}
           PLAYWRIGHT_JSON_REPORT: e2e-json-reports
-          E2E_EXPECTED_REPORTS: '4'
+          # 4 shards + the isolation job, each writing one JSON report.
+          E2E_EXPECTED_REPORTS: '5'
           E2E_REPORT_MODE: ${{ vars.E2E_REPORT_MODE || 'failures' }}
           E2E_RUN_CONTEXT: 'e2e-full · ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}'
         run: |
--- a/docs/tenant-isolation.md
+++ b/docs/tenant-isolation.md
@@ -62,6 +62,14 @@ All of this is exercised against a real DB by `e2e/org-isolation.spec.ts` and
 never called `orgScoped()`** is still isolated purely because it ran inside
 `runWithOrg()` with the flag on, and is a no-op with the flag off.
 
+Both of those run **in the Playwright process**, though: they flip
+`MT_ENFORCE_ISOLATION` in their own env and call the helpers directly, so they
+say nothing about the deployed server. The `isolation` Playwright project
+(#1566) covers that half — `npm run test:e2e:isolation` boots a second app
+server on port 3010 with the flag genuinely on and runs `e2e/isolation/**`
+against it, using the two-tenant fixture in `e2e/helpers/tenants.ts`. See
+[`docs/testing.md`](testing.md#tenant-isolation-the-isolation-playwright-project-1566).
+
 ### Keeping the registry honest (`npm run check:tenant-models`)
 
 The middleware only scopes models that are named in `TENANT_MODELS`. An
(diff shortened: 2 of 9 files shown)


mersahin:
Review findings applied in f2e1f9f. All four were real; nothing was rejected.

**1 — major, `e2e/helpers/tenants.ts`: the single `PipelineStage` row emptied both tenants' boards.** Confirmed. `resolvePipelineStages()` / `resolveCustomStages()` (`src/lib/pipelineStages.ts:22-36`, `43-55`) return a tenant's custom rows *instead of* the default catalogue the moment one row exists, so the seeded `ISO_STAGE` row made the resolved set for both orgs exactly `['ISO_STAGE']` — which does not contain `INTERNSHIP_IN_PROGRESS_450`, the stage the fixture's own relation sits on. `/mentor/board` renders `stages.map(…)` with `itemsFor(s.key)` and `/admin`'s pipeline overview iterates the same set, so both tenants would have shown one empty column and #1559's leak matrix would have passed vacuously in both directions.

Took the third option from the suggested fix (full catalogue per org), because it is the only one that keeps *all* of the fixture's properties: each tenant now gets the canonical stage set, relabelled with its own prefix. The keys stay canonical and therefore identical across the two tenants — which is the `@@unique([orgId, key])` property the fixture set out to exercise — while the labels differ, so B's wording appearing on A's board is itself a visible leak. `two-tenant-fixture.spec.ts` gained the drift guard you asked for, asserted through the app's own resolver rather than by re-listing keys:

```ts
const stages = await resolvePipelineStages(tenant.org.id);
expect(relation.pipelineStatus).toBe(TENANT_RELATION_STAGE);
expect(stages.map((s) => s.key)).toContain(relation.pipelineStatus);
```

**2 — minor, `playwright.config.ts`: two `next dev` processes in one working directory.** Confirmed: `runsIsolation` was true for `E2E_ISOLATION=1` while `runsDefault` stayed true, so the invocation the comment itself advertised started both servers from the same cwd into the same `.next/`.

Did not go the `distDir` route — the isolation server runs `npm run start` in CI, so a separate dist dir would need a second `next build` and the config change would land in production code for a test convenience. Instead `E2E_ISOLATION=1` now means *isolation only*: the `chromium` project is dropped from the config, not just its server, so `E2E_ISOLATION=1 playwright test` with no `--project` cannot point 1000+ specs at a `:3000` that was never started. Naming both projects explicitly (`--project=isolation --project=chromium`, or `E2E_ISOLATION=1 --project=chromium`) now throws with an explanation instead of starting the colliding pair. Verified against the real config: default run still lists **1058 tests in 398 files**, `--project=isolation` and `E2E_ISOLATION=1` each list 7 tests in 2 files, the combined form throws.

**3 — minor, non-atomic `seedTwoTenants()` paired with `tenants?.cleanup()`.** Confirmed. `seedTwoTenants()` now threads a scratch record through seeding (org ids and addresses recorded as they are written) and removes it on the failure path before rethrowing. The teardown was refactored so both paths share one deletion routine — `removeSeededRows({ orgIds, emails })` — rather than duplicating the FK order. The rethrown error is always the seeding failure: the cleanup is best-effort and deliberately silent, so a teardown that also fails cannot mask it.

**4 — minor, no CI job constructs the `isolation` project.** Confirmed: neither `e2e.yml` (`--grep "$E2E_GREP"`) nor `e2e-full.yml` (`--shard=N/4`) selects it, and the `testIgnore` on chromium made `e2e/isolation/**` unreachable from every gate. Took the first option rather than only documenting it: `e2e-full.yml` now has an `isolation` job beside the four shards — same drift gate, same 4×/day cadence, its own MySQL service. It cannot join the sharded job (it needs its own server, and per finding 2 the config now refuses to assemble both projects in one run), and it stays out of the PR gate, which is the `@smoke` subset. Its JSON report is named `e2e-json-shard-isolation`, which matches the report job's existing `e2e-json-shard-*` download pattern, so a red isolation run reaches the same Turkish summary email; `E2E_EXPECTED_REPORTS` went to `5` so a job that crashes before writing a report is caught as red rather than silently dropped.

`docs/testing.md` records all four. Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` clean (same two pre-existing `react-hooks/exhaustive-deps` warnings), `npm run check:i18n` 4341 keys × 3 locales, `npm run check:release-fragments` exit 0 (its "compaction looks stuck" notice is the pre-existing 23-fragment backlog on `main`, not this PR). No schema change, so no `prisma validate`. The suite still has not been run against a database in this container — that first green run is now the `isolation` job's job.

---

Issue #2142, opened 2026-09-03
Release compaction has failed every run since 2026-08-24 — 57 fragments, 25 versions missing from CHANGELOG


mersahin:
## Symptom

Many PRs have merged, but their changes never appear in `CHANGELOG.md` or `src/lib/releaseNotes.ts`. The last recorded section is `## [0.110.1-beta] - 2026-08-24`, while `releases/unreleased/` holds **57 pending fragments** spanning `0.110.2-beta` → `0.135.0-beta`.

The *app* was never wrong: `next.config.js` derives the version and the release-notes entries from base+fragments at build time (`releases/README.md` → "How the version stays correct"), so `/release-notes`, the sidebar footer and `/api/health` all showed the right thing. What broke is **the record** — the committed canonical files stopped reflecting what shipped.

## Root cause — two failures, chained

`.github/workflows/release-compact.yml` has failed **all 10 of its runs** (run #1 2026-08-24 → run #10 2026-09-02).

**1. `RELEASE_BOT_TOKEN` is not set.** Runs 1–7 compacted correctly, pushed the branch, then died at `gh pr create`:

```
pull request create failed: GraphQL: GitHub Actions is not permitted to create
or approve pull requests (createPullRequest)
```

The workflow's own header documents this requirement; the `secrets.RELEASE_BOT_TOKEN || github.token` fallback silently used `github.token`, which this org forbids from opening PRs.

**2. The branch name made that failure permanent.** The branch was `bot/release-compact-${VERSION}` — derived from the fragment set, so it is *the same name* on every retry, while `git commit` produces a *different sha* each run. Runs 8–10 therefore never even reached the PR step:

```
! [rejected]  bot/release-compact-0.128.2-beta -> bot/release-compact-0.128.2-beta (non-fast-forward)
hint: Updates were rejected because the tip of your current branch is behind its remote counterpart.
```

Two orphaned branches are the residue: `bot/release-compact-0.126.0-beta`, `bot/release-compact-0.128.2-beta`.

**3. Nothing ever said so.** The workflow had no failure notification, unlike `e2e-full`, `k6-load`, `stress` and `backup-verify`, which all route through `scripts/send-alert-email.mjs`. Ten daily red runs went unnoticed for ten days.

## Fix

- **Backfill**: run `scripts/release-compact.mjs` — 57 fragments → 57 CHANGELOG sections, 51 release-notes entries, `package.json` → `0.135.0-beta`.
[...]

PR #2244, opened 2026-09-07
feat: optional screenshot or short clip on a release note (#2233)

### LINKED ISSUE #2233: 📸 [UX] Task · Optional screenshot on a release note — captured by Playwright, carried by the release fragment
## 🎯 Goal

A release note reads "Mentees can now request a mentor straight from the directory card." A reader who has not seen the screen still does not know what changed. One small picture answers it instantly — and Playwright is already driving that exact screen to verify the change, so the picture is nearly free.

Make a screenshot — or, where motion actually explains something, a short clip — an **optional** part of a release fragment: captured by a dedicated spec, validated in CI, rendered small on `/release-notes` and linked from `CHANGELOG.md`.

## 🔍 Current state

- `src/lib/releaseNotes.ts:9-15` — `ReleaseNote` is `{ version, date, time?, commit?, highlights: Record<Locale, string[]> }`. Text only.
- Notes are never hand-written: a PR ships `releases/unreleased/<slug>.json`, `next.config.js` inlines the pending fragments as `APP_UNRELEASED_ENTRIES`, and `release-compact.yml` later folds them into `releaseNotes.ts` and `CHANGELOG.md` (#1275, #1457).
- `scripts/check-release-fragments.mjs` validates the fragment shape in CI.
- `public/` currently holds only icons and the service worker — there is no media directory.
- The e2e suite deliberately does **not** assert on screenshots (`e2e/mobile-chat-layout.spec.ts:11`: *"Geometric assertions (bounding boxes / scroll metrics), not screenshots"*). Capturing an image for documentation is a different thing from screenshot-diffing, and this task must not blur that line.

## 🛠 What to build

1. **Fragment field.** An optional `media` on the fragment, in one of two shapes:

   ```json
   "media": {
     "poster": "release-media/request-mentor-cta.png",
     "video": "release-media/request-mentor-cta.webm",
     "alt": { "en": "…", "tr": "…", "de": "…" }
   }
   ```

   `video` is optional; `poster` never is (see **Why a still is always required** below). `alt` is all three locales or none, same rule as `notes`.

   Extend `check-release-fragments.mjs`: every referenced file must exist under `public/`, `poster` must be a `.png` under **150 KB** (a cropped element screenshot is 40-80 KB, so the cap only catches a full-page capture pasted in by mistake), `video` must be a `.webm` under **1.5 MB** and **at most 5 seconds** long.

2. **A dedicated capture spec**, `e2e/release-media.spec.ts`, **not tagged `@smoke`** and excluded from the PR gate. Author runs it locally (or via `workflow_dispatch`) and commits the media with the PR.
   - Stills use an **element** screenshot (`locator.screenshot()`), never `page.screenshot()` — a cropped component stays small and does not churn every time unrelated page chrome moves.
   - Clips use Playwright's own `recordVideo` (native `.webm`, no `ffmpeg`). Video is recorded **per BrowserContext, whole viewport, for the whole test** — Playwright cannot crop to an element or trim to the interesting seconds without external tooling. So do not try to trim: keep each capture test minimal so the whole test *is* the thing worth showing, and set `recordVideo.size` to something modest rather than the default viewport.

3. **Rendering on `/release-notes`.** `max-width: 480px`, rounded, visible border, `width`/`height` set so the page does not reflow, localized `alt`.
   - Still only → ``.
   - With a clip → ``'&lt;video autoplay loop muted playsinline poster="…"&gt;'`` with the poster as the fallback child, and **`@media (prefers-reduced-motion: reduce)` renders the poster instead of the video** — not merely a paused video.

4. **`ReleaseNote` gains `media?`**, carried through `next.config.js`'s inlining and through the compaction workflow so a fragment's media survives into the permanent entry.

5. **Docs**: a section in `releases/README.md` — when media is worth adding, still vs clip, how to capture each, the caps, and the naming convention (`release-media/<fragment-slug>.{png,webm}`).

## 🧭 Decisions already taken (change them deliberately, not by accident)

- **WebM for motion, never GIF.** Playwright records `.webm` natively with no `ffmpeg` dependency, and a WebM is roughly a tenth of the bytes of the same clip as a GIF. GIF has no advantage here.
- **A still is always required, a clip never is.** Three separate surfaces need it and none of them can play video: `CHANGELOG.md` is markdown; Safari's WebM support is recent enough that an older iOS PWA install may not decode it; and `prefers-reduced-motion` must have something to show. One poster PNG serves all three, so "clip + poster" is the natural shape rather than extra work. Most notes will ship the poster alone.
- **Clips are capped at 5 seconds.** WCAG 2.2.2 (Pause, Stop, Hide) requires automatically-moving content longer than five seconds to be pausable. Keeping clips under that threshold avoids needing a control on every card, and the size cap follows from it. This repo has an axe baseline and just cleared its contrast violations in #2131 — do not reintroduce a violation through the release notes page.
- **Committed to `public/release-media/`, not uploaded to R2.** R2 exists (off-site backups) but putting release media there adds a deploy-time dependency and a broken-image failure mode to a page that must always render. Cost of the repo route: ~60 KB × ~30 user-visible releases a month ≈ **20 MB a year** of stills, bounded by the caps; clips will be rare.
- **Light theme only, framed.** Two themes means two files for every note. A light capture inside a distinct border reads as *a screenshot* rather than as part of the page, so it stays legible on the dark `/release-notes`.
- **Optional, permanently.** A backfill, a CI guard or an API fix has no screen to show. A fragment without `media` must stay completely valid — this is never a checklist item that blocks a PR.

## ⚠️ Accepted trade-off

**This media goes stale.** A note from three months ago will show a screen that no longer looks like that. For a historical changelog that is correct behaviour, not a bug — every entry already carries its version, date and commit. Do not build a re-capture job; do not let stale media become a CI failure.

## ✅ Acceptance criteria

- [ ] A fragment with a valid `media` renders on `/release-notes` in all three locales, with localized `alt`
- [ ] A fragment without `media` is still valid and renders exactly as it does today
- [ ] A fragment with a `video` but no `poster` is rejected by `check:release-fragments`
- [ ] `check:release-fragments` also fails on a missing file, a wrong extension, an oversized still or clip, a clip over 5s, or `alt` in fewer than three locales
- [ ] With `prefers-reduced-motion: reduce`, the poster renders and the video does not play
- [ ] The media survives compaction into `releaseNotes.ts` and `CHANGELOG.md`; the markdown reference is the poster
- [ ] The capture spec does not run in the PR smoke gate
- [ ] Media does not shift layout while loading, and is legible in dark mode
- [ ] The `/release-notes` page still passes the axe scan with no new baseline entries
- [ ] `npm run build`, `npm run check:i18n` and `npm run check:release-fragments` pass
- [ ] a release fragment is added — **using the new field, so the feature demonstrates itself**

## 🇹🇷 Türkçe özet

Bir sürüm notu "Mentee'ler artık dizin kartından doğrudan mentor talep edebiliyor" diyor; o ekranı görmemiş biri hâlâ neyin değiştiğini bilmiyor. Küçük bir görsel bunu anında anlatıyor — üstelik Playwright zaten o ekranı sürüyor, yani görsel neredeyse bedava.

Sürüm fragment'ine **opsiyonel** bir `media` alanı ekliyoruz: ayrı bir spec yakalıyor, CI doğruluyor, `/release-notes` küçük ve çerçeveli gösteriyor.

Hareket gerçekten bir şey anlatıyorsa **webm** — GIF değil: Playwright webm'i yerel olarak üretiyor (`ffmpeg` bağımlılığı yok) ve aynı klip GIF'in onda biri kadar yer tutuyor. Ama **hareketsiz kare her zaman zorunlu, klip hiçbir zaman değil**: `CHANGELOG.md` markdown olduğu için video oynatamıyor, eski iOS PWA kurulumları webm çözemeyebiliyor, ve `prefers-reduced-motion` diyene gösterilecek bir şey lazım. Tek bir poster PNG üçünü birden karşılıyor.

Klipler **5 saniyeyi geçmiyor**: WCAG 2.2.2 daha uzun otomatik hareketi durdurulabilir olmaya zorunlu kılıyor, bu repo'nun da axe baseline'ı var (#2131'de kontrast ihlallerini yeni temizledik) — sürüm notları sayfasından yeni bir ihlal girmesin.

Tam sayfa değil element görüntüsü — küçük kalır, alakasız değişikliklerde bozulmaz. Mevcut fonksiyonel spec'lere iliştirilmiyor: testin işi doğrulamak, medya üretmek değil; smoke gate yavaşlamamalı.

Zorunlu **değil** ve hiçbir zaman olmayacak: backfill'in, CI guard'ın, API düzeltmesinin gösterilecek ekranı yok. Ve medyanın eskiyeceğini baştan kabul ediyoruz — sürüm notu tarihsel bir kayıt, her girdi zaten sürümünü ve tarihini taşıyor.

mersahin:
## Summary

A release note that says *"mentees can now request a mentor straight from the directory card"* still leaves a reader who has never seen that screen in the dark. Playwright is already driving that exact screen, so the picture is nearly free. This makes a screenshot — or, where motion is the point, a short clip — an **optional** part of a release fragment: captured by a dedicated spec, validated against the actual bytes in CI, rendered small and framed on `/release-notes`, and linked from `CHANGELOG.md`.

Optional **permanently**: a backfill, a CI guard or an API fix has no screen to show, so a fragment without `media` stays completely valid and this never becomes a checklist item that blocks a PR.

**This PR's own fragment carries media**, captured from the running app — so the feature demonstrates itself on the preview and the three rendering tests actually run.

Closes #2233

## Changes

- **`scripts/release-media.cjs`** (new) — the rule. Shape validation (`poster` required whenever `media` is present, `video` always optional, `alt` in all three locales or none, paths pinned to `release-media/<slug>.{png,webm}` so nothing can climb out of the directory) plus asset validation against the real files: PNG `IHDR` for the intrinsic size, the **EBML header** for the duration and for the clip's own `PixelWidth`/`PixelHeight`. No `ffmpeg` — it is not installed in CI, and Playwright produces WebM natively. Duration comes from `Segment > Info > Duration` when present, otherwise from the last `Cluster` timecode (a slight *under*-estimate, never an over-estimate, which is the safe direction for a cap). Caps: poster **150 KB**, clip **1.5 MB** and **5 s** (WCAG 2.2.2 — staying under the threshold keeps a pause control off every card). It also owns the two **compaction writers**, so they are importable and testable.
- **`scripts/release-derive.cjs`** — `media` is a known fragment field; `resolveRelease()` resolves the assets (stamping the poster's `width`/`height` and the clip's `videoWidth`/`videoHeight` onto the entry) and the timeline/`unreleasedEntries` carry the block through to `APP_UNRELEASED_ENTRIES`. When `public/` is absent (a partial checkout) the asset checks are skipped so a build can still derive its version; CI always has `public/`, which is where the gate has to hold.
- **`scripts/release-compact.mjs`** — media survives compaction: the whole block (both pixel sizes included) goes into `src/lib/releaseNotes.ts`, and the **poster** is written into `CHANGELOG.md` as a markdown image, because markdown cannot play a video.
- **`src/components/ReleaseNoteMedia.tsx`** (new) + `/release-notes` — `max-width: 480px`, rounded, visible border, white background pinned even in dark mode (`dark:!bg-white`, since `globals.css` retints `.bg-white` under `html.dark`) so a light capture reads as *a screenshot* rather than as a broken part of the page. The `` is sized from the PNG header and the `<video>` from the WebM's own header, so the list does not reflow and neither picture is letterboxed inside the other's aspect ratio. Still only → ``; with a clip → ``'&lt;video autoplay loop muted playsinline poster=…&gt;'`` with the poster as the fallback child. Under `prefers-reduced-motion: reduce` the **poster renders and no `<video>` is mounted at all** — not a paused video, and the WebM is never even fetched. Server render is the poster too, so a JS-less browser gets the still.
- **`e2e/release-media.spec.ts`** (new) — the capture spec: element screenshots (`locator.screenshot()`, never `page`) and Playwright's own `recordVideo`. It is a **producer, not a test**, and does not run in CI: the default `chromium` project `testIgnore`s it and the `release-media` project only exists when `CAPTURE_RELEASE_MEDIA` is set (`npm run test:e2e:media`). No trimming is attempted — video is per-context and whole-viewport, so the templates keep the test minimal, warm the route *before* the recording context exists, and set a modest `recordVideo.size` instead.
- **`e2e/release-notes-media.spec.ts`** (new) — the rendering side (size, frame, reserved space, localized `alt`, reduced-motion behaviour). Not `@smoke`. It skips itself while no note carries media — but with this PR's own fragment carrying one, all three run.
- **`scripts/test/release-media.test.mjs`** (new, wired into `npm run test:release`) — 30 tests, every fixture built byte by byte (real CRC'd PNGs, real EBML), so the suite carries no binaries and each rule is proven against a file that really has the property.
- **Docs** — a full section in `releases/README.md` (when media is worth adding, still vs clip, how to capture, the caps and why each exists, the naming convention, where it ends up, and the accepted staleness trade-off), plus `CLAUDE.md`'s versioning bullet.

### Decisions kept as the issue set them

WebM never GIF · a still always required, a clip never · 5 s cap for WCAG 2.2.2 · committed to `public/release-media/` rather than R2 · light theme only, framed · optional forever · **media goes stale and that is correct** — no re-capture job, and stale media is never a CI failure.

## Review round 2 — the four findings, all fixed

| # | Finding | Fix |
|---|---|---|
| **major** | The fragment carried no `media`, `public/release-media/` did not exist, so `ReleaseNoteMedia` was dead code everywhere, all three render tests skipped permanently and the compaction writers had no coverage | The AC is met: real capture committed, fragment carries the block, the three tests pass, the writers are unit-tested and the compaction was run for real |
| minor | The `<video>` inherited the **poster's** pixel size, so a clip rendered letterboxed and snapped size when autoplay began | `webmVideoSize()` reads `PixelWidth`/`PixelHeight` (and `DisplayWidth`/`Height` where present) out of `Tracks > TrackEntry > Video`; the entry carries `videoWidth`/`videoHeight` and the `<video>` uses those |
| minor | `expect(new Set([en, tr, de]).size).toBeGreaterThan(1)` fails the scheduled suite for a legitimate alt that reads the same in three languages | Dropped. The test asserts a non-empty alt per locale and proves the locale switch took by checking `<html lang>` — a real signal, no false failures |
| minor | The clip template starts recording before `goto()`, so a cold route compile lands in the WebM and can blow the 5 s cap | The route is warmed in the unrecorded `page` fixture first; the constraint is documented in the template, in `releases/README.md` and in the cap's own error message. The hand-built context also now passes `baseURL`, which it never inherited from `use` |

**The container turned out to have what the first pass assumed it lacked.** A Chromium sits under `/opt/pw-browsers`, and `next dev` serves `/release-notes` with a dummy `DATABASE_URL` because the page reads only `releaseNotes.ts` plus the fragments — no Prisma. That was enough for a genuine capture, and the assumption is corrected in `CLAUDE.md` and the retrospective so the next session does not repeat it.

## Verification

- [x] `npm run lint` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings in unrelated admin pages)
- [x] `npm run check:i18n`
- [x] `npm run check:release-fragments` — this PR ships as `0.162.0-beta` (the pending-backlog warning is pre-existing, #2142)
- [x] `npm run test:release` — 46/46 (16 existing + 30 media)
- [x] **`e2e/release-notes-media.spec.ts` — 3/3 passing** against the app, with this PR's own media on the page (no skips)
- [x] **`npm run test:e2e:media` — 2/2**, both templates; the recorded clip parses back as 640×400, 3.12 s
- [x] **Compaction run for real** on a scratch copy: 24 fragments folded, the markdown image lands in `CHANGELOG.md`, the `media` block lands in `src/lib/releaseNotes.ts`, and `npx tsc --noEmit` on the result is clean

The validator's rejection messages were also exercised by hand against byte-built fixtures (clip with no poster, 469 KB poster, 9 s clip, missing file, wrong extension, two-locale alt) — each fails with its own message and exit 1.

## Still deliberately out of scope

`/release-notes` is **not** part of the axe scan set (`e2e/a11y-scan.spec.ts` covers `/`, `/auth/signin`, `/accessibility`, `/apply/:mentorId` and the role contexts), so **no a11y baseline entries change** in this PR. Adding that page to the scan is a separate, wider change. The RSS feed also still carries text only — the issue does not ask for an `<enclosure>`, and adding one is a separate decision about what a feed reader should download.

Preview: https://pr2244.interncrm.com

## Contributor terms


mersahin:
Review findings applied in c80b742. **All four were real; none rejected.**

**major — the PR did not use its own feature.** Confirmed: no `media` block on the fragment, no `public/release-media/`, so `ReleaseNoteMedia` was the null branch everywhere, the three render tests skipped permanently, and the compaction writers had zero coverage.

The stated reason for that gap — "no browser and no database in this container" — was wrong, and I checked before accepting it. There is a Chromium under `/opt/pw-browsers` (symlink the revision the pinned Playwright expects, or pass `executablePath`), and `next dev` starts with a dummy `DATABASE_URL`: `/release-notes` returns 200 because the page touches nothing but `releaseNotes.ts` and the fragments. `/features` too; the landing page `/` genuinely does not, it reads public stats.

So the capture was possible. `public/release-media/release-note-media.png` (30 KB) is a real `locator.screenshot()` of the release card on the running app, taken in two passes so the picture shows a note *with* a picture: capture the card, add the `media` block, restart (fragments are read in `next.config.js` at config load), re-capture the card that is now displaying it. Consequences:

- `e2e/release-notes-media.spec.ts` — **3/3 passing**, no skips.
- `npm run test:e2e:media` — **2/2**, and the recorded WebM parses back as 640×400 / 3.12 s, which is also a real-file check of the new EBML code.
- Compaction — run for real on a scratch copy of the repo: 24 fragments folded, the markdown image lands in `CHANGELOG.md`, the `media` block lands in `src/lib/releaseNotes.ts`, and `npx tsc --noEmit` on the generated file is clean. Both writers also moved into `release-media.cjs` (a top-level script cannot be imported) and now have 12 unit tests, one of which evaluates the generated TypeScript to prove it parses with the `alt` intact.

**minor — `<video>` inheriting the poster's aspect ratio.** Real, and the arithmetic in the finding is right. `scripts/release-media.cjs` now reads `PixelWidth`/`PixelHeight` — preferring `DisplayWidth`/`DisplayHeight` when a muxer writes them — from `Tracks > TrackEntry > Video`, gated on the parent element so a one-byte ID like `0xB0` cannot be matched somewhere else in the file. The entry carries `videoWidth`/`videoHeight` and the `<video>` uses those; an older compacted entry without them gets no size attribute at all, which beats the poster's wrong one.

Option (b) from the suggested fix (make the recorder match the poster) is not enforceable: Playwright letterboxes the picture when `recordVideo.size` and the viewport disagree, so an aspect rule would force the poster to be a full-viewport capture — exactly what the 150 KB cap exists to prevent. The guidance is documented instead, in the template and in `releases/README.md`.

**minor — the alt-locale assertion.** Real false-failure generator: `validateMediaShape()` requires three non-empty strings, not three *different* ones, and "Kanban" / "Dashboard" / "CV Upload" are the same word in all three. Dropped. The test asserts a non-empty alt per locale and proves the switch actually happened via `<html lang>`, which is the thing that would really be broken. Not moved into `check:release-fragments` either — with an allowance for legitimately identical strings it would only ever be a warning, and there is already a reviewer looking at the fragment.

**minor — the clip template recording the page load.** Real. `recordVideo` starts at `newContext()`, so the navigation and the cold-compile of the route are in the file; the warm run measures 3.12 s for a 1.2 s interaction, and a cold `/release-notes` compile here is 4–7 s on its own. The route is now warmed in the unrecorded `page` fixture first, the constraint is stated in the template, in `releases/README.md` and in the cap's own error message (which used to say "keep the capture test shorter" and pointed at the wrong thing).

While there: that same hand-built context never passed `baseURL`, which it does not inherit from `use` — the relative `goto()` in the template would have thrown for the first author who copied it. Fixed. The two template captures are now gitignored so a `npm run test:e2e:media` run cannot smuggle 100 KB of `example-*` bytes into someone's PR.

`tsc`, `lint`, `check:i18n`, `check:release-fragments` and `test:release` (46/46) are all green.

---

Issue #1886, opened 2026-08-30
🌱 🧪 [Testing] Task · check:stage-keys CI guard and a renamed-pipeline e2e fixture


mersahin:
## 🎯 Goal
This bug has been reintroduced repeatedly because a hardcoded stage key is invisible: it type-checks, it passes every test on the default seed, and it only misbehaves for a customer who renamed their stages — which no test tenant does. Two cheap mechanical guards make the class of bug impossible to ship again.

## 🔍 Current state
- The repo already runs this style of guard from CI: `npm run check:i18n` (EN/TR/DE key parity) and `npm run check:release-fragments` are both plain Node scripts wired into `.github/workflows/ci.yml`; `npm run check:k6` exists for the same reason ("without it a mistyped threshold key first surfaces at 23:40 UTC as a crash email" — CLAUDE.md).
- There is no fixture anywhere that seeds a second `PipelineStage` set. `resolvePipelineStages()` (`src/lib/pipelineStages.ts:18`) falls back to the canonical defaults whenever an org has no rows, so **every existing test exercises only the fallback path**.
- The allowed remaining literals after this story: `src/lib/pipeline.ts` (the canonical default catalogue) and `src/lib/offers.ts:85` (`DEFAULT_HIRED_STAGE_KEY`, documented as the default).

## 🛠 What to build
1. `scripts/check-stage-keys.mjs` + `"check:stage-keys"` in `package.json`: grep `src/` for `HIRED_660|EMPLOYED_700|APPLICATION_100|INTERNSHIP_DROPPED_460|INTERNSHIP_FOUND_ELSEWHERE_800` and fail with a per-hit file:line report. Allowlist exactly two files (`src/lib/pipeline.ts`, `src/lib/offers.ts`) in a constant at the top of the script, with a comment saying why each is allowed. Ignore comments and strings inside `//`-prefixed lines. Wire it into `.github/workflows/ci.yml` next to `check:i18n`.
2. A Playwright fixture that creates a second org with six `PipelineStage` rows whose keys are `STAGE_A`…`STAGE_F` (none matching a canonical key) and one relation seeded onto `STAGE_F`.
3. `e2e/custom-pipeline-analytics.spec.ts` — signed in as that org's admin, assert that `/admin/analytics` shows a non-zero conversion figure and that the funnel names `STAGE_F` as the finished stage. Tag it `@smoke`: a green smoke run must mean the paid reports work for a custom-pipeline tenant.

## ✅ Acceptance criteria
[...]

Issue #1643, opened 2026-08-30
📄 [Admin] Task · Clone a programme for the next intake — configuration only, never participants


mersahin:
## 🎯 Goal
Every autumn an internship coordinator sets up the same programme again by hand across five admin screens. Mentornity clones a programme for the next cohort; we cannot duplicate anything. "Clone last year's intake" is the single most-used button in a cohort-based product, and it is cheap once the Program object owns its configuration.

## 🔍 Current state
- No duplication of anything exists. `POST /api/cohorts` creates `{ name, term }` and nothing else — `src/app/api/cohorts/route.ts:29-41`.
- Configuration that will need copying, all currently org-keyed and reachable by `orgId` only:
  - `PipelineStage` — `prisma/schema.prisma:192-207`; whole-set replace at `src/app/api/admin/organizations/[id]/pipeline-stages/route.ts:69-84`
  - `StageSla` — `prisma/schema.prisma:176-190`; upsert-per-key at `src/app/api/admin/stage-sla/route.ts:84-90`
  - `DocumentRequirement` — `prisma/schema.prisma:976-996` (`appliesToStage`, `mandatory`, `order`, `active`)
  - `EvaluationTemplate` — `prisma/schema.prisma:1865`
- `MentorshipRelation` rows point at the programme via `cohortId`/`programId` — `prisma/schema.prisma:1688` — so a naive `include`-based copy would drag participants along.

## 🛠 What to build
1. `POST /api/admin/programs/[id]/clone` with a body of `{ name, term?, startsAt?, endsAt? }`. Returns the new programme as **DRAFT** — a clone is never live on creation.
2. Copy in one `prisma.$transaction`:
   - all `PipelineStage` rows for the source programme (or the org-level set when the source has none of its own), re-keyed to the new `programId`
   - all `StageSla` rows for the source programme
   - `DocumentRequirement` rows scoped to the source programme, once they carry a programme (out of scope if they are still org-only — copy nothing rather than duplicating org-wide rows)
   - the programme's own scalar config: `kind`, `description`, `reminderDays`, enrolment settings (from #1647)
3. Copy **nothing** that identifies a person: no relations, no users, no mentor assignments, no interaction logs, no messages, no documents, no evaluations. Assert this in the test — the failure mode here is a GDPR incident, not a bug.
4. `slug` is regenerated from the new name, not copied (`@@unique([orgId, slug])`).
[...]

PR #2247, opened 2026-09-07
feat: programme template catalogue in src/lib/programTemplates.ts (#1641)

### LINKED ISSUE #1641: 🌱 [Admin] Task · Programme template catalogue in src/lib/programTemplates.ts (EN/TR/DE)
## 🎯 Goal
Ship the ready-made programme shapes: a graduate internship pipeline, an onboarding buddy programme, a leadership/mentoring cohort and a career-transition programme, each with its own stage set, its own SLAs and its own reminder cadence. Pure data and pure functions — no schema change, no new table — which makes this the safest, most self-contained piece of the templates story and a good first issue.

## 🔍 Current state
- No template concept exists anywhere. The only fallback is the single canonical set built from the `PipelineStatus` enum — `defaultPipelineStages(locale)`, `src/lib/pipeline.ts:222-233`, used by `resolvePipelineStages` at `src/lib/pipelineStages.ts:35`.
- The repo already has the exact pattern to copy: a curated, localised, code-level library with a documented house style — `src/lib/newsletterContent.ts` (see also `docs/newsletter.md`) — and a smaller one at `src/lib/goalTemplates.ts`.
- Stage rows need `key`, `label`, `order`, `isTerminal`, `isOffPath`, `color`, and (after #1637) `phase` — the shape is `ResolvedStage` in `src/lib/pipeline.ts:215-222`.
- Stage keys are validated as `z.string().regex(/^[A-Za-z0-9_]+$/)`, max 60 chars, max 50 stages per set — `src/app/api/admin/organizations/[id]/pipeline-stages/route.ts:33-40`.
- SLA rows are `{ stageKey, days }` in calendar days — `src/lib/stageSla.ts:18-30`, `prisma/schema.prisma:176-190`.

## 🛠 What to build
1. New file `src/lib/programTemplates.ts` — **client-safe**: no Prisma import, so the wizard can render the catalogue without a server round-trip (same rule that keeps `src/lib/pipeline.ts` importable from client components).
2. Shape:
   ```ts
   export interface ProgramTemplate {
     key: string;                       // 'graduate_internship' | 'onboarding_buddy' | …
     stages: Omit<ResolvedStage, 'label'>[] & { labels: Record<Locale, string> }[];
     slas: { stageKey: string; days: number }[];
     reminderDays: number;              // the programme's default nudge cadence
     documentRequirements?: { key: string; labels: Record<Locale, string>; appliesToStage?: string; mandatory: boolean }[];
   }
   export const PROGRAM_TEMPLATES: ProgramTemplate[];
   export function programTemplate(key: string): ProgramTemplate | null;
   ```
   Names and descriptions for each template go in a `programTemplates:` i18n block in `src/i18n/dictionaries.ts` (EN/TR/DE), matching how `featureCatalog` is structured.
3. Four templates to start. Keep each stage set short — 5 to 8 stages. The canonical 13-stage internship pipeline stays available as its own template so nothing is lost.
4. `validateProgramTemplate(t)` — a pure guard asserting unique keys, keys matching the API regex, ≤50 stages, exactly one first on-path stage, every `slas[].stageKey` present in `stages`, and label parity across all three locales. Call it from a test so a malformed template fails CI rather than at apply time.
5. Document the house style at the top of the file the way `newsletterContent.ts` does: what a good stage label looks like, why keys are stable and labels are not, and that editing a template never rewrites a programme that already applied it.

## ✅ Acceptance criteria
- [ ] Four templates exist, each with a distinct stage shape (not relabelled copies of the canonical set)
- [ ] Every template passes `validateProgramTemplate`, enforced by a test
- [ ] Every stage label and template name exists in EN, TR and DE
- [ ] The file imports nothing from Prisma and can be imported by a client component
- [ ] Nothing in the app's behaviour changes yet — this task only adds the catalogue
- [ ] `npm run build` passes
- [ ] a release fragment `releases/unreleased/program-template-catalogue.json` is added

## 🇹🇷 Türkçe özet
Hazır program şekillerini getiriyoruz: mezun stajı hattı, işe alıştırma (buddy) programı, liderlik/mentorluk kohortu ve kariyer geçişi programı — her biri kendi aşama kümesi, SLA'ları ve hatırlatma temposuyla. Tamamen veri ve saf fonksiyon; şema değişikliği yok, yeni tablo yok. Depoda birebir örneği zaten var: `src/lib/newsletterContent.ts`. Bu yüzden stajyer için uygun, izole bir görev — uygulama davranışı bu PR'da hiç değişmiyor, yalnızca katalog ekleniyor.

## 🤖 Agent prompt
```
Task (junior-friendly, no schema change): add a programme template catalogue.

Open first — copy the house style, do not invent one:
  src/lib/newsletterContent.ts   (curated, localised, code-level library + its header comment)
  src/lib/goalTemplates.ts
  src/lib/pipeline.ts:215-233    (ResolvedStage, defaultPipelineStages)
  src/lib/stageSla.ts:18-30
  src/app/api/admin/organizations/[id]/pipeline-stages/route.ts:33-40  (the key/length rules)
  src/i18n/dictionaries.ts       (find `featureCatalog` and mirror its structure)
  docs/newsletter.md

Do:
1. New src/lib/programTemplates.ts. CLIENT-SAFE: no Prisma import — the wizard renders it
   directly, same rule that keeps src/lib/pipeline.ts importable from client components.
2. PROGRAM_TEMPLATES with four entries: graduate_internship, onboarding_buddy,
   leadership_cohort, career_transition. 5-8 stages each, genuinely different shapes — not
   relabelled copies of the canonical 13. Keep the canonical set as its own template too.
3. Names/descriptions go in a `programTemplates:` i18n block in EN, TR and DE.
4. validateProgramTemplate(): unique keys, keys match /^[A-Za-z0-9_]+$/ and <=60 chars,
   <=50 stages, exactly one first on-path stage, every sla stageKey exists, three-locale
   label parity. Cover it with a test so a malformed template fails CI.
5. Header comment documenting the style and the rule that editing the library never rewrites
   an already-applied programme.

Constraints:
- Nothing in app behaviour changes in this PR — catalogue only. The apply/wizard work is
  #1641 and #1643.
- npm run check:i18n enforces EN/TR/DE key parity in CI.
- Never z.enum a stage key anywhere in this repo.
- Add releases/unreleased/program-template-catalogue.json.

Done when: four validated, fully localised templates exist, the file is Prisma-free, and
build + check:i18n are green.
```

**Zorluk** · S · **Tahmini** · 1 day · **Bağımlılık** · #1637 (for the `phase` field; the catalogue can land first and gain `phase` in the same PR if #1637 is already merged)

mersahin:
## Summary

A new organization opens the pipeline editor and faces an empty stage list — inventing a pipeline from nothing is the hardest part of the first hour. This ships the ready-made shapes it can start from: the canonical thirteen-stage pipeline plus a **graduate internship**, an **onboarding buddy programme**, a **leadership cohort** and a **career transition**, each with its own ordered stage set, its own per-stage service levels, its own reminder cadence and, where it matters, the documents the programme expects.

Pure data and pure functions. No schema change, no new table, and no user-visible surface yet — the picker and the launch wizard are #1643. Catalogue only, as the issue scopes it.

Two design decisions worth reviewing:

**Keys in code, labels where they belong.** Stage keys are stable identifiers (they are what `pipelineStatus` stores and what an SLA row points at), prefixed per template so two templates never collide. None of them names a canonical default key, and the canonical template derives its stages from the enum instead of retyping them — a hardcoded default key in `src/` is exactly what `check:stage-keys` (#1886) exists to catch. Stage labels ship in EN/TR/DE inside the module, the same way `src/lib/pipeline.ts` already holds its own labels; the template **names and descriptions** are UI copy and live in a new `programTemplates` dictionary block, so `check:i18n` enforces the parity.

**Applying goes through the existing writers, and it is not allowed to break a live pipeline.** This module has no writer of its own. `templateStagePayload()` produces the exact body `PUT /api/admin/organizations/[id]/pipeline-stages` already accepts and `templateSlaPayload()` the body `PUT /api/admin/stage-sla` accepts, so a template inherits those routes' authz, premium gate and validation rather than getting a second, laxer writer. `templateApplyBlockers()` — pure, client-safe, testable without a database — is the pre-flight for both, and it **refuses outright** in two cases:

- any mentee currently sits on a stage the template does not contain. The stage editor replaces the set with `deleteMany` + `createMany` and remaps no relation (#1634), so their `pipelineStatus` would point at a stage with no board column, no funnel row and no way out.
- the organization has a `StageSla` row for a stage the template drops. `StageSla.stageKey` carries **no foreign key** to `PipelineStage`, so that row *survives* the swap: `resolveStageSlas()` keeps returning a non-empty map, `stageDeadlineUpdate()` keeps reading the org as SLA-managed, and since no new stage has an SLA every later stage move writes `stageDeadline: null` over a hand-typed date. It is unrepairable afterwards too — `/api/admin/stage-sla` lists one row per *resolved* stage and its PUT refuses a stageKey that is no longer a stage, so the orphan can never be submitted for deletion. Hence a refusal *before* the swap, not a warning after it.

A third case, replacing a stage set the organization already customised, **requires an explicit confirmation**. Confirmation unlocks that one only; it never unlocks the two refusals.

Closes #1641

## Changes

- `src/lib/programTemplates.ts` — the catalogue (5 templates), `programTemplate()`, `templateStages()`, `templateStagePayload()`, `templateSlaPayload()`, `templateApplyBlockers()` / `canApplyTemplate()` / `staleSlaKeys()`, `validateProgramTemplate()` / `validateProgramTemplates()`, `programTemplateCopy()`. Prisma-free and client-safe so the wizard can render it with no server round-trip. A header comment documents the house style (5–8 stages, a stage is a waiting room someone can sit in, one way in and at least one way out, genuinely different shapes, all three languages) and the rule that editing a template never rewrites a programme that already applied it. `validateProgramTemplate()` mirrors the editor endpoint's zod schema exactly — key syntax and length, **label length (120)**, the **order ceiling (1000)**, the 50-stage limit — so a template that would take a 400 at apply time is a red build instead.
- `src/i18n/dictionaries.ts` — new `programTemplates` block in EN/TR/DE: the five names and descriptions, plus the copy the apply flow will need (the replace confirmation and the blocker explanations).
- `e2e/program-templates.unit.spec.ts` — runs `validateProgramTemplate()` over the whole catalogue (`@smoke`, so a malformed template is a red PR rather than a 400 at apply time), asserts each curated template is its own shape rather than a relabelled canonical set, checks three-locale copy, pins the payload/label resolution, covers every apply blocker including the orphaned-SLA refusal, and pins that an unknown *or prototype* key (`constructor`, `toString`, …) resolves to `null` in both lookups. Eight tests, no browser, no database.
- `e2e/program-template-apply.spec.ts` — applies a template to a **fresh** organization through the existing editor endpoint and asserts the resulting stage set, order, labels, terminal/off-path flags and row count; then applies a second template in Turkish to show the replacement is the same endpoint under the same rules.
- `releases/unreleased/program-template-catalogue.json` — `minor`, **changelog only, no `notes`**: this change has no user-visible surface, and `releases/README.md` says to omit the field in that case. The user-facing highlight ships with the picker in #1643.

## Verification

- [x] `npm run check:i18n` (4361 keys × 3 locales), `npm run check:release-fragments`, `npm run check:stage-keys` all green
- [x] The pure helpers executed directly against the compiled module: catalogue validates clean, the two new validator bounds fire, every prototype key resolves to `null` in both lookups, `stale_slas` fires on a surviving `StageSla` and the SLA payload lists every stage with `null` where the template sets no rule
- [ ] `e2e/program-template-apply.spec.ts` — written but **not executed here**: this container has no database and no app server, so it runs for the first time in CI / the scheduled full suite
- [x] Added e2e tests

Preview: https://pr2247.interncrm.com

## Contributor terms


mersahin:
Review follow-up in 3dc9c6b. I verified all four findings against the code first; **all four were real, none rejected**, so nothing was changed to satisfy a reviewer I disagree with.

**1 (major) — `templateApplyBlockers()` ignores the org's `StageSla` rows.** Confirmed end to end: `StageSla.stageKey` is a plain string with no FK (`prisma/schema.prisma`), the stage editor's transaction is `pipelineStage.deleteMany` + `createMany` and touches nothing else, `resolveStageSlas()` therefore still returns a non-empty map, and `stageDeadlineUpdate()` reads `slas.size === 0` as its only "leave this org alone" test — so one surviving row flips the org into SLA-managed and every later move writes `stageDeadline: null`. The orphan really is unreachable afterwards: `/api/admin/stage-sla` GET lists one entry per *resolved* stage, and its PUT rejects any `stageKey` not in `resolvePipelineStages()`, so the row can never be submitted for deletion — which is why this had to become a refusal *before* the swap rather than a warning after it. `ProgramApplyTarget` gains a **required** `configuredSlaKeys: string[]` (required on purpose: an optional field would have made "I did not look" and "there are none" the same, always-safe value) and a `stale_slas` blocker that confirmation does not unlock, plus `staleSlaKeys()` so the UI can name the rows. The related half of the finding — the template's own `slas` being written nowhere — is fixed with `templateSlaPayload()`, which produces the `PUT /api/admin/stage-sla` body with every stage listed and `null` where the template sets no rule (that route's own way of removing one), so no stale rule survives an apply. The docstring now states what "empty" actually promises (the *stage swap* is safe), spells out the two-call sequence #1643 owns, and says plainly that `reminderDays` and `documentRequirements` are metadata nothing consumes yet.

**2 (minor) — validator vs. endpoint schema.** Reproduced: a clone of `onboarding_buddy` with a 200-character German label and `order = 5000` returned `[]`. `MAX_LABEL_LENGTH = 120` and `MAX_ORDER = 1000` added, both covered in the malformed-template test. Length is measured on the **raw** string rather than the trimmed one, because that is what the endpoint's `z.string().max(120)` sees (it trims only when storing).

**3 (minor) — prototype keys in `programTemplateCopy()`.** Reproduced: `programTemplateCopy(dict, 'constructor')` returned `Object`. It now resolves the template first (`if (!programTemplate(key)) return null`) *and* guards with `hasOwnProperty`, so the two exported lookups cannot disagree by construction. Pinned for `constructor` / `toString` / `valueOf` / `hasOwnProperty` / `__proto__` in a new test.

**4 (minor) — release fragment.** Agreed: the fragment's own changelog line says "no user-visible surface yet" while its EN note announced the feature. `notes` dropped, the changelog prose rewritten to cover the new refusal and payload, and the file renamed to `program-template-catalogue.json` to match the issue's acceptance criteria.

Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings in unrelated admin pages), `npm run check:i18n` 4361 keys × 3 locales, `npm run check:release-fragments` exit 0, `npm run check:stage-keys` exit 0. The pure helpers were also executed directly against the compiled module — the catalogue still validates clean, both new bounds fire, and `stale_slas` fires on a surviving `APPLICATION_100` row. No schema change, so no `prisma validate`.

---

Issue #1621, opened 2026-08-30
🔁 [Infra] Task · Move the 17 setting readers onto the accessor + `check:settings-accessor` CI guard


mersahin:
## 🎯 Goal
A resolution chain nobody uses is decoration. Move every reader onto `resolveSetting()` from `src/lib/settingsResolver.ts` (the accessor built by #1619) so a per-tenant or per-program value actually takes effect at the point it matters — the AI budget, the retention window, the mentor nudge, the blind-review flag — and add a CI guard so the next feature cannot quietly re-open a direct read.

## 🔍 Current state
Every one of these reads the global row and re-implements its own parse/default:
- `src/lib/aiGate.ts:25` — `parseInt(await getSetting('aiMonthlyQuota'), 10) || 0`
- `src/lib/retention.ts:67` — `parseInt(await getSetting('retentionMonths'), 10) || 12`
- `src/lib/mentorAttention.ts:32` — `parseInt(await getSetting('reminderDays'), 10) || 14`
- `src/services/emailService.ts:1677` — the same `reminderDays` expression, duplicated
- `src/services/emailService.ts:2879` — `(await getSetting('premiumAnalytics')) !== 'true'`
- `src/lib/twoFactorPolicy.ts:9` — `(await getSetting('require2fa')) as Require2faMode`
- `src/lib/outcomeComms.server.ts:86` — `(await getSetting('outcomeAutoSend')) !== 'true'`
- `src/lib/newsletterDispatch.ts:371, 388, 404` — cadence, audience, send hour
- `src/app/api/register/route.ts:130` — `(await getSetting('selfRegistration')) === 'manual'`
- `src/app/api/company/talent-pool/route.ts:60` — `earlyAccessWindowDays`
- `src/app/api/interview-panels/route.ts:70, 122` and `src/app/api/interview-panels/[id]/route.ts:66` — `blindReview`
- `src/app/api/admin/analytics/cohorts/route.ts:20`, `benchmark/route.ts:27`, `sources/route.ts:18` — `premiumAnalytics`
- `src/app/api/admin/settings/route.ts:13, 51` — `getSettings()` for the admin screen

There is no guard: `scripts/` already holds `check:i18n`, `check:release-fragments`, `check:k6` and `check:authz` style gates wired into `ci.yml`, but nothing stops a new route from calling `prisma.setting.findFirst()` directly.

## 🛠 What to build
[...]

Issue #1619, opened 2026-08-30
🔑 [Infra] Task · programId scope + typed `resolveSetting()` with a per-request cache


mersahin:
## 🎯 Goal
Give the settings table its third scope and give the codebase one function that answers "what is this setting, here, for this program". Today the answer is a string from a global row and the caller guesses the type.

## 🔍 Current state
- `prisma/schema.prisma:1397-1401`:
  ```prisma
  model Setting {
    key       String   @id
    value     String   @db.Text
    updatedAt DateTime @updatedAt
  }
  ```
  #1553 turns this into `orgId String?` + a surrogate `id` + `@@unique([orgId, key])`. **This task starts from that shape.**
- `src/lib/settings.ts:4-62` — 18 keys, all `string`, defaults expressed as string literals (`reminderDays: '14'`, `premiumAnalytics: 'false'`, `newsletterSendHour: '9'`).
- `src/lib/settings.ts:66-77` — the two readers, neither taking a scope.
- `src/lib/pipelineStages.ts:17-37` — `resolvePipelineStages(orgId, locale)`: one indexed query, falls back to the built-in defaults for a null org. This is the shape to copy.
- `src/lib/orgContext.ts:34-37` — `currentOrgId()` returns `string | null | undefined` from the AsyncLocalStorage; `undefined` means "no context, do not scope".
- `prisma/schema.prisma:1659-1668` — `Cohort` has `orgId String?`; it is the Program the third level hangs off until the Cohort→Program promotion renames it.

## 🛠 What to build
1. **Schema (additive, safe):** add `programId String?` to `Setting`, a `program`/`cohort` relation with `onDelete: Cascade`, and replace the `@@unique([orgId, key])` from #1553 with `@@unique([orgId, programId, key])` plus `@@index([orgId, key])`. Adding a nullable column and widening a unique index is not destructive under `db push`; **do not** touch the primary key here — #1553 owns that and carries the `migration:dangerous` label (#1529).
2. **`src/lib/settings.ts`: one declarative catalogue.** Replace the string-valued `SETTING_DEFAULTS` with `SETTING_SCHEMA`, one entry per key: `{ type: 'number' | 'boolean' | 'enum' | 'string' | 'email', default, values?, scopes: ('global'|'org'|'program')[] }`. Keep `SETTING_DEFAULTS` exported as a derived string map so #1553's code and the existing API keep compiling.
3. **`src/lib/settingsResolver.ts` (server-only):**
[...]

Issue #1529, opened 2026-08-30
✋ [Infra] Task · `migration:dangerous` label + recorded acknowledgement before a destructive migration merges


mersahin:
## 🎯 Goal
The current escape hatch for a data-destroying schema change is an environment variable typed on a server at deploy time (`ALLOW_DESTRUCTIVE=1`). Nobody reviews it, nothing records who decided, and the decision is made after the code has already merged. Replace it with a decision made in the PR, by a named person, that leaves a permanent trace on the issue tracker.

## 🔍 Current state
- `infra/schema-guard.sh:263-269` — `ALLOW_DESTRUCTIVE=1` proceeds, and only refuses when `BACKUP_TAKEN != 1`. The operator override is documented in the deploy script header at `infra/deploy-prod.sh:48-51`.
- The stop message tells the operator to re-run with the flag (`infra/schema-guard.sh:272-274`) — i.e. the fix path is "try again with the safety off", at whatever hour the deploy failed.
- `.github/PULL_REQUEST_TEMPLATE.md` exists and already carries a confirm-by-checkbox mechanism (contributor IP terms, `CONTRIBUTING.md` § Contributor terms (IP)).
- `.github/workflows/project-status.yml:23-45` already reacts to issue/PR events, so label automation has a precedent in this repo.
- Standing maintainer instruction (`CLAUDE.md`): agents open a PR and merge it themselves once CI is green. A destructive migration is precisely the case that must **not** auto-merge on green.

## 🛠 What to build
1. In the migration-review workflow, apply the `migration:dangerous` label when `npm run check:migrations` returns `destructive`, and remove it when it does not.
2. Add a section to `.github/PULL_REQUEST_TEMPLATE.md` that appears only when relevant:
   - [ ] This PR contains a destructive migration. I have read the SQL in the review comment, and I confirm: the data being dropped is either backfilled first (expand/contract) or genuinely disposable.
   - a free-text line naming the rollback plan and the expected row count affected.
3. A required check that fails while the `migration:dangerous` label is present and the acknowledgement checkbox is unticked — so a green CI cannot auto-merge a destructive migration.
4. Keep `ALLOW_DESTRUCTIVE` on the server as the break-glass path, but change its stop message (`infra/schema-guard.sh:272-274`) to point at the PR acknowledgement as the normal route.
5. Document the whole flow in `docs/migrations.md` (created by the conventions task) and link it from `CONTRIBUTING.md`.

## ✅ Acceptance criteria
[...]

Issue #1515, opened 2026-08-30
🧱 Epic · Reviewable schema migrations — retire `db push --accept-data-loss` from the deploy path


mersahin:
## 🎯 Goal
Every merge to `main` currently reshapes the production MySQL with `prisma db push --accept-data-loss`, and the only thing that inspects the SQL is a bash guard that runs **on the server, mid-deploy, after the PR is merged**. That is the wrong end of the process: an enterprise change-management questionnaire asks for a reviewed, versioned artefact showing what ran against production, and a rejected deploy at 02:00 is a far worse place to discover a destructive diff than code review. This epic introduces a real migration workflow, makes the SQL visible in the PR, dry-runs it against a production-shaped snapshot, writes down the rollback and backfill conventions — and thereby **unblocks every schema-changing epic in Initiative #1514**.

## 🔍 Current state
- Deploy runs the push directly: `infra/deploy-prod.sh:503-504` (`run_tool npx prisma db push --accept-data-loss`), described in the script header at `infra/deploy-prod.sh:16-18`.
- The same shape in every other environment: `infra/server/topic-deploy.sh:232`, `infra/server/demo-refresh.sh:97`, the legacy `.github/workflows/deploy.yml:173` and `:284`.
- The destructive gate exists and is good, but is server-side: `infra/schema-guard.sh:143-152` (`prisma migrate diff --script`), pattern at `infra/schema-guard.sh:175`, invoked at `infra/deploy-prod.sh:479-484`. Its unit test runs in CI (`.github/workflows/ci.yml:106`) — the *guard logic* is tested in the PR, the *actual diff* never is.
- There is **no `prisma/migrations/` directory**; `prisma/` holds only `schema.prisma` (2 664 lines, 88 models) plus seeds and nine ad-hoc backfill scripts.
- Two hand-rolled pre-push phases already exist because `db push` alone cannot express them: `prisma/push-company-interest-expand.mjs:65` (expand/contract by hand) and `prisma/backfill-json-columns.mjs --repair` run **twice**, before and after the push (`infra/deploy-prod.sh:501`, `:518`).
- Compensating controls that this epic builds on rather than replaces: pre-deploy backup (`infra/backup-db.sh:1-25`), daily freshness check + monthly restore drill (`.github/workflows/backup-verify.yml:23`, `:87`, `:110`), sanitiser (`scripts/sanitize-db.mjs:1-20`), runbook `docs/disaster-recovery.md`.
[...]

PR #2249, opened 2026-09-07
feat(multi-tenancy): give Setting an orgId with a global fallback

### LINKED ISSUE #1553: 🔑 [Infra] Task · Setting gets an orgId: composite key + per-tenant resolution with a global fallback
## 🎯 Goal
Eighteen product decisions — whether 2FA is mandatory, how long candidate data is kept, how many AI calls a month are allowed, whether sign-up is open, whether interview review is blind — live in a single global key/value table. With two tenants on the box, one customer's compliance policy silently becomes everybody's. Give `Setting` a tenant key and resolve it per org, falling back to the global row so a fresh tenant inherits sane defaults instead of an empty config.

## 🔍 Current state
- `prisma/schema.prisma:1397-1401`:
  ```prisma
  model Setting {
    key       String   @id
    value     String   @db.Text
    updatedAt DateTime @updatedAt
  }
  ```
- `src/lib/settings.ts:4-62` — `SETTING_DEFAULTS` with 18 keys (`reminderDays`, `supportEmail`, `weeklyDigest`, `retentionMonths`, `require2fa`, `earlyAccessWindowDays`, `premiumAnalytics`, `selfRegistration`, `aiMonthlyQuota`, `outcomeAutoSend`, `blindReview`, `newsletterSchedule`, `newsletterAudience`, `newsletterSendHour`, …).
- `src/lib/settings.ts:68-77` — `getSettings()` does a bare `prisma.setting.findMany()`; `getSetting(key)` does `findUnique({ where: { key } })`.
- `src/app/api/admin/settings/route.ts:9-13` (GET) and `:36-52` (PUT) — admin-only, not wrapped in `withTenantScope`; PUT upserts `where: { key }`.
- `src/lib/aiGate.ts:25` reads `aiMonthlyQuota` through `getSetting`; `src/services/emailService.ts:13` imports `getSetting` for the cron cadence.

## 🛠 What to build
1. **⚠️ DANGEROUS SCHEMA CHANGE.** Re-key `Setting` to `orgId String?` + `@@id([orgId, key])` (or, if a nullable component in a composite `@@id` proves awkward on MySQL, keep `id String @id @default(cuid())` and add `@@unique([orgId, key])` — prefer this second shape). `db push --accept-data-loss` resolves a primary-key change on a `String @id` table as **DROP + CREATE**: every stored setting is lost, and a per-PR topic env pushes to its own empty database so the diff looks harmless on the PR. This task **must not merge before the migrations epic #1515** has landed `prisma migrate deploy`, and its PR must carry the `migration:dangerous` label from **#1529**.
2. `src/lib/settings.ts`: `getSettings(orgId)` / `getSetting(key, orgId)` resolving **tenant row → global row (`orgId: null`) → `SETTING_DEFAULTS`**. Keep the existing zero-argument signatures working (resolving to the current org via `currentOrgId()` from `src/lib/orgContext.ts`, else global) so the ~20 existing call sites (enumerated in #1621) do not all have to change in one PR.
3. `src/app/api/admin/settings/route.ts`: wrap GET and PUT in `withTenantScope(session, …)`; PUT upserts against `{ orgId, key }` so a tenant admin can only ever write their own org's row.
4. Add `Setting` to `TENANT_MODELS` in `src/lib/orgContext.ts:43`.
5. Backfill: keep the existing rows as the global (`orgId: null`) layer — do **not** stamp them with the `default` org, or every future tenant starts from nothing.
6. Document the resolution order at the top of `src/lib/settings.ts` and in `docs/tenant-isolation.md`.

## ✅ Acceptance criteria
- [ ] Two orgs can hold different `require2fa` / `aiMonthlyQuota` / `retentionMonths` values and each resolves its own
- [ ] An org with no row for a key falls back to the global row, then to `SETTING_DEFAULTS`
- [ ] `PUT /api/admin/settings` as tenant B's admin cannot modify tenant A's row (server-side, verified by request, not by UI)
- [ ] `Setting` appears in `TENANT_MODELS`
- [ ] The PR body states the change is DANGEROUS under `db push` and names the migration that performs it
- [ ] `npm run build` passes
- [ ] `npx prisma validate` passes
- [ ] an e2e spec covers per-tenant resolution and the fallback chain
- [ ] a release fragment `releases/unreleased/setting-org-scope.json` is added

## 🇹🇷 Türkçe özet
Bugün 18 ayarın tamamı tek bir global satırda: bir müşterinin 2FA zorunluluğu, saklama süresi ve AI kotası otomatik olarak herkesin oluyor. `Setting`'e `orgId` ekliyoruz ve çözümleme sırasını "kiracının satırı → global satır → koddaki varsayılan" yapıyoruz; böylece yeni bir kiracı boş bir konfigürasyonla değil makul varsayılanlarla başlıyor. **Dikkat:** bu tablonun birincil anahtarı değişiyor ve `db push --accept-data-loss` bunu DROP+CREATE olarak çözer — yani canlı ayarlar silinir. Bu yüzden #1515 (migration'lar) inmeden bu iş merge edilmemeli. Üçüncü kapsam (`programId`), tipli `SETTING_SCHEMA` ve yeni `resolveSetting()` erişim fonksiyonu bu işin kapsamında **değil** — onlar #1619'a ait (aşağıdaki kapsam sınırına bakın).

## 🔗 Scope boundary

#1619 and this issue both re-shape the `Setting` model and both build settings resolution, from opposite ends. **This issue (#1553) owns the re-key**: the `orgId` column, the primary-key change (prefer the surrogate `id String @id @default(cuid())` + `@@unique([orgId, key])` shape), the `TENANT_MODELS` entry, the tenant-scoped admin write path, and the org → global → `SETTING_DEFAULTS` fallback added **in place** inside the existing `src/lib/settings.ts` readers. It is the P0 isolation blocker, carries the `migration:dangerous` label (#1529) and blocks on #1515.

**#1619 owns the third scope and the accessor**: it adds a nullable `programId` to this same table, widens the unique index to `@@unique([orgId, programId, key])`, and introduces the one typed accessor `src/lib/settingsResolver.ts` (`resolveSetting()`, `SETTING_SCHEMA`, per-request cache) on top of the chain this issue establishes. #1621 then moves the ~20 readers onto **#1619's** `resolveSetting()`.

Do not create a second settings table, a second `orgId` column, or a resolver module here — re-keying `Setting` twice under `db push` is the worst outcome this backlog can produce. Land this issue first; #1619 must stay purely additive on top of it.

## 🤖 Agent prompt
```
Objective: give Setting a tenant key and resolve settings per org with a global
fallback.

Open first:
  prisma/schema.prisma:1397   model Setting
  src/lib/settings.ts         SETTING_DEFAULTS + getSettings/getSetting
  src/app/api/admin/settings/route.ts
  src/lib/orgContext.ts:43    TENANT_MODELS
  src/lib/aiGate.ts:25, src/services/emailService.ts:13   (callers)

Constraints that WILL trip you up:
- DANGEROUS: Setting is `key String @id`. Changing the primary key under
  `db push --accept-data-loss` is DROP+CREATE — live settings are destroyed.
  Every PR topic env has its own EMPTY database, so the diff looks green there
  and only bites production. Do not merge before epic #1515 lands
  `prisma migrate deploy`; label the PR migration:dangerous (#1529).
  Prefer the non-destructive shape: keep a surrogate `id String @id
  @default(cuid())` and add `@@unique([orgId, key])`.
- SCOPE BOUNDARY: this task owns the re-key and the ORG level only. Do NOT add
  a programId column, do NOT create src/lib/settingsResolver.ts, do NOT
  introduce resolveSetting() or a typed SETTING_SCHEMA — #1619 adds all of
  that additively on top of your shape, and #1621 migrates the callers.
- Backfill: existing rows stay orgId = NULL (the GLOBAL layer). Do NOT stamp
  them with the `default` org or new tenants inherit nothing.
- Resolution order is tenant row -> global row -> SETTING_DEFAULTS. Keep the
  zero-arg getSetting()/getSettings() signatures working via currentOrgId()
  from src/lib/orgContext.ts; there are ~20 call sites (listed in #1621) and
  none of them change in this PR.
- Authorization is server-side: PUT must scope the upsert by orgId, not hide a
  field in the UI.
- Add 'Setting' to TENANT_MODELS in src/lib/orgContext.ts.
- No new user-facing copy is needed; if you add any, EN/TR/DE keys in
  src/i18n/dictionaries.ts or `npm run check:i18n` fails CI.

Done when: two orgs resolve different values for require2fa and aiMonthlyQuota,
an org with no row falls back to the global row, tenant B's admin cannot write
tenant A's setting, `npx prisma validate` and `npm run build` pass, and an e2e
spec proves the fallback chain.
```

**Zorluk** · L · **Tahmini** · 2 days · **Bağımlılık** · #1515, #1529

mersahin:
## Summary

Eighteen product decisions — 2FA enforcement, retention window, AI quota, sign-up policy, blind review, newsletter cadence — lived in a single global key/value table. With two tenants on the box, one customer's compliance policy silently becomes everybody's.

`Setting` is now keyed by **(`orgId`, `key`)** and resolves **tenant row → global row (`orgId = NULL`) → `SETTING_DEFAULTS`**, so a fresh tenant inherits sane platform defaults instead of an empty config.

Closes #1553

## ⚠️ DANGEROUS under `db push` — needs #1515 before it reaches a real database

This changes `Setting`'s **primary key**. `prisma db push --accept-data-loss` resolves a primary-key change on a `String @id` table as **DROP + CREATE**: every stored setting is destroyed. Every PR topic environment has its **own empty database**, so this PR's own preview looks perfectly healthy and only an environment that holds real rows is hurt.

- **Do not merge/deploy before the migrations epic #1515 lands `prisma migrate deploy`.** There is no `migrations/` folder in the repo today, so there is no migration file to name — the change is applied by whatever first migration #1515 generates from this schema.
- This PR should carry the `migration:dangerous` label (#1529); I could not apply it from here.
- The non-destructive shape was chosen where there was a choice: a surrogate `id String @id @default(cuid())` + `@@unique([orgId, key])`, **not** a composite `@@id([orgId, key])` (a nullable column cannot be part of a primary key). The equivalent hand-run SQL, if the maintainer wants to pre-apply it before a `db push`, is `ALTER TABLE Setting ADD COLUMN orgId VARCHAR(191) NULL, ADD COLUMN id VARCHAR(191) NOT NULL DEFAULT (uuid()), DROP PRIMARY KEY, ADD PRIMARY KEY (id), ADD UNIQUE KEY Setting_orgId_key_key (orgId, \`key\`)` — after which the push is a no-op. I did not commit that as a migration file, per the repo's "no SQL migrations" rule.

## Changes

- **`prisma/schema.prisma`** — `Setting` gains `id String @id @default(cuid())`, `orgId String?`, an `Organization` relation (`onDelete: Cascade`) and `@@unique([orgId, key])`. `key` is no longer the primary key.
- **`src/lib/settings.ts`** — the one place the resolution rule lives:
  - `getSetting(key, orgId?)`, `getSettings(orgId?)`, new `setSetting(key, value, orgId?)`.
  - The org argument is optional. Omitted, it resolves to the org bound by `withTenantScope()` (`currentOrgId()`), and to the **global layer** when no org is bound — so the ~20 existing zero-argument call sites (#1621) are untouched, and a single-tenant installation (and every request while `MT_ENFORCE_ISOLATION` is off) reads and writes exactly the rows it does today.
  - `setSetting` is a read-modify-write, not `upsert`: a compound unique cannot address a `NULL` component.
- **`src/app/api/admin/settings/route.ts`** — GET and PUT wrapped in `withTenantScope(session, …)`; PUT writes through `setSetting`, so the layer comes from the session's bound org and **never from the request body**. Tenant B's admin cannot reach tenant A's row regardless of what it posts. Writes are sequential so two values for the same `(org, key)` cannot race into a duplicate row.
- **`src/lib/orgContext.ts`** — `Setting` added to `TENANT_MODELS`, with the reasoning in place.
- **`prisma/backfill-cron-baseline.mjs`** — reads/writes its `cronBaselineAt` row in the global layer explicitly.
- **`scripts/check-tenant-models.mjs`** — the `EXEMPT` prose said `Setting` was deliberately left un-exempted so the register-vs-exempt call would be made against the real column. That call is now made: registered, not exempt. Comment updated to record it.
- **`docs/tenant-isolation.md`** — new section "Settings: per-tenant with a global fallback", covering the layer table, the resolution order, the opt-out below, and the MySQL NULL-uniqueness wrinkle.
- **`e2e/setting-org-scope.spec.ts`** (new) + `e2e/helpers/db.ts` `setGlobalSetting()`; the three existing specs that wrote `Setting` directly moved onto the new shape.
- Release fragment `releases/unreleased/setting-org-scope.json` (`minor`, EN/TR/DE notes).

## The auto-filter question (asked explicitly in the issue)

**Yes, an org-scoped auto-filter would have broken the deliberate global-fallback read, and it is handled — not avoided.**

`Setting` **is** registered in `TENANT_MODELS`, so any code that reaches for `prisma.setting` outside `src/lib/settings.ts` is auto-scoped to its own tenant like every other tenant model. But with `MT_ENFORCE_ISOLATION=true` that same middleware rewrites *every* query to `where: { orgId: <tenant> }` — which is precisely the filter that hides the `orgId = NULL` row and would turn step 2 of the chain into a silent "code default" for every tenant.

So the queries inside `settings.ts` run through `runWithOrg(null, …)`, which binds an empty tenant context that the middleware skips (and is a plain passthrough when the flag is off). That is safe because the module computes the org **itself**, from the bound context, and never from input — the escape hatch buys back the global row, not the ability to address someone else's tenant. The second test in `e2e/setting-org-scope.spec.ts` pins exactly this: with the flag on and an org bound, a key the org never set still resolves to the global row.

## Backfill / migration path

Existing rows stay `orgId = NULL` — they *are* the global layer. `prisma/backfill-organization.mjs` already excludes `Setting` by name with that reason written down, so no backfill change was needed. Stamping the legacy rows with the `default` org would turn platform-wide defaults into one tenant's private settings and leave every other tenant with no configuration at all.

## Verification

- [x] `npx prisma format && npx prisma validate && npx prisma generate` — valid
- [x] `npm run lint` — clean (only the two pre-existing `react-hooks/exhaustive-deps` warnings)
- [x] `npm run check:tenant-models` — passes; `Setting` now counted among the registered models
- [x] `npm run check:release-fragments`, `npm run check:openapi`, `npm run check:auth-reads`, `npm run check:query-scalars` — all pass
- [x] `npx playwright test --list e2e/setting-org-scope.spec.ts` — 2 tests collected
- [ ] E2E suite not run locally (no database or browser build in this environment) — CI runs the smoke set on this PR

## Deliberately out of scope

- No `programId` column, no `src/lib/settingsResolver.ts`, no `resolveSetting()` and no typed `SETTING_SCHEMA` — those belong to #1619, which stays purely additive on top of this shape, and #1621 then moves the readers.
- The ~20 existing `getSetting()` call sites are unchanged by design; they keep working through the ambient-org resolution.
- The new spec is **not** tagged `@smoke` (it needs two orgs and DB round-trips); it runs in the scheduled full suite.
- No new user-facing copy, so no `src/i18n/dictionaries.ts` change.

## Contributor terms

- [x] I accept the [contributor terms](https://github.com/21072026/internship/blob/main/CONTRIBUTING.md#contributor-terms-ip): my contribution is licensed under AGPL-3.0-or-later, the exclusive right to use it passes to the rights holder (Mehmet Erşahin) who may also license it commercially, it is my own work, and I make no claim over the application.

https://claude.ai/code/session_01Ra7wrp478VPjzHJHh4QMvZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ra7wrp478VPjzHJHh4QMvZ)_

### DIFF
--- a/docs/tenant-isolation.md
+++ b/docs/tenant-isolation.md
@@ -144,6 +144,19 @@ module computes the org itself, from the bound context, and never from request
 input: `PUT /api/admin/settings` passes no org at all, so a tenant admin can only
 ever write their own row.
 
+It reaches both of those through `src/lib/tenantAmbient.ts` rather than importing
+`orgContext.ts` directly, and that indirection is load-bearing rather than
+stylistic. `orgContext.ts` imports `node:async_hooks`, and `settings.ts` sits in a
+**client** module graph — a client component imports a constant from
+`documentAccess.ts`, which reaches `settings.ts` via `retention.ts` — so a direct
+import fails the production build outright with `Reading from "node:async_hooks"
+is not handled by plugins`. `tenantAmbient.ts` imports nothing, holds two function
+slots, and `orgContext.ts` fills them as a side effect of being loaded; every
+request that binds a tenant has loaded it, because `withTenantScope` lives there.
+If it was never loaded, the fallbacks (no bound org, call `fn` directly) are the
+right answers anyway — there is no middleware installed to escape from. The same
+seam is the way in for any other client-reachable module that needs the bound org.
+
 Uniqueness has a MySQL wrinkle worth knowing: the pair is enforced by
 `@@unique([orgId, key])`, and MySQL treats `NULL`s as distinct in a unique index.
 The constraint therefore binds the per-tenant rows only; the global layer stays
--- a/docs/tenant-isolation.md
+++ b/docs/tenant-isolation.md
@@ -84,12 +84,12 @@ Two escape hatches live inside the script, and both require a written reason.
 `EXEMPT` is for a model that is deliberately never auto-scoped. Today it holds
 one name: `Organization`, which *is* the tenant rather than a row inside one and
 can never grow an `orgId` of its own. An exemption is only granted against a
-schema someone has read, so `Setting` is **not** pre-exempted even though its
-column is coming: when #1551 adds `orgId` to `Setting` this check will fail, and
-the register-vs-exempt call gets made then, against the real shape. (#1557
-records the intent — `Setting`'s legacy rows will stay `orgId = NULL` as the
-global fallback layer — but #1560 also says `Setting` must be *registered* and
-behave specially, and those two have to be reconciled with the column in hand.)
+schema someone has read. `Setting` is the worked example: it was deliberately
+left un-exempted while its column was still only planned, the check duly failed
+the day #1553 added `orgId`, and the call was then made against the real shape —
+**registered, not exempt**, with its own readers opting out (see
+[Settings: per-tenant with a global fallback](#settings-per-tenant-with-a-global-fallback-1553)
+below).
 
 `PENDING_REGISTRATION` is for a model that is known to be unprotected and is
 waiting on its own reviewed change — the eight models of #1559. It is a ratchet,
@@ -108,6 +108,53 @@ not an allowlist:
 
 So the guard's job while #1559 is open is to stop the set of unprotected models
 from *growing*, and to say on every run exactly which ones they are.
+
+### Settings: per-tenant with a global fallback (#1553)
+
+`Setting` is the one tenant model that must be able to read *outside* its tenant,
+and it is worth understanding before touching it.
+
+Eighteen product decisions — `require2fa`, `retentionMonths`, `aiMonthlyQuota`,
+`selfRegistration`, `blindReview`, the newsletter cadence — live in this one
+key/value table. It is now keyed by **(`orgId`, `key`)**:
+
+| `orgId` | meaning |
+|---------|---------|
+| a tenant id | that tenant's override |
+| `NULL` | the **global** layer: the platform-wide value every tenant inherits until it sets its own, and the only layer a single-tenant installation ever writes |
+
+Resolution is **org row → global row → `SETTING_DEFAULTS`**, and that rule lives
+in exactly one file, `src/lib/settings.ts`. Nothing else reads `prisma.setting`
+directly — `getSetting(key, orgId?)` / `getSettings(orgId?)` / `setSetting(key,
+value, orgId?)` are the whole surface. The org argument is optional: omitted, it
+resolves to the org bound by `withTenantScope()` (`currentOrgId()`), and to the
+global layer when no org is bound. That is what keeps the ~20 existing
+zero-argument call sites correct without changing any of them, and what makes a
+single-tenant deployment behave exactly as it did before.
+
+**The auto-filter had to be opted out of, on purpose.** `Setting` *is* registered
+in `TENANT_MODELS`, so any code that reaches for `prisma.setting` outside this
+module is scoped to its own tenant like everything else. But with enforcement on,
+that same middleware would rewrite the readers' query to `where: { orgId: <tenant> }`
+— which is precisely the filter that hides the `orgId = NULL` row and would turn
+step 2 of the chain into a silent "code default" for every tenant. So the queries
+in `settings.ts` run inside `runWithOrg(null, …)`, which clears the tenant context
+for the duration and lets the module see both layers. This is safe because the
+module computes the org itself, from the bound context, and never from request
+input: `PUT /api/admin/settings` passes no org at all, so a tenant admin can only
+ever write their own row.
+
+Uniqueness has a MySQL wrinkle worth knowing: the pair is enforced by
+`@@unique([orgId, key])`, and MySQL treats `NULL`s as distinct in a unique index.
+The constraint therefore binds the per-tenant rows only; the global layer stays
+single because `setSetting()` is its only writer and does a read-modify-write
+rather than a blind insert. (`upsert` is not an option either way — a compound
+unique cannot address a `NULL` component.)
+
+Existing rows keep `orgId = NULL` and go on working as the global layer;
+`prisma/backfill-organization.mjs` excludes `Setting` by name for exactly this
+reason. Stamping them with the `default` org would turn platform-wide defaults
+into one tenant's private settings and leave every other tenant with nothing.
 
 ### Per-route rollout status
 
--- a/e2e/admin-settings.spec.ts
+++ b/e2e/admin-settings.spec.ts
@@ -39,7 +39,7 @@ test('admin saves system settings and bulk-imports mentees from CSV', async ({ p
     expect(a?.role).toBe('MENTEE');
   } finally {
     await prisma.user.deleteMany({ where: { email: { in: [importedA, importedB] } } });
-    await prisma.setting.deleteMany({ where: { key: { in: ['reminderDays', 'supportEmail', 'weeklyDigest'] } } });
+    await prisma.setting.deleteMany({ where: { orgId: null, key: { in: ['reminderDays', 'supportEmail', 'weeklyDigest'] } } });
     await cleanupByEmail(adminEmail);
   }
 });
--- a/e2e/helpers/db.ts
+++ b/e2e/helpers/db.ts
@@ -24,6 +24,16 @@ export const prisma = new PrismaClient();
  * `page.waitForURL` timeout in `signInAsFreshUser`, pointing at the wait rather
  * than at the address (#2043).
  */
+// Write a Setting row in the GLOBAL layer (orgId = NULL) — the layer a
+// single-tenant test server reads, and the fallback every tenant inherits
+// (#1553). Deliberately a read-modify-write: the natural key (orgId, key)
+// contains a nullable column, so `upsert` cannot address the NULL row.
+export async function setGlobalSetting(key: string, value: string) {
+  const existing = await prisma.setting.findFirst({ where: { orgId: null, key } });
+  if (existing) await prisma.setting.update({ where: { id: existing.id }, data: { value } });
+  else await prisma.setting.create({ data: { orgId: null, key, value } });
+}
+
 export function uniqueEmail(prefix: string) {
   const slug = prefix.replace(/[^A-Za-z0-9._-]+/g, '-').replace(/^[-.]+|[-.]+$/g, '') || 'e2e';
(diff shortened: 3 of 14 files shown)


mersahin:
Note on the `migration:dangerous` label: it does not exist in this repository yet — #1529 has not landed, so there is no label to apply. The danger is spelled out in the PR body instead. Please do not deploy this to an environment that holds real settings before #1515 provides `prisma migrate deploy`; a `db push --accept-data-loss` will resolve the primary-key change as DROP + CREATE and take the stored settings with it.

---

mersahin:
All 13 checks are green on `bc143ea` — Lint · Typecheck · Build, Playwright smoke, CodeQL, Demo seed fidelity, both image builds, and the topic environment deployed: https://pr2249.interncrm.com

**Deliberately NOT merged, and not set to auto-merge.** The repo's standing instruction is to merge a green PR, but this one is the documented exception: merging lands on `main`, `main` deploys to production, and `infra/deploy-prod.sh` still runs `prisma db push --accept-data-loss`. That resolves this primary-key change as DROP + CREATE and takes the live settings with it. #1515 (which introduces `prisma migrate deploy`) is still open at 0/4 sub-issues and there is no `prisma/migrations/` directory yet — #1515 names *this exact change* as one of the six that will "eat live data" under the current deploy path.

Green CI here is not evidence to the contrary: the topic environment has its own **fresh, empty** database, which is precisely the blind spot #1515's story 3 exists to close.

A note on the first red run, since it is visible in the history: `src/lib/settings.ts` turns out to sit in a **client** module graph (a client component imports a constant from `documentAccess.ts`, which reaches it via `retention.ts`), so importing `orgContext.ts` from it failed the build on `node:async_hooks`. `bc143ea` adds `src/lib/tenantAmbient.ts` — an import-free seam that `orgContext.ts` fills when it loads — instead of pulling the server-only module into a client bundle.

---

Issue #1585, opened 2026-08-30
🌱 [Admin] Task · Audit log CSV export and a nightly AuditLog retention sweep


mersahin:
## 🎯 Goal
Two small finishing moves on the audit trail: let a reviewer take the filtered rows away as a CSV (every security questionnaire asks for "export"), and stop the table growing forever by pruning rows past a configured retention window. Both are self-contained and follow patterns that already exist in the repo — a good first issue.

## 🔍 Current state
- `prisma/schema.prisma:2477-2489` — `AuditLog`; nothing ever deletes from it. The viewer lands in the sibling task.
- Export precedent: the candidate list already exports CSV/XLSX **over the whole filtered set, not just the visible page** (`src/app/admin/candidates/page.tsx:292-298`), using `src/lib/excel.ts`. Copy that behaviour — an export that only covers page 1 is worse than none.
- Retention precedent: `src/lib/settings.ts:11` holds `retentionMonths: '12'` (GDPR candidate review), and `src/lib/trustedDevice.ts:340-365` is a nightly purge registered in `src/services/emailService.ts:3194-3197`. That is the shape of an age-based sweep — but **the schedule for it is #1678's**, not a new `cron.schedule` here (see the scope boundary below).
- Cron registration lives in `src/services/emailService.ts:3051-3200`, started from `/api/cron/start` (`src/app/api/cron/start/route.ts:10-31`).
- `src/lib/download.ts` exists for the client-side download helper.

## 🛠 What to build
1. `GET /api/admin/audit/export` behind `requireCapability('audit.export')`: takes the **same** query params as the viewer, streams a CSV of every matching row (not just the current page), columns `createdAt, action, actorEmail, actorId, targetId, detail`. Cap at a sane row limit and say so in the response header if truncated.
2. An export button on `/admin/audit`, EN/TR/DE labels.
3. A new setting `auditRetentionMonths` (default `'24'`) in `SETTING_DEFAULTS` (`src/lib/settings.ts:4-62`) with its comment explaining the default, plus the field in the settings page (`src/app/admin/settings/page.tsx:171-250`).
[...]

PR #2255, opened 2026-09-07
feat: who accessed my account — impersonation history on the account page (#1587)

### LINKED ISSUE #1587: 🌱 [Trust] Task · &#34;Who accessed my account&#34; — impersonation history on the account page
## 🎯 Goal
An impersonated user gets one in-app notification at the moment it happens and can never look it up again. Give every user a permanent, self-service list of who entered their account, when, for how long and why — next to the trusted-devices list they already have. It is the single cheapest trust feature in the product: the rows are already written, only the read is missing.

## 🔍 Current state
- `src/app/api/admin/impersonate/route.ts:43-45` writes `AuditLog { actorId: <admin>, action: 'IMPERSONATE_START', targetId: <user> }`; `:57-61` sends the one-off notification (`impersonation.accessed` / `impersonation.accessedWithReason`). The reason itself is stored on the `ActivityLog` row's `detail` (`:53`), **not** on the audit row — that is a small gap this task should close by also putting the reason in `AuditLog.detail`.
- `src/app/api/impersonate/stop/route.ts:19-21` writes the matching `IMPERSONATE_STOP` row, so a start/stop pair gives a duration.
- `prisma/schema.prisma:2477-2489` — `AuditLog` is indexed on `actorId` only. This query filters on `targetId`, so it needs `@@index([targetId])`.
- `src/components/AccountSettings.tsx:428-443`, `:691-700` — the existing self-service trusted-device list (device, IP, last used, revoke). The new section sits beside it and looks the same.
- `src/app/api/account/devices/route.ts:13-19` — the shape of a small self-service account endpoint to copy.
- `src/lib/auth.ts:388-389` — the 30-minute cap, which bounds any session with no explicit stop row.

## 🛠 What to build
1. Add `@@index([targetId])` to `AuditLog` in `prisma/schema.prisma` (additive).
2. Also record the impersonation reason on the audit row: pass `detail: reason` in `src/app/api/admin/impersonate/route.ts:43-45` (it already has `reason` in scope at `:41`).
3. `GET /api/account/impersonations` — the **signed-in user's own** history only. Never accept a `userId` parameter; read `session.user.id`. Pair each `IMPERSONATE_START` with the next `IMPERSONATE_STOP` for the same actor/target to derive a duration; when there is no stop row, show "ended automatically (30 min cap)". Return the admin's display name, the timestamp, the duration and the reason — never the admin's email or any other admin field beyond the name.
4. A read-only "Account access history" card in `src/components/AccountSettings.tsx`, empty-state included, EN/TR/DE strings in `src/i18n/dictionaries.ts`.
5. It is available to **every** role — mentors and mentees use this product for free and this is a transparency control, not a premium feature.

## ✅ Acceptance criteria
- [ ] Any signed-in user sees their own impersonation history on the account page, with an empty state when there is none
- [ ] The endpoint reads `session.user.id` and cannot be pointed at another user (no `userId` param, no IDOR)
- [ ] A start with no matching stop renders as auto-expired rather than "still active"
- [ ] The reason is shown when one was given, and is written to `AuditLog.detail` going forward
- [ ] `AuditLog` has `@@index([targetId])`; `npx prisma validate` passes with no DROP in the diff
- [ ] `npm run check:i18n` passes (EN/TR/DE)
- [ ] `npm run build` passes
- [ ] an e2e spec impersonates a mentee and asserts the row appears on that mentee's account page
- [ ] a release fragment with EN/TR/DE `notes` is added

## 🇹🇷 Türkçe özet
Hesabına girilen kullanıcı bugün sadece o anda bir bildirim alıyor, sonra bir daha asla göremiyor. Her kullanıcıya kalıcı ve kendi kendine erişilebilir bir liste veriyoruz: hesabıma kim girdi, ne zaman, ne kadar süreyle, hangi gerekçeyle — zaten var olan güvenilen cihazlar listesinin hemen yanında. Satırlar zaten yazılıyor, eksik olan tek şey okuma. Uçtan uca ücretsiz bir şeffaflık özelliği: mentör ve mentee için de açık, çünkü onlar ürünü zaten ücretsiz kullanıyor.

## 🤖 Agent prompt
```
Objective (junior-friendly): give every user a self-service "who accessed my account"
history built from the AuditLog rows impersonation already writes.

Open first:
  src/app/api/admin/impersonate/route.ts:41-61  <- IMPERSONATE_START row + the reason
  src/app/api/impersonate/stop/route.ts:19-21   <- IMPERSONATE_STOP row (gives the duration)
  prisma/schema.prisma:2477-2489                <- AuditLog; add @@index([targetId])
  src/app/api/account/devices/route.ts:13-19    <- copy this self-service endpoint shape
  src/components/AccountSettings.tsx:428-443, :691-700  <- the trusted-device card to sit next to
  src/lib/auth.ts:388-389                       <- the 30-minute cap (no stop row = auto-expired)
  src/i18n/dictionaries.ts

Constraints:
  - The endpoint must read session.user.id and accept NO user id parameter. This is personal
    data; a userId query param would be an IDOR.
  - Return the admin's display NAME only - not their email, not their id.
  - Available to every role. This is a free transparency control, never gated by plan or
    entitlement (the app is always free for mentors and mentees).
  - EN/TR/DE keys in src/i18n/dictionaries.ts or npm run check:i18n fails CI.
  - Schema change is one index. Additive, no DROP in the diff.
Done when: an e2e spec impersonates a mentee and that mentee's account page shows the row.
Open a PR (CLAUDE.md standing rule) and post the crm-pr<N>.ersah.in preview URL.
```

**Zorluk** · S · **Tahmini** · 0.5-1 gün · **Bağımlılık** · yok (bağımsız; #1583 ile aynı tabloya dokunur)

mersahin:
## Summary

Support impersonation is a legitimate feature and a trust liability at the same time. Today it writes `IMPERSONATE_START` / `IMPERSONATE_STOP` audit rows and tells the account holder once, in an in-app notification that scrolls away — after which the one party who cannot see what happened is the person whose account was entered. The rows were already there; only the read was missing.

Every user now has a read-only **Account access history** card on `/account`, next to the trusted-device list: who entered, when, how long the visit lasted, and why. It is available to every role — a transparency control is never a paid feature, and mentors and mentees use this product for free.

Closes #1587

## Changes

- **`GET /api/account/impersonations`** — the signed-in user's own rows only. It reads `session.user.id`, takes **no** user id parameter (a hopeful `?userId=` cannot widen it), and reads through an explicit `select` allowlist. No endpoint anywhere deletes these rows: an access record the accessing party can erase is worth nothing.
- **What the viewer learns about the admin — the judgement call, decided: the display NAME and nothing else.** Not the e-mail, not the id, not the role. Entering someone else's account is done in an official capacity, and accountability without a name is not accountability — "an administrator" (what today's notification says) gives the user nobody to ask about. A name is what a support conversation needs; an e-mail address additionally hands a staff member's direct contact *and* their login identifier to anyone who ever filed a ticket, and buys the user nothing they cannot get by asking. When the admin's account is gone the name is simply absent and the card says so rather than substituting an id. The reasoning is in the comment at the top of the route.
- **The reason is now written to `AuditLog.detail`** as well as the ActivityLog row, so "why" is readable from the account holder's own history rather than only from the admin-side log.
- **Pairing** (`src/lib/impersonationHistory.ts`, deliberately Prisma-free): starts pair with the following stop **per admin**, so two overlapping visits never cross-pair. A start with no stop renders as *"ended automatically"* with its duration clamped to the 30-minute session cap — never as a live session — and an orphan stop (its start off the window's edge) is dropped rather than shown as a zero-length visit. The 30-minute cap is now one exported constant shared with the JWT callback in `src/lib/auth.ts`.
- **Empty state.** `AuditLog` is *not* pruned today — it is one of the tables queued behind #1585 — so the empty state is the reassuring, unqualified one: "Nobody has ever signed in to your account except you." The endpoint carries a `retentionDays` field (null today) and the card already has the bounded copy ("…in the last N days. Anything older is no longer kept.") plus a footnote under a non-empty list; when #1585 registers an `auditLog` entry, setting the one literal in `impersonationHistory.ts` switches the copy. The two answers are never allowed to look the same.
- **Schema**: `AuditLog` gains `@@index([targetId])` — the column this list filters on, and the only schema change. Additive, no DROP.
- EN/TR/DE strings for every new string; e2e spec `e2e/account-access-history.spec.ts` walks the whole loop (admin impersonates a mentee with a reason, returns, the mentee signs in and finds the visit — plus the `?userId=` IDOR probe and an assertion that the mid-impersonation row reads as auto-ended); release fragment with EN/TR/DE notes.

Nothing was left out of the issue's scope.

## Verification

- [ ] `npm run test:e2e` passing (or CI green) — this container has no database and no browser, so the new spec was written but not executed; CI runs it
- [x] `npm run check:i18n` (4427 keys × 3 locales), `npm run check:release-fragments`, `npm run check:tenant-models` (no new drift), `npx prisma validate` all pass

Preview: https://pr2255.interncrm.com

## Contributor terms


mersahin:
Pushed `9ca4519`. All three findings were real — I reproduced each against the code before changing anything, and rejected none.

**A — `$` patterns in the reason (confirmed).** Reproduced exactly as filed:

```
'Reason given: {reason}'.replace('{reason}', "Ticket $' 12")        -> "Reason given: Ticket  12"
'Reason given: {reason}'.replace('{reason}', "Refund $& duplicate") -> "Reason given: Refund {reason} duplicate"
'Reason given: {reason}'.replace('{reason}', "Cost $` x")           -> "Reason given: Cost Reason given:  x"
```

`interpolate()` in `src/lib/notificationText.ts` was the right fix and was already there, just not exported — it now is, with a doc comment saying *why* it is the only safe way to do this here. Every template in the card goes through it (reason, duration, outcome, bounded note, empty state), not just the reason, so the next placeholder added to this card can't reintroduce it. The e2e spec's reason is now `'Ticket 4711 $& portal shows no mentor'`, asserted verbatim plus `getByText(/\{reason\}/)` → count 0, so the corruption can't come back unnoticed.

**B — a live session described as ended (confirmed).** `autoExpired: stop === null` ignored `now` entirely. `autoExpired: boolean` is now `outcome: 'closed' | 'open' | 'autoExpired'`:

- `closed` — a stop row exists, exact duration.
- `open` — no stop row, start inside the cap. Copy: *"No end has been recorded yet. A session that is not closed ends automatically after {minutes} minutes."* Deliberately **not** a claim that the session is active — an absent stop row is not evidence the admin is present, only that nothing was written. `{minutes}` comes from `IMPERSONATION_SESSION_MAX_MS`, so the string and the constant can't drift.
- `autoExpired` — past the cap, so the JWT callback has certainly reverted it. Only here is "ended automatically" a fact.

One case the finding didn't mention and I handled while in there: a **re-entry by the same admin** closes the earlier visit. The clock test alone would have called that earlier one `open` when both starts are minutes apart, which is wrong — the same admin demonstrably started again, so it's over. It stays `autoExpired`, and its duration is now bounded by the newer start's timestamp instead of by `now`, which is a strictly tighter upper bound.

`e2e/account-access-history.spec.ts:44` updated in the same commit — mid-impersonation it now asserts `/no end has been recorded yet/i` **and** that `/ended automatically/i` has count 0.

**C — silent truncation (confirmed), and the row-vs-session cap.** The endpoint returns `total` (a `count` on `targetId` + `IMPERSONATE_START`, served by the new `@@index([targetId])`), and the card renders *"Showing the {shown} most recent of {total} recorded sign-ins"* whenever they differ. `total` counts starts rather than sessions because exactly one session comes out of each start — orphan stops produce none — which is now pinned by an assertion rather than left as a comment.

The row cap was real, and I fixed it rather than explaining it away: `take: LIMIT * 2 + 1` meant each orphan stop at the older edge of the window ate a row and yielded nothing, so a user with 60 sessions could be shown 49 with no way to tell that apart from "there are 49". The boundary is now drawn on **starts** — read the Nth-newest start, then every `IMPERSONATE_*` row at or after its timestamp — which returns exactly `min(50, total)` sessions. Costs one extra read of the same index.

**New in CI:** `npm run test:impersonation-history` (10 cases, wired into `ci.yml`). The three outcomes are precisely what no e2e run can pin down — an unclosed start five minutes old and one five hours old are the same two columns in the database and mean opposite things, and Playwright can't hold the clock still across the cap boundary. It covers both sides of the boundary, exactly-at-cap, same-admin re-entry, two admins overlapping (no cross-pairing), orphan-stop drop, the one-session-per-start invariant `total` depends on, `$&`/`$'` in the reason surviving verbatim, and negative-duration clamping.

Release fragment prose updated — it described the old "a start with no stop renders as *ended automatically*" behaviour, which is now the thing this commit removes. Added a user-facing note in EN/TR/DE about unclosed visits, and extended the "permanent and read-only" note to mention the total.

**Static checks** (no DB, no browser): `npx tsc --noEmit` clean · `npm run lint` clean (only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`) · `npm run check:i18n` OK, 4429 keys × 3 locales · `npm run check:release-fragments` OK (the "compaction looks stuck: 32 fragments pending" line is repo-wide and the script itself flags it as not a problem with this PR) · `npm run test:impersonation-history` 10/10 · `check:auth-reads` and `check:query-scalars` OK. Schema untouched, so no `prisma validate` needed.

---

Issue #1678, opened 2026-08-30
🌱 [Infra] Task · One retention job for ActivityLog, PageView, PushSubscription and the queue's own tables


mersahin:
## 🎯 Goal
Exactly one table in this product is pruned. Three others hold personal data — browsing history, IP addresses, user agents — and grow forever. That is both a GDPR storage-limitation exposure and a slow capacity problem on a single MySQL box. This adds one scheduled retention job that prunes them on a configurable window, using the pattern `pruneEmailLog` already established — and it is **the one retention mechanism for the whole product**, the registry every other retention issue plugs into.

## 🔍 Current state
- The only prune in the codebase: `src/services/emailService.ts:3043` (`pruneEmailLog`, `EMAIL_LOG_RETENTION_DAYS = 90` at `:3041`), folded into the 09:00 tick at `:3074-3077`. Its doc comment at `:3035-3040` is the exact reasoning to reuse: the log answers a question asked within hours, but every row holds a recipient address.
- Not pruned:
  - `ActivityLog` — `prisma/schema.prisma:2412-2432`, stores `ip String? @db.VarChar(64)` and `userAgent String? @db.Text` (`:2419-2423`).
  - `PageView` — `prisma/schema.prisma:407-420`, per-user per-path browsing history with dwell time, consent-gated on write only.
  - `PushSubscription` — `prisma/schema.prisma:729-760`; `lastSeenAt` exists at `:747` and the schema comment says a row unseen for months "can be recognised as dead weight", but only 404/410 and 5 consecutive failures delete rows (`src/lib/webPush.ts:138`). Nothing sweeps by age.
- **Already filed, do not duplicate — each of these registers into the mechanism this task builds:** `AuditLog` is #1585 (CSV export + `auditRetentionMonths` + sweep), `Notification` is #1646 (setting + the unread/30-day rails), `DomainEvent` is #1691 (`DOMAIN_EVENT_RETENTION_DAYS` + prune helper), `Message` is #2056 (per-org window that **masks** with `deletedForEveryoneAt` instead of deleting). #2012 only *states* a retention policy in its docs and adds no sweep at all. This task covers the tables none of them claims, plus the queue's own `Job`/`JobRun`/`OutboxEvent` rows — and it provides the single `retention.prune` handler and its registry, which all four register into rather than adding four more schedules.
[...]

Issue #1561, opened 2026-08-30
⏱ [Infra] Task · Background jobs run per tenant — cron sweeps bind an org with runWithOrg


mersahin:
## 🎯 Goal
Every scheduled job in the product runs with **no tenant context**. The middleware treats "no bound org" as "do not scope", so with isolation on, the reminder sweep, the weekly digest, the dormant-first-contact sweep and the newsletter queue all still read across every tenant — and they read the *global* settings row to decide cadence, so tenant B's reminder policy is tenant A's. This is the one class of code where turning the flag on changes nothing, which is worse than a visible failure: the product looks isolated and the mail does not follow suit.

## 🔍 Current state
- `grep -rn "runWithOrg" src/` returns only `src/lib/orgContext.ts:133` (the definition) and `:161` (`withTenantScope`). **No job binds a context.**
- `src/lib/orgContext.ts:92-95` — `const orgId = currentOrgId(); if (!orgId) return next(params);` — undefined context ⇒ unscoped query.
- `src/services/emailService.ts:13` imports `getSetting` and reads the global cadence (`reminderDays`, `weeklyDigest`) for all tenants.
- `src/services/emailService.ts:46` — `emailBrand(orgId)` already resolves per-tenant branding, so the mail *looks* tenant-specific while the recipient set is not.
- The newsletter cron is registered from `/api/cron/start` (never from `initCronJobs`) per `docs/newsletter.md`; the dormant sweep is `sweepDormantFirstContacts` in `src/lib/dormantFirstContact.ts`.
- `src/app/api/cron/route.ts` and `src/app/api/cron/start/route.ts` are (correctly) not wrapped in `withTenantScope` — they have no user session to resolve an org from.

## 🛠 What to build
1. A small helper next to `runWithOrg`, e.g. `forEachOrg(fn)` in `src/lib/orgContext.ts`, that lists active organizations and runs `fn(org)` inside `runWithOrg(org.id, …)` sequentially, logging per-org failures without aborting the remaining tenants.
2. Convert each sweep to the per-org shape:
   - interaction reminders and the overdue-stage-deadline reminder,
   - the weekly digest,
   - `sweepDormantFirstContacts` (`docs/dormant-first-contacts.md` — the two hard rules stay: **no third mail**, **no automatic stage change**),
[...]

Issue #1211, opened 2026-08-09
Risk hardening for the email work: token lifetime, delivery-log retention, digest link count


mersahin:
Risk review of #1195 before it ships to production in one piece. Three findings worth closing in the same change set, plus one standing guard.

## 1. Email action tokens never expired

`/m/[token]` links (mark-as-read, emoji reactions) were signed but unbounded in time. A notification email that gets forwarded, or an old mailbox archive that leaks, was an indefinite licence to act as that user.

Severity is genuinely low — the actions are a reaction and a read flag, both reversible, and neither can post a message — but the exposure window never closed.

**Closed:** 90-day TTL. An expired link answers `410 Gone` and the page says "this link has expired, open it in the app" rather than the misleading "invalid link". 90 days, not hours, because a digest opened weeks later should still work.

## 2. `EmailLog` sat outside the retention and erasure rules

The new delivery log stores a recipient address and subject on every send. It is keyed by **address**, not by a relation or a `userId` — so nothing cascades to it:

- erasing an account (`hardDeleteUser` / `anonymizeUser`) left the address behind in the log;
- nothing pruned it, so addresses accumulated indefinitely next to the data the retention rules already govern.

**Closed:** both erasure paths clear it (reading the address *before* the row is deleted or rewritten — afterwards there is nothing to match on), and a daily job prunes rows older than `EMAIL_LOG_RETENTION_DAYS` (90).

## 3. The unread digest was undermining its own goal

Per-line reaction links meant a five-item digest carried **25 extra links**. A high link count is one of the strongest spam signals there is — so the change set built to fix inbox placement was pushing the digest toward the spam folder, and long digests risked Gmail's 102 KB clipping.

**Closed:** reactions stay on the single-message notification, where "the message this email is about" is unambiguous. The digest keeps its per-conversation "mark as read" link and nothing else.

## 4. Standing guard: the password column

`accountState` has to **read** `User.password` to tell a mentor-created record (no login) apart from an admin-deactivated account. That puts the bcrypt hash one spread operator away from a response, permanently.

Audited all three `/api/users` response paths plus `/api/users/[id]`, `resend-verification` and the messages counterpart lookup — none leak it. Added a regression spec that asserts no `/api/users` response contains a bcrypt prefix, matching on `$2a$`/`$2b$` rather than a key name so a rename cannot silence it.

## Residual risks (accepted, documented)

[...]

PR #2256, opened 2026-09-07
feat: notification retention window, nightly prune and history search (#1646)

### LINKED ISSUE #1646: 🌱 [Comms] Task · Notification retention setting, a nightly prune and search in the notification centre
## 🎯 Goal
`EmailLog` is pruned after 90 days, and the reason is written out in the code: every row holds personal data, so keeping them forever builds a second unmanaged store next to the one the retention rules already govern. `Notification` rows hold exactly the same kind of thing — a rendered sentence about a person, a link to their record — and are **never** deleted. Apply the same rule, make the window configurable per tenant, and while we are in the history page, add the search it never got.

## 🔍 Current state
- `src/services/emailService.ts:3036-3047`:
  ```ts
  export const EMAIL_LOG_RETENTION_DAYS = 90;
  export async function pruneEmailLog(retentionDays: number = EMAIL_LOG_RETENTION_DAYS) {
    const cutoff = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
    const { count } = await prisma.emailLog.deleteMany({ where: { createdAt: { lt: cutoff } } });
    return { deleted: count, cutoff };
  }
  ```
  There is no `pruneNotifications`.
- `prisma/schema.prisma:690-715` — `Notification { id, userId, type, text String? @db.Text, params Json?, link, read, createdAt, announcementId }`, `onDelete: Cascade` from `User`.
- `src/app/api/notifications/route.ts:21-32` — the GET accepts `page`, `pageSize`, `read` and `type`. The `where` is built at `:27-32`; there is no text filter.
- `src/app/notifications/page.tsx:37-61` — client state is `readFilter` and `typeFilter` only; `:88-97` builds the type options from `data.types`.
- `src/services/emailService.ts:3051` onward — `initCronJobs()` with the existing `cron.schedule` registry (`scheduledTasks`). **Do not add a schedule here**: #1678 introduces the one retention job and the registry this prune plugs into (see the scope boundary below).
- `scripts/sanitize-db.mjs:231-232` already treats a notification's `text`/`params` as personal data when sanitising a snapshot — the same judgement, applied to a different problem.

## 🛠 What to build
1. `notificationRetentionDays` in `SETTING_SCHEMA` (default `180`, `0` = keep forever), scoped `global|org`, resolved through the chain.
2. `pruneNotifications(retentionDays)` in `src/services/emailService.ts`, modelled exactly on `pruneEmailLog`, returning `{ deleted, cutoff }`. **Never delete an unread row** and never delete a row younger than 30 days regardless of the setting — an over-eager admin should not be able to erase this morning's bell.
3. Register it as **one entry in the retention registry from #1678** (`src/lib/retentionPrune.ts` — `{ key: 'notification', settingKey: 'notificationRetentionDays', run(cutoff) }`), rather than adding a second `cron.schedule` next to the email-log prune. The per-org resolution lives **inside this issue's own `run()`** — resolve the window per org inside the loop (see #1561; a global read at the top would apply the first tenant's window to everyone). #1678 supplies the schedule and the count reporting, not the tenant iteration. If #1678 has not landed when you start, export `pruneNotifications` with that `run(cutoff)` signature so registering it later is a one-line change, and say so in the PR.
4. Log the run through `logActivity` so the deletion is itself auditable — the registry's `retention.pruned` row already carries the per-table counts, so add detail rather than a second row.
5. **Search:** add `q` to `GET /api/notifications` — a `contains` match on `text`, plus a match on the rendered type label is out of scope (rows written through the i18n contract have `text = null`; say so in the response docs and in a code comment). Cap `q` length, and keep the existing default response shape untouched — `NotificationBell` depends on it (comment at `src/app/api/notifications/route.ts:13-16`).
6. A search box on `/notifications` with a `data-testid`, debounced, resetting to page 1, included in the existing "clear filters" control at `src/app/notifications/page.tsx:70`. EN/TR/DE labels.
7. Document the retention window on the settings screen next to `retentionMonths` so the two are not confused (that one is about candidate records, this one about notification rows).

## ✅ Acceptance criteria
- [ ] A notification older than the configured window and already read is deleted by the nightly job; an unread one is kept
- [ ] Nothing younger than 30 days is ever deleted, whatever the setting says
- [ ] Two orgs with different windows each prune to their own cutoff
- [ ] The prune runs from #1678's `retention.prune` job — there is no second retention schedule
- [ ] The prune run is recorded in the activity log with the deleted count
- [ ] `GET /api/notifications` with no params returns exactly the shape it returns today (the bell keeps working)
- [ ] Searching for a word in a notification's text finds it and only it; the page resets to page 1
- [ ] `npm run build` passes
- [ ] an e2e spec covers the search and a unit test covers the prune cutoff
- [ ] a release fragment `releases/unreleased/notification-retention-and-search.json` is added

## 🇹🇷 Türkçe özet
`EmailLog` 90 günde budanıyor ve gerekçesi kodun içinde yazılı: her satır kişisel veri taşıyor, sonsuza kadar saklamak yönetilmeyen ikinci bir veri deposu kurmak demek. `Notification` satırları tam olarak aynı türden veri taşıyor — biri hakkında yazılmış bir cümle ve kaydına giden bir link — ama hiç silinmiyor. Aynı kuralı uyguluyoruz: kiracı başına ayarlanabilir bir saklama penceresi, budama fonksiyonu ve silme işleminin kendisinin de log'lanması. İki güvenlik freni: okunmamış bildirim silinmez ve ayar ne derse desin 30 günden yeni satıra dokunulmaz. **Zamanlayıcı bu issue'ya ait değil**: `pruneNotifications` ikinci bir gecelik cron olarak değil, #1678'deki tek saklama kaydına (registry) bir girdi olarak ekleniyor; kiracı döngüsü ise bu issue'nun kendi `run()` fonksiyonunun içinde kalıyor. Bir de geçmiş sayfasına hiç eklenmemiş olan metin aramasını ekliyoruz. Dikkat: parametresiz `GET /api/notifications` yanıtının biçimi değişmemeli, zil bileşeni ona bağlı.

## 🤖 Agent prompt
```
Objective: a per-tenant notification retention window with a prune registered into the
shared retention job, and text search in the notification history.

Open first:
  src/services/emailService.ts:3036-3047  pruneEmailLog (copy this shape)
  issue #1678 + src/lib/retentionPrune.ts THE retention registry and the one daily
                                          retention.prune job — your prune is ONE entry there
  prisma/schema.prisma:690-715            model Notification
  src/app/api/notifications/route.ts:13-32  (the default shape the bell depends on)
  src/app/notifications/page.tsx:37-61, :70, :88-97
  src/lib/settings.ts / src/lib/settingsResolver.ts (#1619)
  scripts/sanitize-db.mjs:231-232 (notifications already treated as PII)
  related: #1561 (cron sweeps bind an org)

Constraints that WILL trip you up:
- Do NOT add a cron.schedule in initCronJobs. #1678 owns the single retention schedule;
  you register pruneNotifications into its registry. If #1678 has not merged, export the
  function with the registry's run(cutoff) signature and note it in the PR.
- The sweep iterates tenants: resolve the retention window INSIDE your own run(), per org,
  or every tenant gets the first one's window.
- Never delete an unread notification, and never delete anything younger than 30
  days regardless of the setting.
- GET /api/notifications with no query params MUST keep its exact current shape
  — NotificationBell depends on it (see the comment at route.ts:13-16).
- Rows written through the i18n contract have text = null and carry `params`
  instead; a `contains` search cannot see those. Say so in a comment rather than
  pretending the search is complete.
- E2E locator pitfall: an unscoped input[type="search"] selector hits the
  AdminNav sidebar filter on admin pages. Give the new box a data-testid and
  target that. getByText is substring matching — use { exact: true }.
- New strings need EN/TR/DE keys or `npm run check:i18n` fails CI.
- Add a release fragment (bump: "minor").

Done when: the shared retention job prunes per tenant with both safety rails, the run is
audited, the bell is untouched, and search works from the history page.
```

## 🔗 Scope boundary

#1678 and this issue both prune a table on a nightly schedule. **#1678 owns the retention mechanism** — `src/lib/retentionPrune.ts` (the registry), the single daily `retention.prune` schedule and the one `retention.pruned` `ActivityLog` row carrying every table's count. This issue owns the **`Notification` rule only**: the `notificationRetentionDays` setting, `pruneNotifications()` with its two rails (never an unread row, never anything younger than 30 days) and the per-org window resolution inside its own `run()` — registered as one entry in #1678's registry instead of a second `cron.schedule` beside the email-log prune. Do not create a second retention job or a second prune module. The notification-centre search in this issue is entirely its own and touches nothing of #1678's.

**Zorluk** · S · **Tahmini** · 1-1.5 days · **Bağımlılık** · #1619; saklama işi için #1678

mersahin:
## Summary

`Notification` was the one table this product never deleted from, while its rows carry the
same category of personal data `EmailLog` has been pruned for since #1211 — a rendered
sentence about a person plus a link to their record. This gives it a window, prunes it from
the shared nightly retention job, and adds the text search the notification centre never got.

The window is `notificationRetentionDays`: default **180** (matching `PageView`, the other
per-user history table, so the product defends one number rather than two), `0` keeps
forever. Unlike the four telemetry windows of #1678 it **is** on the admin settings form —
how long a bell keeps its history is a programme decision an org has an opinion about — and
it sits directly under `retentionMonths`, whose neighbour hint now spells out which of the
two is which (that one is candidate records; this one is notification rows).

**The prune is one entry in #1678's registry.** `src/lib/retentionPrune.ts` owns the
mechanism, the single 03:20 schedule and the one combined `retention.pruned` activity row;
this change adds `{ key: 'notification', settingKey: 'notificationRetentionDays', run }` to
`retentionEntries.ts` and nothing else — no second cron, no second prune module, no second
reporting style. The deleted count rides in the existing summary line, which is what the
issue asks for ("add detail rather than a second row"). The issue's own text says to put
`pruneNotifications` in `emailService.ts`; the registry that landed since says a new table
arrives as one more entry next to the other five, so it lives in `retentionEntries.ts` and
takes the registry's `run(ctx)` signature.

**The window is resolved per org inside the entry's own `run()`**, not once by the runner:
the runner's `resolveDays` is right for the instance-wide telemetry tables and wrong here,
because a notification belongs to a user and a user belongs to a tenant — a single read at
the top would hand the first tenant's choice to everybody (#1561). `pruneNotifications`
walks the orgs, re-resolves the setting for each, and finishes with the `null` scope (users
not bound to an org — a single-tenant installation reads the global layer exactly as before).
The run budget is shared across scopes, so one large tenant cannot starve the job.

### What is never deleted

Three rails, none of which any setting can lower. They are stated with their reasoning in
the new **Prisma-free** `src/lib/notificationRetention.ts`, split out for the same reason
the mechanism was — so they can be unit-tested without a database:

1. **An unread row is never deleted**, at any age. It is still the notification the person
   has not seen.
2. **Nothing younger than 30 days is deleted**, whatever the form says
   (`NOTIFICATION_RETENTION_FLOOR_DAYS`). The setting is on a form, so a number typed in a
   hurry reaches the job unreviewed; this is the rail that makes the worst typo survivable.
3. **`RETAINED_NOTIFICATION_TYPES` is never deleted at all.** Most notifications are a
   convenience — the fact behind them lives in another row and the bell entry is a pointer.
   These are not:
   - `retention.confirm`, the GDPR re-consent prompt. Its link carries the renewal token, so
     an unanswered one is live work; the e-mail copy of the same notice is itself pruned at
     90 days, and `User.retentionReminderSentAt` records a date and nothing else — it cannot
     show *what* the person was asked. Deleting the row leaves a consent request we can no
     longer evidence.
   - `impersonation.accessed` / `impersonation.accessedWithReason`, the notice that an
     administrator entered somebody's account. `ActivityLog` has the operator-side record,
     but that is admin-only: this row is the account holder's only copy, and a transparency
     notice its subject can no longer see has stopped being one.

   `security.*` is deliberately **not** in that list, and the comment says why: those mirror
   an admin action the security ledger already keeps for a year under its own longer window,
   and the user-facing copy records no right exercised and no consent given. Widening the
   exception past rows that are genuinely sole evidence would quietly turn the window off.

### Search

`q` on `GET /api/notifications` — a length-capped (100 chars) `contains` on `text`, folded
into the existing `where` next to `read`/`type` rather than a parallel path. The box on
`/notifications` is debounced at 300ms, resets to page 1 in the same tick as the query
change (one fetch, not two), and is part of the existing "clear filters" control; a search
that matches nothing gets its own empty state so it reads differently from an empty inbox.

`q` **cannot see rows written through the i18n contract** (#921): those carry `params` and
`text = null`, and their sentence is rendered from the dictionary in the reader's locale, so
it exists in the browser and never in a column. Both the route's doc comment and the `where`
say so; matching the rendered label would mean re-rendering every row in every locale
server-side, which the issue puts out of scope. The type filter remains the usable handle on
templated rows.

The parameterless response shape is byte-for-byte what it was, so `NotificationBell` is
untouched — asserted in the new e2e alongside the search itself.

Closes #1646

Preview: https://pr2256.interncrm.com

## Changes

- `src/lib/notificationRetention.ts` **(new)** — the rule: the 30-day floor, the retained
  types with their reasoning, and `notificationRetentionCutoff()`. No Prisma import.
- `src/lib/retentionEntries.ts` — `pruneNotifications(ctx)` (per-org window, batched through
  the registry's `pruneInBatches`) and its registry entry with the required `reason`.
- `src/lib/settings.ts` — `notificationRetentionDays: '180'` with the reason for the default.
- `src/app/api/admin/settings/route.ts` — validation for the new key.
- `src/app/admin/settings/page.tsx` — the field, immediately under `retentionMonths`.
- `src/app/api/notifications/route.ts` — the `q` parameter and what it can and cannot see.
- `src/app/notifications/page.tsx` — the debounced search box (`data-testid`
  `notifications-search`) and the search-specific empty state.
- `prisma/schema.prisma` — `@@index([createdAt])` on `Notification`. Without it every batched
  select in the sweep is a filesort over the whole table, which is exactly the load the
  batching exists to avoid. No migration (this repo uses `db push`).
- `src/i18n/dictionaries.ts` — EN/TR/DE for the search box, its empty state and the setting.
- `e2e/notification-retention.unit.spec.ts` **(new)** — the cutoff rule with no database:
  the floor, `0` = keep forever, a corrupted value, and the retained-type list by name.
- `e2e/notifications-history.spec.ts` — two search specs: narrowing + clearing + the
  no-match state, and page-reset + cross-user isolation + the untouched default API shape.
- `docs/pii-access-lifecycle.md` — the window in the table, a section on the two rails and
  the evidence exception, and the "how to add a table" example moved off `notification`
  (which is now real) onto `domainEvent`.
- `releases/unreleased/notification-retention-and-search.json` — `minor`, with EN/TR/DE notes.

## Verification

- [ ] `npm run test:e2e` passing (or CI green) — **not run here**: this container has no
      database and no Docker daemon, so the app and Playwright cannot start. The specs are
      written and left for CI.
- [x] `npm run check:i18n` (4418 keys × 3 locales), `npm run check:release-fragments`,
      `npm run check:tenant-models` (no new drift), `npx prisma validate` — all clean.
- [x] Added an e2e spec for the search and a unit spec for the prune cutoff.

## Contributor terms


mersahin:
Review follow-up pushed as `a7b07a2`. All five findings verified against the code and fixed — none rejected as a false positive.

1. **The search only ever matched announcements, and said nothing about it** (major). Confirmed: the repo has exactly two `Notification` writers, and `notify()` (`src/lib/notify.ts:17`) never sets `text`. Added `notifications.searchHint` in EN/TR/DE, rendered as the `Input` hint under the box and repeated as a sibling of the empty state (a sibling, not inside it — `notifications-history.spec.ts` asserts that element's exact text). `searchNone` itself is unchanged. The release fragment's "both the route **and the UI** say so" clause now describes what actually ships.
2. **No database-backed test for either irreversible rail** (minor). Confirmed, and worse than reported: `e2e/retention-prune.spec.ts`'s entry-key assertion still listed five keys and would have **failed** on the sixth. Fixed, plus four notification rows for the org-less user (400d+read goes; 400d+unread, 400d+`retention.confirm`, 1d+read stay) and a new test seeding three tenants at 30 / 365 / 0 days whose 60-day rows resolve three different ways in one run — an outcome no single global cutoff can produce. The unit spec's claim that the unread rail is "exercised against a real database" is now true and names the file.
3. **Unguarded out-of-order responses** (minor). Confirmed. `load` now carries a sequence number; a superseded response writes no state and does not clear the newer request's spinner.
4. **Scanned prefix grows forever, re-walked per org** (minor). Confirmed. `@@index([createdAt])` → `@@index([read, createdAt])`, so the rows the sweep may never delete are outside the scanned range rather than permanently inside it, and scopes resolving to the same cutoff are swept in one select — one index walk a night on an instance where nobody overrode the default, instead of one per tenant. Per-scope *resolution* is unchanged (#1561).
5. **The keep-forever note never reached the audit row** (minor). Confirmed. `formatRetentionSummary` now renders an entry's `note` after its counts and treats a note as enough to make the entry non-silent, so `notification=0(off:N/M)` distinguishes "every tenant has this switched off" from "nothing to prune". Note shortened to fit the shared 191 characters.

`npx tsc --noEmit`, `npm run lint`, `npm run check:i18n`, `npm run check:release-fragments`, `npm run check:tenant-models` and `npx prisma validate` are all green. The e2e suite cannot run here (no database in the container).

---

PR #2257, opened 2026-09-07
feat: list and revoke push devices on /account (#1716)

### LINKED ISSUE #1716: 🌱 📱 [Comms] Task · Push device list on /account — "Chrome on Android", last seen, revoke
## 🎯 Goal

The schema promises a feature that has no endpoint and no UI. A user who enabled push on
three devices cannot see them, cannot tell which is which, and cannot turn one off without
turning push off everywhere. Small, isolated, entirely additive — and the honest completion
of a half-built feature the schema itself documents.

## 🔍 Current state

- `prisma/schema.prisma:741-743`:
  ```prisma
  // Which browser this is, so /account can list "Chrome on Android" rather than
  // an opaque endpoint. Free-form and untrusted — display only.
  userAgent    String?  @db.VarChar(255)
  ```
  The column exists **specifically** for a list that was never built.
- `prisma/schema.prisma:745-747` — `lastSeenAt DateTime @default(now())`, "bumped every
  time the browser re-confirms the subscription".
- `src/app/api/push/subscribe/route.ts:25` (POST) and `:59` (DELETE) — the **only** two
  endpoints. There is no GET, so nothing can list a user's devices.
- `src/components/AccountSettings.tsx:197` and `:935` — the browser-notification and push
  controls live here; the device list belongs directly under them.
- `src/lib/webPush.ts:138` `MAX_FAILURES` — dead endpoints are already pruned on 404/410
  and after repeated failures, so a listed device is a device that plausibly still exists.

## 🛠 What to build

1. `GET /api/push/subscribe` — return the caller's own subscriptions:
   `{ id, label, lastSeenAt, current }`. **Never the endpoint, `p256dh` or `auth`** — those
   are delivery secrets and have no business in a JSON response.
2. `label` is derived server-side from `userAgent` by a tiny pure helper
   (`src/lib/pushDeviceLabel.ts`): a short allowlist match producing "Chrome on Android",
   "Safari on iPhone", "Firefox on Windows", falling back to a generic "Browser". The
   column is documented as *untrusted, display only* — never render it raw.
3. `current: true` for the subscription matching the endpoint the page's own service worker
   holds, so the user can tell "this device" from the others.
4. `DELETE /api/push/subscribe?id=<id>` — revoke **one** device. Must verify the row
   belongs to the caller (`where: { id, userId: session.user.id }`), server-side. The
   existing endpoint-based DELETE stays for the service worker's own unsubscribe.
5. Render the list in `AccountSettings` under the push toggle: label, relative last-seen,
   a revoke button, and an empty state. EN/TR/DE keys for all of it.

## ✅ Acceptance criteria

- [ ] A user with two subscribed browsers sees two rows, one marked as the current device
- [ ] Revoking one device stops push to that device only; the other keeps working
- [ ] The API never returns `endpoint`, `p256dh` or `auth`
- [ ] Revoking a subscription id that belongs to someone else returns 404, not 200 —
      checked on the route, not in the component
- [ ] `userAgent` is rendered only through the label helper, never as raw text
- [ ] `npm run build` passes
- [ ] an e2e spec covers list + revoke
- [ ] a release fragment `releases/unreleased/push-device-list.json` is added

## 🇹🇷 Türkçe özet

Şema, olmayan bir özelliği vaat ediyor: `PushSubscription.userAgent` alanı "hesap
sayfası 'Chrome on Android' yazabilsin diye" var ama ne bir endpoint ne de bir arayüz
yapılmış. Üç cihazında push açan kullanıcı hangisinin hangisi olduğunu göremiyor ve
birini kapatmak için hepsini kapatmak zorunda. Küçük, izole ve tamamen ekleme yapan bir
iş — stajyer için iyi bir ilk görev. Dikkat: `endpoint`, `p256dh` ve `auth` teslimat
sırlarıdır, JSON yanıtında yeri yoktur; silme yetkisi rotada kontrol edilir.

## 🤖 Agent prompt

```
Objective: let a user see and revoke their individual push devices on /account.

This is a good first issue: small, additive, no schema change, no new dependency.

Open first:
  prisma/schema.prisma:729-758   (PushSubscription - read the comments, they describe
                                  exactly this feature)
  src/app/api/push/subscribe/route.ts  (POST at :25, DELETE at :59 - add GET, and a
                                        by-id DELETE)
  src/components/AccountSettings.tsx around :197 and :935 (where push controls live)
  src/lib/webPush.ts  (sendPushToUser :79, pruning at :138)
  src/lib/browserNotifications.ts
  docs/realtime-and-push.md
  src/i18n/dictionaries.ts (add EN/TR/DE keys for every new label)

Constraints that will trip you up:
  - NEVER return endpoint, p256dh or auth from the API. They are delivery secrets.
  - Authorization is server-side: scope every query and the delete by
    `userId: session.user.id`. Do not filter in the component.
  - userAgent is untrusted free text. Map it through a small pure helper to a fixed set
    of labels; never render it raw (XSS + garbage UX).
  - Every new string needs EN/TR/DE keys or `npm run check:i18n` fails CI.
  - No schema change is needed. Do not add a column; do not run db push.
  - In AccountSettings, keep any description <p> a SIBLING of its <label> - Playwright's
    hasText is a substring match and e2e/notif-prefs.spec.ts dies on strict mode
    otherwise. Give new controls a data-testid.

Done when: GET lists devices, by-id DELETE revokes one, secrets never leave the server,
build/lint/check:i18n pass, an e2e spec covers list+revoke, a release fragment is added,
and a PR is open with `Closes #<this issue>`.
```

**Zorluk** · S · **Tahmini** · 0.5-1 day · **Bağımlılık** · yok

mersahin:
## Summary

`PushSubscription.userAgent` has carried the comment *"so /account can list 'Chrome on Android' rather than an opaque endpoint"* since #1464, but nothing ever read it. There was no `GET` on `/api/push/subscribe`, so someone who had granted push on a laptop, a phone and a work machine could see none of them and could only switch all three off at once. This finishes the half-built feature the schema already documents.

The shape follows the trusted-device list that sits a card away on the same page (`docs/remember-me.md`) rather than inventing a second device-management UI: same list markup, same "this device" badge, same per-row revoke button, same `list(userId)` / `revoke(id, userId)` split in a lib.

Closes #1716

Preview: https://pr2257.interncrm.com

## Changes

- **`GET /api/push/subscribe`** returns the caller's own subscriptions as `{ id, label, createdAt, lastSeenAt, current }`, newest first. The projection is an explicit `select` allowlist in the new `src/lib/pushDevices.ts`, so `endpoint`, `p256dh` and `auth` — the credential for pushing to that browser — never leave the server, and a column added to the model later cannot quietly start leaking. `enabled` (from `pushConfigured()`) rides along so a deployment with no VAPID keys hides the section instead of showing a list nobody can fill; rows are still returned in that case, so turning push off cannot strand rows a user can no longer revoke.
- **Labels are derived server-side**, never rendered raw. The regex table moved out of `trustedDevice.ts` into a new pure, dependency-free `src/lib/deviceLabel.ts` and is now shared by both lists — a person should recognise "Chrome on Android" by the same words in both, and two copies of these regexes would drift. `trustedDevice.ts` re-exports `deviceLabel`, so every existing import is unchanged.
- **`DELETE /api/push/subscribe?id=<id>`** revokes one browser. The query is scoped `{ id, userId }` **on the route**, so another account's id deletes nothing and returns 404 rather than a silent 200 — an id that is not yours and an id that no longer exists are deliberately indistinguishable. The service worker's existing endpoint-body `DELETE` is untouched and still idempotent.
- **The list in `AccountSettings`**, under the browser-notifications toggle: label, "Added … · last seen …", a "this device" badge, a Revoke button and an empty state. Deliberately *not* nested inside the `browserNotifSupported` block — a phone can hold a subscription that a desktop browser without the Notification API must still be able to see and switch off.
- **EN/TR/DE** for all nine new strings.
- An e2e spec, `e2e/push-devices.spec.ts`, and a release fragment.

### Three decisions worth reviewing

**No schema change.** The brief allowed adding a `label` column captured at subscribe time, but the schema already stores `userAgent` and the issue explicitly rules a column out. Deriving at read time is a handful of regexes over at most a few rows, once per `/account` load, and it avoids a `db push` plus a backfill that would have left every existing row labelled "Browser" until its next re-subscribe. `POST` already captures the user-agent, so there was nothing to add.

**`current` is matched from a header, not the session.** The server cannot tell which of a person's browsers is asking — a push subscription belongs to a browser profile, not to a person — so the page sends the endpoint its own service worker already holds in an `x-push-endpoint` header and the comparison happens server-side. A header rather than a query parameter on purpose: query strings end up in access logs, and this value is half of a delivery credential. It is shape-checked, compared only against rows the caller already owns, and never echoed back.

**Revoking must actually stop pushes.** There is no cache to invalidate — `sendPushToUser` reads the table on every send — so deleting the row is the whole of the revocation. The real hazard was the opposite direction: revoking *the browser you are sitting at* would have been undone by the silent re-subscribe that runs on the next `/account` visit, handing the server back the row just deleted. So revoking the current device also unsubscribes it locally and clears the local notification preference, which is the same single switch that turned both halves on.

**"Revoked here" is not "expired there".** #1678's retention work and the push service's own 404/410 rejections both delete rows nobody revoked, so a `DELETE` that 404s is reported as *"That browser was already unsubscribed — it had expired on its own"*, and the hint under the list says a browser can drop off by itself when its site data is cleared or the push service stops accepting it.

### Collateral

- `playwright.config.ts` now **generates a throwaway VAPID keypair per run** and passes it to the e2e server, because `pushConfigured()` is false without one and the section hides itself, leaving the spec asserting against nothing. Generated rather than hard-coded: `web-push` rejects anything that is not a real P-256 key at `setVapidDetails` time, and a private key belongs in no repository. Nothing is ever delivered with it — no spec creates a real browser subscription.
- `e2e/message-unread-notification.spec.ts` had a global `pushSubscription.count() === 0` assertion, which the new spec's seeded rows would break from a parallel worker. Scoped to its own user.

### Not done

No entry was added to `src/lib/features.ts`: push notifications are not in the feature catalogue at all, and adding a landing-page card for a device list without one for the feature it manages would be out of proportion. Worth a separate issue if the catalogue should cover push.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint reports only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`)
- [ ] `npm run test:e2e` passing (or CI green) — no database and no browser in the authoring container, so `e2e/push-devices.spec.ts` was written and `--list`ed but not executed; CI is the first real run
- [x] Tested the change manually / added an e2e test — an e2e test covering list, ordering, the current-device flag, the 404 on someone else's id, per-device revoke and the empty state
- [x] `npm run check:i18n` — OK, 4422 keys × 3 locales
- [x] `npm run check:release-fragments` — `releases/unreleased/push-device-list.json` validates (the "compaction looks stuck" warning it prints is pre-existing and unrelated)
- [x] `npm run check:tenant-models` — no new drift

## Contributor terms


mersahin:
Both review findings addressed in 941c26a.

**A — the device list disappeared exactly when it was needed.** Confirmed, and worse than "a hidden section": the section was mounted behind `pushDevicesEnabled`, which is set from the last GET's `enabled` flag *and* was reset to `false` whenever that GET came back non-OK. So the two failure modes were:

1. A deployment whose VAPID keys went away reports `enabled: false`. The route already returned the rows in that case, with a comment saying it does so "because a deployment that turned push off should not strand rows a user can no longer revoke" — and then the page threw them away. The route's promise and the page's behaviour were in direct contradiction.
2. `res.ok === false` on the refresh GET set `enabled` false and `devices` `[]`. The GET most likely to fail is the one fired immediately after a successful revoke, i.e. while the user is looking at the list — a transient 500 there unmounted the whole section, taking the *other*, still-subscribed browsers off screen with it.

Fixed by decoupling the two facts on the page:

- The section renders when `pushDevicesEnabled || rows.length > 0`, and is hidden only when push is off **and** there is nothing on record (nothing to revoke, nothing that could arrive). When rows exist with push off, an amber note says push is switched off on this server and nothing is being delivered, but the browsers are still on record and revocable — `push-devices-unavailable`.
- A failed GET no longer overwrites what is known. The last answer stays on screen and a `push-devices-stale` note says it may be out of date; only a successful GET clears it.
- A revoked row is removed from local state *before* the refresh, so a revoke that succeeded shows a correct list even if the reload behind it fails.

**The response shape needed no change.** `{ enabled, devices }` already expresses "the operator has push configured" and "I have subscriptions on record" independently — only the client was collapsing them into one boolean. I corrected the route's JSDoc instead, since it described the old (hide-the-section) behaviour.

**B — docs.** Confirmed: `### Cihaz listesi (#1716)` had been inserted between the `src/lib/deviceLabel.ts` bullet and the last two bullets of `### Parçalar`, leaving `public/sw.js` and `src/lib/pushNotifications.ts` stranded under the wrong heading. The subsection now follows the complete list, and gained a paragraph documenting the `enabled`-vs-rows split.

**Also:** a second e2e spec (`e2e/push-devices.spec.ts`) routes the GET to a 500 while letting the DELETE through, then asserts the section, the remaining browser and its Revoke button all survive the revoke — and that the remaining browser is still revocable while the GET is down. Two new i18n keys in EN/TR/DE. The release fragment's changelog and user notes now describe the rows-decide-visibility rule.

Static checks: `npx tsc --noEmit` clean (it covers `e2e/` too), `npm run lint` clean, `npm run check:i18n` OK (4424 keys × 3 locales), `npm run check:release-fragments` exit 0 — it prints the repo-wide "release compaction looks stuck: 32 fragments pending" warning, which it labels as not a problem with this PR. No schema touched, so no `prisma validate` run. The e2e specs need a seeded DB and were not executed here.

---

Issue #1696, opened 2026-08-30
🚦 [Infra] Task · Rate limiting behind a shared store, with the in-memory limiter as fallback


mersahin:
## 🎯 Goal
Rate limits are a security control on 28 endpoints — sign-in, registration, password reset, the public application form, `/api/v1`. They are counted in one process's memory, so the moment a second replica exists every limit silently becomes twice as generous, and a redeploy resets them all to zero. Move the counter behind a store both replicas can see, without making a single-container install depend on a service it does not have.

## 🔍 Current state
- `src/lib/rateLimit.ts:10-13` states the problem in its own words: *"Simple in-memory fixed-window rate limiter. Per-process (fine for a single container); resets on redeploy. Not distributed — for stronger guarantees move to Redis. Keyed by client IP + a bucket name."*
- The implementation: a module-level `Map` of `{ count, resetAt }` (`:17`), fixed-window `rateLimit(key, { limit, windowMs })` at `:36`, opportunistic housekeeping every 100 calls with a 50k-bucket ceiling (`:20-33`), and coalesced breach auditing into `ActivityLog` (`:54-60`, one row per key per minute).
- Client IP resolution is proxy-hop aware — `src/lib/clientIp.ts`, `TRUSTED_PROXY_COUNT` in `.env.example:203`, hostile-XFF handling documented in `infra/README.md`.
- Applied to 28 endpoints (`grep -rn enforceRateLimit src/app/api | wc -l`), including `src/app/api/auth/forgot/route.ts:13`, `src/app/api/auth/reset/route.ts:24`, `src/app/api/register/route.ts:41`, `src/app/api/v1/candidates/route.ts:9`, `src/app/api/company-inquiry/route.ts:37`.
- k6 deliberately excludes rate-limited routes because all VUs share the runner's IP — `CLAUDE.md` (k6 section) and `docs/testing.md`. That constraint does not change here.
- #1547 is adding `enforceRateLimit` to five currently unprotected write endpoints — it will land on whatever interface this task produces, so keep the public signature stable. It is **not** blocked on this task (see Scope boundary).

## 🛠 What to build
1. **`src/lib/rateLimitStore.ts`** — a tiny interface (`incr(key, windowMs) → { count, resetAt }`) with two implementations:
   - `MemoryStore` — the existing `Map`, moved verbatim including the housekeeping and the 50k ceiling.
[...]

Issue #1599, opened 2026-08-30
🌱 📊 [Infra] Task · Coverage thresholds on the wave-0/1 modules, wired into the CI gate


mersahin:
## 🎯 Goal
A unit runner nobody has to keep green rots in a month. This wires `npm run test:unit` into the PR gate and puts a coverage floor under exactly the modules the three foundational refactors are about to rewrite — so a refactor that quietly deletes a rule also deletes its test, and CI says so.

## 🔍 Current state
- `.github/workflows/ci.yml:19-111` runs, in order: `npm ci`, `npx prisma validate`, `npm run lint`, `npx tsc --noEmit`, `check:i18n`, `check:release-fragments`, `test:release`, `check:auth-reads`, `check:query-scalars`, `check:demo-blocklist`, the k6 install + `check:k6`, `check:openapi`, three `infra/test/*.sh` shell tests, then `npm run build`. Each step carries a comment explaining the incident that motivated it — follow that style.
- `.github/workflows/ci.yml:26-31` checks out with `fetch-depth: 0` because the release gate dates each pending change from its commit.
- No coverage tooling exists (see the parent story #1592 and its first task #1598, which introduces Vitest and `@vitest/coverage-v8`).
- The modules Waves 0–1 will rewrite, all pure and all currently untested:
  - `src/lib/orgContext.ts` (162 lines) — `TENANT_MODELS` at :43-55 lists 11 models; #1549 adds 8 more and flips `MT_ENFORCE_ISOLATION`
  - `src/lib/entitlements.ts` (49 lines) and `src/lib/planGate.ts` (59 lines) — the two disagreeing billing subjects the programme has to collapse
  - `src/lib/pipeline.ts` (241 lines) — stage keys, labels and ordering; #1588 adds transition gates on top
  - `src/lib/dormantFirstContact.ts` (171 lines) — the "two mails, never a third" rule documented in `docs/dormant-first-contacts.md`
  - `src/lib/relativeTime.ts` (51 lines) — the locale-aware date helper the i18n guard (#1597) will push everyone onto

## 🛠 What to build
1. In `vitest.config.ts`, enable `test.coverage` with the `v8` provider, `reporter: ['text', 'json-summary']`, and `coverage.include` limited to the module list above (a repo-wide percentage would be a number nobody acts on).
[...]

Issue #1598, opened 2026-08-30
🧪 [Infra] Task · Add Vitest and move the 8 unit specs out of the Playwright suite


mersahin:
## 🎯 Goal
Make it possible to test a pure function without launching Chromium. Eight suites of pure logic currently boot a browser on every PR for no reason other than module resolution, and any new pure module has the same bad choice: a browser test or no test.

## 🔍 Current state
- `package.json:13-40` — the scripts block has `test:e2e`, `test:e2e:smoke`, `test:e2e:headed`, `test:stress`, `test:load`, `test:release` and nine `check:*` gates. **No unit test runner dependency exists.**
- `scripts/test/release-derive.test.mjs` runs under `node --test` (`npm run test:release`) — the only non-browser test in the repo.
- Eight files match `e2e/*.unit.spec.ts`:
  - `e2e/a11y-ci-gate.unit.spec.ts`
  - `e2e/availability-slots.unit.spec.ts`
  - `e2e/duplicate-detection.unit.spec.ts`
  - `e2e/email-groups-footer.unit.spec.ts`
  - `e2e/meeting-link-hybrid.unit.spec.ts`
  - `e2e/notification-link.unit.spec.ts`
  - `e2e/notification-text.unit.spec.ts`
  - `e2e/unsubscribe-token.unit.spec.ts`
- `e2e/unsubscribe-token.unit.spec.ts:1-8` states the reason in its header: *"pure node, no browser, no database. (Playwright resolves the tsconfig `@/` paths, which `node --test` cannot…)"*. It imports from `@/lib/emailGroups`, `@/lib/notificationPrefs` and the unsubscribe-token module — all pure.
- `playwright.config.ts:35-59` — `testDir: './e2e'`, `workers: 1`, `timeout: 60_000`, `globalSetup: './e2e/global-setup.ts'`, `storageState: './e2e/.state/consent.json'`. The global setup and storage state are irrelevant to these eight files but they pay for them anyway.
- `.github/workflows/ci.yml:36-111` — `npm ci`, `prisma validate`, `lint`, `tsc --noEmit`, nine checks, `npm run build`. No place to hang a unit test step yet.

## 🛠 What to build
1. Add dev dependencies: `vitest` and `@vitest/coverage-v8`.
2. Add `vitest.config.ts` at the repo root:
   - `resolve.alias` maps `@` → `./src` (mirror `tsconfig.json`'s `paths`; do not rely on a plugin guessing it).
[...]

PR #2259, opened 2026-09-07
ci: per-module unit coverage floor on Node's test runner (#1599)

mersahin:
## Summary

Coverage that nothing enforces is a number nobody looks at. This puts a **per-file line-coverage floor** under the `src/lib` modules that already have unit suites, and wires it into the PR gate — so a refactor that quietly deletes a rule *and* the assertions that pinned it can no longer stay green.

Refs #1599 — **deliberately not `Closes`.** See "What is left out" below: the six wave-0/1 modules the issue names cannot get a non-zero floor until #1598 lands, so the issue should stay open for that half.

**No new dependency.** The issue's plan assumed Vitest from #1598, which is still open and unstarted. It turns out the repo does not need it for this: Node 22's own test runner strips TypeScript (`--experimental-strip-types`) *and* measures coverage (`--experimental-test-coverage`), and `scripts/test/*.test.mjs` already imports `src/lib/*.ts` through it. The harness therefore costs zero packages, runs in ~2 s, and the repo does not gain a third test runner.

Preview: https://pr2259.interncrm.com

## Changes

- **`scripts/check-unit-coverage.mjs`** (new) — runs every `scripts/test/*.test.mjs` under Node's coverage, parses the lcov report, and fails when a listed module drops below its floor. Prints a table to the log and to `$GITHUB_STEP_SUMMARY`.
- **`package.json`** — `test:unit` (the suite, no coverage) and `test:unit:coverage` (the suite + the floor).
- **`.github/workflows/ci.yml`** — one step, `unit tests + per-module coverage floor (#1599)`, immediately before `npm run build`, commented in the house style (what incident it prevents).
- **`docs/testing.md`** — a "Unit tests and the coverage floor" section: the unit-vs-e2e split, and the ratchet rule in writing.
- **`releases/unreleased/unit-coverage-gate.json`** — `patch`, changelog only, no `notes` (nothing a user of the app can see changes). See "Release fragment" below for why this reversed.
- No i18n keys and no schema change.

## The numbers as of this PR

Measured on `main` by `npm run test:unit:coverage`. **Lines** are what the gate enforces; branch/funcs are printed because they help read a drop, but gating on branch % is noisy in a way nobody can act on.

| Module | Line % | Floor | Branch % | Funcs % | Why it has a suite |
|---|---:|---:|---:|---:|---|
| `src/lib/mentorDirectory.ts` | 100.00 | **95%** | 100.00 | 100.00 | scan cap / truncated-result reporting (#1820) |
| `src/lib/rateLimitRedis.ts` | 86.67 | **85%** | 64.81 | 85.71 | shared-store transport + its outage fallback (#1696) |
| `src/lib/rateLimitStore.ts` | 98.78 | **95%** | 89.86 | 81.82 | rate limits are shared across replicas, and fail safe (#1696) |
| `src/lib/stageAging.ts` | 100.00 | **95%** | 96.00 | 100.00 | stage visits counted separately from candidates (#1427) |

Floors are the measured value **rounded down to the nearest 5, capped at 95** — green on day one, with a little headroom for an honest refactor, exactly as the issue asks. Both numbers (floor and measured) sit next to each entry in the script, so the next person can see whether the floor moved without running anything.

The six wave-0/1 modules, for the record — all **0.00%** under this runner today:

| Module | Line % | Floor | Note |
|---|---:|---:|---|
| `src/lib/orgContext.ts` | 0.00 | — | wave-0 rewrite (#1549) |
| `src/lib/entitlements.ts` | 0.00 | — | wave-1 billing-subject collapse |
| `src/lib/planGate.ts` | 0.00 | — | wave-1 billing-subject collapse |
| `src/lib/pipeline.ts` | 0.00 | — | assertions exist, but under Playwright — needs #1598 |
| `src/lib/dormantFirstContact.ts` | 0.00 | — | "two mails, never a third" |
| `src/lib/relativeTime.ts` | 0.00 | — | assertions exist, but under Playwright — needs #1598 |

## Review round 2 — loading a module is not coverage

Review found one real design defect in the promotion rule and two robustness gaps. All three were reproduced in the worktree before being fixed, and each fix was re-verified against the same probe. Commit: `f642c8f`.

**1. A bare import minted a permanent, inflated floor (major).** The promotion out of `AWAITING_TESTS` treated *any* lcov row as coverage. But importing a module runs its top-level code, and `pipeline.ts` is almost entirely const stage/label tables — so a test that imports it and asserts nothing scores **85.29% of its lines with 0.00% of its nine functions**, and the gate went red instructing the author to pin an 85% floor that the documented ratchet then forbids ever lowering. Two bad outcomes, both permanent: the gate would certify module *loading* as coverage, and #1592's real pipeline suite would land under a floor it already satisfied before it existed. Worse, the PR that trips it is usually a test for some *other* module that merely imports this one, so its author is handed a failure they cannot act on except by writing a fake floor — precisely the "number nobody can act on" the file's own header says it exists to avoid.

Promotion now requires that something actually **ran**: at least one of the module's functions was called. A module that declares no functions at all (a pure table) is judged on lines, there being nothing else to call. An imported-but-unexercised module reports `loaded, not exercised`, keeps no floor, and **stays in the ratchet** — said out loud on every run, never red, because that is a normal state for a module another test transitively pulls in.

**2. The suggested floor could be `0` (minor).** `Math.floor(lines / 5) * 5` returns 0 below 5%, and this file's own `AWAITING_TESTS` comment says why a floor of 0 is worse than no floor: it can never fail, yet it reads as coverage on the summary table and quietly drops the module off the knowingly-unprotected list. Floors below a **25% minimum** are no longer suggested; the entry stays put with an explanation.

**3. Neither list was checked against the tree (minor).** The two failure modes are silent in opposite directions. Wave-1 is specified to delete `planGate.ts` when it collapses the billing subject — and a deleted `AWAITING_TESTS` entry never reaches the lcov, which is indistinguishable from "no tests yet", so the row would rot forever while `EXPECTED_AWAITING` still looked honest. A *renamed* floored module is the mirror image: it drops to 0% and fails with "restore the assertions this change removed", which is wrong advice for a rename. Both now fail with their own message, before the suite runs.

**4. Two documentation errors (minor).** The new `docs/testing.md` section said "the eight `e2e/*.unit.spec.ts` files" — there are **16** (the number came from #1598, written 2026-08-30; eight more have landed since). And "the first twelve are functional tests" was wrong before this branch too (it said eleven against twelve rows) and this branch incremented it to twelve against thirteen, so the count is replaced with a partition that cannot go stale: "Every row above the two load rows".

Verified after the fixes:

- bare-import probe → **exit 0**, reports `pipeline.ts — 85.29% of its lines ran, but none of its 9 function(s) were called … no floor is suggested and the entry stays put`;
- probe with **real** assertions (`nextOnPathStatus`, `pipelineLabel`, `pipelineOptions`) → **exit 1**, still trips the ratchet: `88.24% line / 40.00% function coverage … (floor 85%)`. The message now carries the function percentage too, so a reviewer can judge whether the suggested floor is honest;
- a renamed/deleted entry in **either** map → exit 1 with its own message;
- the suite itself unchanged: 4 floors satisfied, exit 0.

## Release fragment

`releases/unreleased/unit-coverage-gate.json` (AC #5) **is now included** — this reverses the earlier decision, which was taken on the brief's instruction to skip it. CLAUDE.md does exempt pure CI config, but `check-events.mjs` — the same shape of guard, an architecture check wired into `ci.yml` — carries `webhook-call-sites-on-emit.json`, so the house precedent is that a new red-able gate earns a developer-facing bullet. `bump: patch`, `changelog` only, no `notes`.

## What is left out, and why

The issue's headline ask — a floor on those six — **cannot be honoured today, and a floor of `0` would be a lie on a dashboard.** None of the six is exercised by the coverage-capable runner:

- `orgContext`, `entitlements`, `planGate`, `dormantFirstContact` have no unit tests at all;
- `pipeline` and `relativeTime` *do* have real unit assertions (`e2e/program-templates.unit.spec.ts`, `e2e/locale-dates.unit.spec.ts`) — but they live under Playwright, which measures nothing. That split is precisely what #1598 exists to close.

So the gap is: **the runner where the pure-logic tests live and the runner that can measure coverage are different runners.** Two ways to close it, neither of which belongs in this PR on my own initiative:

1. Land #1598 (Vitest + move the `e2e/*.unit.spec.ts` files into `src/`), then move `pipeline` and `relativeTime` into `FLOORS` with their measured numbers. Adds two dev dependencies.
2. Cheaper and dependency-free, now that Node 22 resolves the blocker #1598 was filed against: port those specs to `scripts/test/*.test.mjs` (swap `@playwright/test` for `node:test` + `node:assert/strict`, and `@/lib/x` for a relative path). **There are 16 of them, not eight** — #1598's count is a 2026-08-30 snapshot and the set keeps growing — so re-cost this before spending on Vitest either way.

Rather than hide that behind six zero-floors, the six sit in an **`AWAITING_TESTS` ratchet** — the same shape as `PENDING_REGISTRATION` in `check-tenant-models.mjs`, which CLAUDE.md already establishes as this repo's pattern for "knowingly not protected yet":

- printed on every run, with a `::warning` annotation so the PR page shows it instead of burying it in a green step, and the verdict line drops the word "OK";
- the size is pinned by an `EXPECTED_AWAITING = 6` literal, so a seventh unfloored module cannot be appended quietly;
- the check **fails** the moment one of them is *exercised* by the node runner, and tells you exactly which floor to write — so the number lands in the same PR as the first test, not months later.

### ⚠️ Script-name collision with #1598

#1598 is specified to define `"test:unit": "vitest run"` and `"test:unit:coverage": "vitest run --coverage"`. **Both keys now exist**, pointing at `node --test` and `scripts/check-unit-coverage.mjs`. Whoever picks #1598 up will hit a name collision the issue does not anticipate, and should also reconsider its premise — Node 22's runner may make Vitest unnecessary altogether. Flagged on #1598.

## The ratchet rule

Written into `docs/testing.md`, and again in the script header:

> **A floor goes up when a module gains tests. It is never lowered to make a PR pass.**

A PR that drops a module below its floor has deleted coverage — that is the whole thing the gate exists to notice. Restore the assertions; do not edit the number.

## Verification

Static checks only (no database and no Docker in this container, so the app and the Playwright suite could not be run):

- [x] `npm run lint` — clean (pre-existing `react-hooks/exhaustive-deps` warnings in `src/app/admin/companies/page.tsx` and `src/app/admin/mentorship/page.tsx`, both untouched)
- [x] `npm run check:i18n` — OK, 4413 keys × 3 locales
- [x] `npm run check:release-fragments` — exit 0; this change ships as `0.166.1-beta` (it also warns that 32 fragments are pending compaction — pre-existing, unrelated to this PR)
- [x] `npm run test:unit` — 49/49 pass, ~1.5 s
- [x] `npm run test:unit:coverage` — exit 0, table above
- [x] `ci.yml` parses, and the new step sits immediately before `npm run build`
- [x] Gate proven red, not just green — every failure mode exercised against a temporary copy: a floor raised above the measured value, an `AWAITING_TESTS` module that gains **real** coverage, an `EXPECTED_AWAITING` mismatch, and (new) a renamed or deleted entry in either map
- [x] Gate proven **not** red where it should not be — a test that only imports a tracked module exits 0
- [ ] `npm run build` — not run locally; CI covers it (nothing under `src/` changed)
- [ ] `npm run test:e2e` — not runnable here; no e2e spec touched

## Contributor terms


mersahin:
## Review response — all six findings actioned, none rejected

Every finding was reproduced in the worktree before being touched. Nothing was dismissed as a false positive. Fixes are in `f642c8f`; the PR body now carries a "Review round 2" section with the details.

**1 — bare import mints an inflated floor (major). Fixed.** Reproduced exactly as reported: a probe test containing only `import { PIPELINE_STATUSES } from '../../src/lib/pipeline.ts'` plus one `assert.ok(...)` turned the gate red at `85.29 | — | 100.00 | 0.00 | NEWLY COVERED` and instructed the author to pin `floor: 85` on a module where **not one of nine functions ran**. This was the real defect — the gate would have certified module *loading* as coverage, under a number the documented ratchet forbids ever lowering, and it would have failed PRs whose only sin was a new test that transitively imports `pipeline.ts`.

Promotion now requires that something actually ran: at least one function called (a module declaring no functions at all is judged on lines, there being nothing else to call). An imported-but-unexercised module reports `loaded, not exercised`, keeps no floor and stays in the ratchet — warned about, never red, since that is a normal state.

Both directions verified:

| probe | before | after |
|---|---|---|
| import only, no behavioural assertions | ❌ exit 1, "pin floor 85%" | ✅ exit 0, "none of its 9 function(s) were called … the entry stays put" |
| real assertions (`nextOnPathStatus`, `pipelineLabel`, `pipelineOptions`) | ❌ exit 1 | ❌ exit 1 — ratchet still fires: `88.24% line / 40.00% function … (floor 85%)` |

The failure message now quotes the **function** percentage alongside the line percentage, so the next reviewer can see at a glance whether a suggested floor is honest.

**2 — suggestable `floor: 0` (minor). Fixed.** Correct, and it contradicted the file's own comment. A `MIN_FLOOR = 25` now applies: below it no floor is suggested and the entry stays in `AWAITING_TESTS` with an explanation, rather than leaving the knowingly-unprotected list for a floor that can never fail.

**3 — no existence check on either list (minor). Fixed.** Both scenarios reproduced. A deleted `AWAITING_TESTS` entry (`planGate.ts`, which wave-1 is specified to delete) exited **0** and printed "no tests yet" forever while `EXPECTED_AWAITING` still looked honest; a renamed floored module failed with "Restore the assertions this change removed", wrong advice for a rename. Both maps are now `existsSync`-checked before the suite runs, with their own message naming the rename/remove remedy and the `EXPECTED_AWAITING` move.

**4 — "eight" unit specs (minor). Fixed.** `ls e2e/*.unit.spec.ts` returns **16**. Taking the advice not to hardcode a number next to a growing glob, the docs now say "(16 of them at the time of writing, and the set keeps growing — count them, don't trust this sentence)". The same wrong count was in the PR body's follow-up estimate; corrected there too.

**5 — "first twelve" (minor). Fixed.** Confirmed 13 functional rows, and confirmed the pre-change text was already wrong (eleven against twelve) — this branch had incremented an incorrect number rather than fixing it. Replaced with a partition that cannot go stale: *"Every row above the two load rows is a functional / correctness test."*

**6 — deviation from the issue (minor). Both suggested actions taken; the deviation itself kept, per your own recommendation.**

- **(a)** Flagged the `test:unit` / `test:unit:coverage` collision on #1598, including the suggestion to re-examine its Vitest premise now that Node 22 strips TypeScript and measures coverage natively. Also cross-referenced in the PR body under "Script-name collision with #1598".
- **(b)** Added `releases/unreleased/unit-coverage-gate.json` (`patch`, changelog only, no `notes`) rather than waiving AC #5. CLAUDE.md does exempt pure CI config, but `check-events.mjs` — the same shape of guard, wired into `ci.yml` the same way — carries `webhook-call-sites-on-emit.json`, so house precedent is that a new red-able gate earns a developer-facing bullet. The earlier "no fragment" note in the body was taken on the brief's instruction; that is now reversed and explained rather than silently dropped.
- The substance of the deviation — no floor on the six named modules, no Vitest, `Refs` not `Closes` — is unchanged, as you recommended keeping it. **#1599 must stay open** for the half that #1598 blocks.

## Static checks after the fixes

| check | result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npm run lint` | clean (two pre-existing `react-hooks/exhaustive-deps` warnings, both untouched) |
| `npm run check:i18n` | OK — 4413 keys × 3 locales |
| `npm run check:release-fragments` | exit 0 — ships as `0.166.1-beta` |
| `npm run test:unit:coverage` | exit 0 — 4 floors satisfied, unchanged |

`npx prisma validate` not run: the schema was not touched. No `src/` file was touched in either commit, so the diff is still confined to the script, the docs, the workflow, `package.json` and the new fragment.

---

PR #2260, opened 2026-09-07
fix: recognise, list and clean up orphan applicant accounts from /apply (#1780)

### LINKED ISSUE #1780: 🌱 🧹 [Admin] Task · Orphan applicant accounts from /apply — recognise them, list them, clean them up
## 🎯 Goal

The public apply link creates a real user account before any human has said yes. When the mentor declines, that account stays forever: it cannot sign in, nobody owns it, and it silently inflates every candidate count and every retention queue. This is a small, well-bounded clean-up with one genuine bug in it.

## 🔍 Current state

- `src/app/api/apply/route.ts:97-111` — `prisma.user.create({ data: { email, password: '!apply-no-login', role: 'MENTEE', … } })`, executed **before** the mentor decides. :117-119 then files the `MentorshipRequest` that the mentor accepts or declines (#1188).
- `src/lib/menteeAccount.ts:20-24` — `NO_LOGIN_PASSWORD = '!created-no-login'`, `IMPORTED_NO_LOGIN_PASSWORD = '!imported-no-login'`, and `NO_LOGIN_PASSWORDS` contains **only those two**. `'!apply-no-login'` is missing, so `isPendingActivation()` (:38) returns `false` for an apply-link account — the one helper whose entire job is "this account has never had a password" does not recognise the largest source of exactly that. 🐛
- `src/app/api/mentorship/[id]/route.ts:11, :100-106` — `isPendingActivation` is what the relation detail page uses to offer "fix the address and send the activation link". An apply-link mentee never gets that offer today.
- `src/lib/mentorshipDecision.ts` — the reject branch decides the request; nothing at all happens to the account behind it.
- `src/app/api/users/[id]/route.ts` + the erase flow (`src/components/UserEraseForm.tsx`, `src/lib/accountErasure.ts`) already implement delete/anonymize with the right GDPR semantics — reuse them, do not write a second deletion path.
- `src/lib/retention.ts` — the existing retention sweep, anchored on `User.consentAt`. An orphan applicant has no `consentAt` to anchor on, which is why they never expire.

## 🛠 What to build

1. **The bug first**: add `'!apply-no-login'` to `NO_LOGIN_PASSWORDS` in `src/lib/menteeAccount.ts:24` (keep the constant exported by name, e.g. `APPLY_NO_LOGIN_PASSWORD`, and have `src/app/api/apply/route.ts:100` import it instead of repeating the literal). Check the other places that compare a password sentinel and make sure none of them regress — grep for `no-login` across `src/` and `scripts/`.
2. Define "orphan applicant" once, in a small helper: `role = 'MENTEE'` **and** the password is a no-login sentinel **and** the account has no `MentorshipRelation` at all **and** its most recent `MentorshipRequest` is `REJECTED` (or there is none) **and** it was created more than N days ago. Do not include accounts a mentor created by hand and is still working on — those are the `'!created-no-login'` rows with an open request.
3. `/admin/candidates` (or a small `/admin/orphan-applicants` section — reuse the existing list component rather than building a new table): a filter that shows exactly these rows, with the count, the applied-on date and which mentor declined. Per-row actions: send the activation link (the flow already exists), or erase — calling the **existing** erase endpoint, never a new delete.
4. A daily housekeeping step in the existing 09:00 cron (`src/services/emailService.ts:3051-3080`, next to `pruneEmailLog`): after a configurable grace period (default 90 days, read through `src/lib/settings.ts`, not a hard-coded literal), anonymize orphan applicants using `src/lib/accountErasure.ts`. Log the count like the other jobs do. Anonymize rather than hard-delete so the funnel counts for closed cycles stay stable.
5. Add a line to `docs/DATA_ACCESS_POLICY.md` or the retention doc stating the new rule in one sentence — a rule that only exists in code is one nobody can answer a DPA question with.

## ✅ Acceptance criteria

- [ ] `isPendingActivation({ password: '!apply-no-login' })` returns `true`, and the relation detail page now offers to activate an apply-link mentee
- [ ] No literal `'!apply-no-login'` string remains outside `src/lib/menteeAccount.ts`
- [ ] An admin can list every orphan applicant, with a count, and act on each one
- [ ] The nightly sweep anonymizes orphan applicants older than the configured grace period and is idempotent
- [ ] The grace period is a setting, not a hard-coded number
- [ ] Anonymization goes through `src/lib/accountErasure.ts` — no second deletion path exists
- [ ] `npm run build` passes
- [ ] An e2e spec: apply → mentor declines → the account appears in the orphan list → erase → it is gone from the candidate count
- [ ] A release fragment is added (`bump: patch`)

## 🇹🇷 Türkçe özet

Genel başvuru bağlantısı, mentor daha kabul etmeden gerçek bir kullanıcı hesabı açıyor. Mentor reddettiğinde o hesap sonsuza kadar kalıyor: giriş yapamaz, sahibi yok, ama bütün aday sayılarını ve saklama kuyruğunu şişiriyor. İçinde gerçek bir hata da var: `menteeAccount.ts:24`'teki `NO_LOGIN_PASSWORDS` listesinde `'!apply-no-login'` yok, yani "bu hesabın hiç şifresi olmadı" diyen yardımcı fonksiyon tam da bu hesapları tanımıyor. Önce onu düzelt, sonra yetim başvuruları listele ve mevcut silme/anonimleştirme akışıyla temizle — yeni bir silme yolu yazma.

## 🤖 Agent prompt

```
Objective: fix the orphan-applicant accumulation from the public apply link in the
Internship CRM. Small, self-contained, includes one real bug fix.

Open first:
  src/lib/menteeAccount.ts:20-44   (NO_LOGIN_PASSWORDS is missing '!apply-no-login' — the bug)
  src/app/api/apply/route.ts:97-119
  src/lib/mentorshipDecision.ts    (the reject branch — nothing happens to the account)
  src/lib/accountErasure.ts + src/components/UserEraseForm.tsx  (REUSE these)
  src/lib/retention.ts             (why an orphan never expires: no consentAt to anchor on)
  src/services/emailService.ts:3051-3080  (the daily 09:00 cron and pruneEmailLog next door)
  src/lib/settings.ts              (how a configurable value is read)

Rules:
  - Do not write a second deletion path. Anonymization goes through accountErasure.ts.
  - Anonymize, do not hard-delete: closed-cycle funnel counts must stay stable.
  - The grace period is a setting, not a literal.
  - `prisma db push`, no migrations folder; this task should need no schema change at all.
  - EN/TR/DE keys for any new admin string.
  - e2e: /admin/candidates renders every candidate TWICE (a md:hidden mobile list
    `candidates-mobile-list` and a desktop grid `candidates-desktop-list`). An unscoped
    name locator there resolves to 2 elements and throws in strict mode — scope to
    getByTestId('candidates-desktop-list') or to a specific card testid.

Done when: build + check:i18n green, an e2e covers apply -> decline -> orphan list -> erase,
the doc line is written, release fragment added, PR open with the preview URL.
```

**Zorluk** · S · **Tahmini** · ~1 gün · **Bağımlılık** · yok

mersahin:
## Summary

The public apply link (`POST /api/apply`) creates a real `User` row before any human has said yes — the applicant needs an account to set a password on, and the mentor decides afterwards through the `MentorshipRequest` the same call files (#1188). When the mentor declines, nothing at all happens to that account. It can never sign in, belongs to nobody, no relation points at it, and it has no `consentAt` for the consent-based retention review (`src/lib/retention.ts`) to anchor on — so it stayed in the user table forever, counting towards every candidate number while holding a name, an address, a phone number and a university with no purpose left to serve. That is a storage-limitation failure (GDPR Art. 5(1)(e) / KVKK m. 4), not untidiness.

This recognises those accounts, lists them where an admin can act on them, and cleans them up on a conservative timer that can never surprise anybody.

Closes #1780

## Changes

**The bug first.** `'!apply-no-login'` was missing from `NO_LOGIN_PASSWORDS` in `src/lib/menteeAccount.ts`, so `isPendingActivation()` — the one helper whose entire job is "this account has never had a password" — did not recognise the *largest* source of exactly that, and the relation detail page never offered to correct an apply-link mentee's address. The literal is now the exported `APPLY_NO_LOGIN_PASSWORD`, `/api/apply` imports it, and `grep -rn "apply-no-login" src/ scripts/` finds it only in `menteeAccount.ts` (plus one assertion in the new e2e spec). The other sentinels (`!created-no-login`, `!imported-no-login`, `!sso-no-login`, `!company-no-login`, `!source-no-login`) were checked: only the first two were ever in that list and both stay in it; nothing else compares a password sentinel.

**One definition, in one place.** `src/lib/orphanApplicant.ts` owns the rule the way `dormantFirstContact.ts` owns dormancy, with the reasoning in its header. An account is an orphan applicant when it is a MENTEE still carrying the apply sentinel, never verified, never signed in, with **no** mentorship relation in either direction, whose **every** request is `REJECTED` (or which filed none), with no consent record — and no other sign of life at all: no tag, CV, document, support ticket, company interest, interview request, project membership, conversation participation or started onboarding. Any single one of those excludes the account. The age clock runs from the **later** of `createdAt` and the decision, so an old account declined yesterday is one day old to this rule. Two deliberate narrowings, both documented in the module: only the **apply** sentinel counts (the issue sketched the rule over any no-login sentinel, which would also sweep up mentees a mentor typed in by hand and is still working on), and messages cannot be a sign of life because `Message.senderId` has no FK to `User` — noted rather than silently omitted.

**The rescue is part of the rule (clause 7).** An account with an **outstanding** password link — unused and unexpired — is excluded from the *sweep* (not from the listing) for as long as that link lives. Without it the panel's advertised "send set-password link" would be a mail with no effect: nothing on the `User` row changes when it is sent, so the same night's run would anonymise the account **and** delete the link it had just been mailed, and the applicant would click a dead link into an erased account. It postpones rather than cancels, deliberately — a link nobody clicks must not keep an abandoned account alive forever, and re-sending is one click. `daysUntilAnonymize` counts down to whichever of the two gates clears last, so the date on the page is a date on which something actually happens, and the row carries a caption saying why. (`/apply` issues its own 7-day `SET_INITIAL` link to every account it mints, so a fresh applicant is held for its first week — a fortnight dead by the time a 90-day gate opens, so the hold only ever decides the answer when a human sent a new one.)

**Recognition.** `/api/candidates` marks each returned row with `orphan` (one extra query per rendered page — never a second copy of the rule on the client) and accepts `?orphan=1` to narrow the list to exactly them. `/admin/candidates` shows the badge on both the mobile and desktop cards and gains a filter toggle next to the Active/Archive tabs. The sentinel fix also means an apply-link account now resolves to the `no_login` account state everywhere `accountState()` is read.

**Listing / the dry run.** `/admin/retention` — the page that already answers "whose data are we still holding?" — gains an orphan-applicant section: every orphan with the applied-on date, **who decided the request** and when (`MentorshipRequest.decidedBy`, falling back to the preferred mentor only when that account is gone), the days left before automatic anonymisation, the total count, and how many the next nightly run would take (counted in the database, so a truncated list cannot under-report it — and reported against the per-run cap, so a backlog is never announced as "all of them tonight"). `GET /api/admin/orphan-applicants` is read-only and shares `listOrphanApplicants()` with the sweep, so what the page shows and what the job takes cannot drift apart; it reads the grace period from the **global** settings layer for the same reason, since that is the only layer the sweep can see. A failed fetch says so, with a retry — the one page whose job is to warn before irreversible deletion must never render "nothing to clean up" when it knows nothing. Per-row actions reuse existing endpoints only: `UserEraseForm` (name + the acting admin's own password) against `POST /api/admin/users/[id]/erase`, and `POST /api/admin/users/[id]/reset-password` to send a set-password link when the decline was a mistake.

**Clean-up.** One `registerRetention()` entry on #1678's registry, not a new cron (03:20 UTC, the one retention schedule). It **anonymizes** through `accountErasure.anonymizeUser` — there is no second deletion path — rather than hard-deleting, so the declined request and the row survive and closed-cycle funnel counts stay stable. The grace period is the new `orphanApplicantGraceDays` setting (default 90, writable through `PUT /api/admin/settings`, same fallback contract as the other windows: 0/negative/unparseable falls back to the default rather than erasing everything). It is an installation-wide window read from the global layer — the sweep runs with no tenant bound and crosses every org in one pass, so a per-tenant override would be written and never read; that is stated in the entry, in `SETTING_DEFAULTS`, in the settings route and in the doc. Capped at 200 accounts per run because `anonymizeUser` is a transaction per account, not a `deleteMany` per 500 ids; reports `capped` while catching up, and the admin page says so when a backlog exceeds one run. Idempotent by construction: anonymising rewrites the address into the `@erased.local` namespace the rule excludes. It leaves its own `retention.orphanApplicants` audit row on top of the shared `retention.pruned` line, because this is the only scheduled job in the product that erases a person's data and "which night did that account go?" has to be answerable on its own.

**What an orphan may already be attached to — everything checked.** `MentorshipRequest`: kept, and its `message` scrubbed by the existing `scrubFreeTextOps`; it is the row that records the application ever happened, so hard-deleting the user (which cascades it) is left to the admin's explicit choice, and the automatic path never does. `UserConsent`: an account that has one is **not** an orphan at all (it belongs to the consent-based review instead), and `anonymizeUser` revokes rather than deletes. `InvitationToken`: keyed by e-mail with no FK to the applicant, and an apply account is never invited — untouched either way, and an invitation for the address would still work. `PasswordResetToken`: every apply account is issued a 7-day `SET_INITIAL` link, and a hard delete cascaded those away while an anonymise left them live — `/api/auth/reset` checks only the token, not the state of the user behind it — so `anonymizeUser` now deletes them (one line, in the same transaction), and clause 7 above makes sure the sweep can never reach an account whose link is still live. `EmailLog`/`NewsletterSend`: already handled by `forgetEmailLog`. `Notification` (the mentor's "new application" row) and `ActivityLog`/`AuditLog`: keyed to the mentor or the actor, not to the applicant, and left alone. `MentorshipRelation`, `InteractionLog`, `Message`, `CvFile`, `Document`, `SupportTicket`, `Tag`, `CompanyInterest`, `InterviewRequest`, `ProjectMember`, `ConversationParticipant`, `MenteeOnboarding`: an account holding any of them is excluded from the rule in the first place.

**Docs.** `docs/pii-access-lifecycle.md` — the retention doc — gets the rule in one sentence, its three non-negotiable boundaries (anonymise not delete, dry run first, any sign of life wins), what the rescue does and how far the per-run cap reaches, a row in the windows table, and why these accounts were invisible to the existing review.

**i18n.** EN/TR/DE for the new `orphanApplicants` namespace and the three `candidates.*` strings; `npm run check:i18n` green at 4437 keys × 3 locales.

## Verification

- [x] `npm run build` passes
- [x] `npm run check:i18n`, `npm run check:release-fragments`, `npm run check:tenant-models` pass (the fragment checker's "compaction looks stuck" warning is pre-existing and unrelated)
- [x] Added `e2e/orphan-applicants.spec.ts`: apply → still PENDING, so **not** listed (asserted against the *loaded* panel, not the skeleton) → the request is declined → listed on `/admin/retention` with the decider and no "due now" label → badge on `/admin/candidates`, and `?orphan=1` keeps the orphan while an ordinary seeded mentee disappears (scoped to `candidates-desktop-list`, since every candidate renders twice) → clause 7 asserted against the rule itself: due → live link → not due → link spent → due again → erased through the existing gates → gone from the candidate list. Not tagged `@smoke` — it is a five-page admin path, too slow for the PR gate.
- [ ] Not executed here: this container has no database and no Docker daemon, so the Playwright suite could not be run. Verified statically and by a full production build.
- No schema change (`prisma/schema.prisma` untouched), so no `db push` concern.

Preview: https://pr2260.interncrm.com

## Contributor terms


mersahin:
Review follow-up pushed as `9251838`. I verified all seven findings against the code first; **all seven were real and all seven are fixed** — nothing rejected as a false positive.

### The two blockers

**1. The rescue did not stop the sweep (`src/lib/orphanApplicant.ts`).** Confirmed exactly as described: `POST /api/admin/users/<id>/reset-password` changes nothing on the `User` row, `PasswordResetToken` was not in the signs-of-life list, and this branch's new `prisma.passwordResetToken.deleteMany` in `anonymizeUser` meant the sweep would erase the account *and* the link it had just mailed, in one transaction.

An outstanding (unused, unexpired) link is now **clause 7** of the documented rule, on the age-gated branch only, using the model's real `used: Boolean` field:

```ts
passwordResetTokens: { none: { used: false, expiresAt: { gt: new Date() } } }
```

Two things beyond the suggested one-liner, because the one-liner alone would have left the page lying about the date:

- `listOrphanApplicants` reads the same predicate back, so `daysUntilAnonymize` is now `max(graceRemaining, linkRemaining)` — the sweep needs *both* gates open, and a countdown that ignored the hold would print a day on which nothing happens. The row carries a caption saying why the number moved (`orphan-hold-<id>`), and the panel reloads after a successful send so the reprieve is visible rather than taken on trust.
- Worth knowing for anyone reading the rule: **`/apply` itself mints a 7-day `SET_INITIAL` link for every account it creates**, so a fresh applicant is held for its first week. That costs nothing — the link is a fortnight dead by the time a 90-day age gate opens — but it is why the e2e ages that token out before asserting anything about the rescue.

It postpones rather than cancels, deliberately: a link nobody clicks must not keep an abandoned account alive forever, and re-sending is one click.

**2. `orphanApplicantGraceDays` missing from the settings PUT schema.** Confirmed — `z.object` with no `.strict()` strips it, the handler's `Object.entries(parsed.data)` then sees nothing, and the call returns 200 having written no row. Added next to the other four windows with the same `^\d{1,4}$` bound.

### The five smaller ones

- **Grace resolved differently by page and sweep.** Real, and latent only because of finding 2. Fixed in the direction the finding's first option suggests: the route now reads `getSetting(KEY, null)` — the global layer, which is the only layer `runRetentionPrune` can see, since it calls `getSettings()` with no org bound and sweeps every tenant in one pass. A per-tenant override would be written and never read, so that is now stated in the entry, in `SETTING_DEFAULTS`, in the settings route and in the doc table rather than left as a difference between two files.
- **"{n} will be anonymised by the next nightly run" ignored the 200/run cap.** Real, and wrong by 4× in exactly the first-run-with-a-backlog case. `ORPHAN_ANONYMIZE_PER_RUN` moved into `orphanApplicant.ts` next to the rest of the rule (the dry run has to be able to say the same number), the route reports it as `perRun`, and a backlog gets its own string in all three languages: "{n} are past the grace period. One nightly run anonymises at most {cap} of them, oldest first…".
- **"Declined by" named the preferred mentor.** Real, both halves. Now selects `decidedBy: { select: { fullName: true } }` and prefers it, with `preferredMentor` as the fallback for when the decider's account is gone (`decidedById` is nulled by the erasure path, `preferredMentorId` by `SetNull`). And the cell can no longer contradict itself: `noDecision` renders only when `declinedAt` is also null, otherwise a new `unknownDecider` string ("Declined — decider no longer on the system" / "Reddedildi — kararı veren artık sistemde değil" / "Abgelehnt — entscheidende Person nicht mehr im System").
- **Failed fetch rendered the empty state.** Real. `failed` is tracked apart from `data === null`, with its own branch (`orphan-applicants-error`), a message that says the page cannot tell what the run would take *and* that opening it anonymised nothing, and a retry button.
- **Two vacuous e2e assertions.** Real, both. The pending check now waits for the loaded state (`orphan-applicants-empty` **or** the first `orphan-row-*`) before the `toHaveCount(0)`; the filter check seeds an ordinary MENTEE and asserts its card **disappears** from `candidates-desktop-list` once `?orphan=1` is on. The decline assertion also flipped from "Orphan Mentor" to "Orphan Admin", which is the point of the `decidedBy` fix. Added a step that asserts clause 7 against the rule itself (`orphanApplicantWhere(cutoff)` + a count) — due → live link → not due → link spent → due again.

### Checks

`npx tsc --noEmit` clean, `npm run lint` clean (only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`), `npm run check:i18n` OK (4437 keys × 3 locales), `npm run check:release-fragments` exit 0 — its "compaction looks stuck: 32 fragments pending" warning is pre-existing and unrelated to this PR. Schema untouched, so no `prisma validate` run. The release fragment's prose was updated to match what the diff now does (the rescue hold, the writable knob and its global layer, the cap wording, `decidedBy`, the error state).

---

mersahin:
## Merged `main` in — conflicts resolved

Four sibling PRs from the same batch landed while this was in review, so the branch went stale. `git merge origin/main` produced **three conflicts, and in all three both sides were purely additive** — nothing had to be given up.

| File | The two sides | What I did |
|---|---|---|
| `src/lib/settings.ts` | #1646's `notificationRetentionDays` vs this PR's `orphanApplicantGraceDays` in `SETTING_DEFAULTS` | Kept both. `notificationRetentionDays` goes with the four telemetry windows it belongs to; `orphanApplicantGraceDays` last, since it is the one entry that governs **a person's account** rather than a telemetry table — the same ordering argument the entry's own comment makes. |
| `src/app/api/admin/settings/route.ts` | The same two keys in the zod schema | Kept both, same order. Both are `\d{1,4}`, so #1646's "unlike the four above" comment still reads correctly with this one after it. |
| `docs/pii-access-lifecycle.md` | Both sides add a row to the retention-windows table; main *also* adds the whole "Bildirim penceresinin iki freni ve tek istisnası (#1646)" subsection right there | Kept both rows and the whole subsection. |

**One thing that needed more than marker-cleanup.** This PR's table row ended with *"Ayrıntısı hemen aşağıda"* ("detail immediately below"), which was already loose — the #1780 subsection is at the end of the file — and became plainly wrong once #1646 dropped a long subsection between the table and it. I repointed it at the heading by name: *"Ayrıntısı bu bölümün sonundaki «Sahipsiz başvuru hesapları (#1780)» başlığında"*. No claim changed; the merge just invalidated a cross-reference and it needed fixing rather than shipping as a broken pointer.

**Auto-merged but hand-verified**, because a silently dropped list entry is exactly the failure a clean auto-merge hides and the type checker does not catch:
- `src/lib/retentionEntries.ts` — both `orphanApplicant` and #1646's `notification` entries are present and registered (7 keys total).
- `src/i18n/dictionaries.ts` — all three locale blocks took both sides; `check:i18n` confirms parity.

Nothing else conflicted: this PR does not touch `package.json`, `infra/deploy-prod.sh`, the schema, or `docs/agent-experience.md`'s contested spot. The release fragment still describes exactly what the branch does, so its prose is unchanged.

### Checks after the merge

| Check | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npm run lint` | clean — only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship` |
| `npm run check:i18n` | OK — **4444** keys × 3 locales (up from 4437; main's new keys) |
| `npm run check:release-fragments` | exit 0 (the "compaction looks stuck" line is the pre-existing repo-wide warning, and the checker itself says "Not a problem with this PR") |
| `npm run check:tenant-models` | no new drift, the 8 `#1559` pending models unchanged |
| `npm run test:unit:coverage` | green — main's new #1599 coverage gate, no floor breached |
| `npx prisma validate` / `db push` | n/a — this PR still touches no schema |

Playwright is unchanged and still not run here (no database in this container), same caveat as the original description.

---

PR #2261, opened 2026-09-07
fix: stop recording no-op stage-history rows (#934)

### LINKED ISSUE #934: 🐛🌱 [Pipeline] Task · Aşama geçmişinde "100 · İlk temas → 100 · İlk temas" gibi no-op kayıtlar oluşuyor
Story: #894 · Epic: #861

## Amaç (stajyer-dostu 🌱)

Aşama geçmişi, gerçek ilerlemeyi anlatan bir zaman çizgisi olmalı. Bugün aynı aşamadan **aynı aşamaya** "geçiş" kayıtları oluşuyor ve geçmiş gürültüye dönüşüyor.

## Tekrar üretim (yerel)

1. `npm install && npx prisma generate && npx prisma db seed && npm run seed:demo && npm run dev`
2. `mentor.aylin@demo.example.com` / `DemoPass123!` → `/mentor/mentees` → Deniz Demo → **View details**
3. "Stage history (1)" bölümünde: **`100 · İlk temas → 100 · İlk temas`** (Aylin Demo · tarih)

## Mevcut durum (kod kanıtı)

- `src/app/api/status-changes/route.ts:35` → `prisma.statusChange.create({ … })` — kaynak ve hedef aşamanın **eşit olup olmadığı kontrol edilmiyor**.
- Diğer yazma yolu: `src/app/api/mentorship/[id]/route.ts:155-164` — burada `if` bloğu var (bildirim yalnızca değişimde atılıyor), ama `StatusChange` yazımının aynı korumaya sahip olduğu **doğrulanmalı**.
- Not: bu proje `pipelineStatus`'ü artık **String** olarak tutuyor (özel org aşamaları için — `docs/agent-experience.md`, #747). Karşılaştırma string bazlı olacak, enum varsayma.

## Ne yapılacak

1. Sunucu tarafında: `from === to` ise `StatusChange` **yazma**. Davranış seçimi (200 + "değişiklik yok" mu, 400 mü) PR'da gerekçelendirilir — mevcut istemcileri kırmamak önemli.
2. Gösterimde mevcut no-op kayıtları filtrele (eski veri geriye dönük temizlenmeyecek).
3. Aynı korumayı ikinci yazma yolunda da doğrula (#926 bu iki yolu birleştirmeyi öneriyor — çakışma olursa o issue ile koordine et).

## Kabul kriterleri

- [ ] Aynı aşamaya "geçiş" isteği yeni bir geçmiş kaydı **üretmiyor**.
- [ ] Gerçek aşama değişimleri normal kaydediliyor (regresyon yok).
- [ ] Arayüzde mevcut no-op kayıtlar görünmüyor.
- [ ] İstemci tarafı hata yönetimi bozulmuyor (seçilen yanıt koduna göre UI doğru davranıyor).
- [ ] Mevcut `e2e/admin-status.spec.ts` / pipeline spec'leri yeşil.
- [ ] `npm run build` yeşil.

## Hazır AI prompt

```
Aynı aşamaya no-op geçiş kaydını engelleyeceğim.
1) src/app/api/status-changes/route.ts'i tamamen oku. Gövde neyi kabul ediyor, mevcut aşama
   nereden okunuyor? Karşılaştırma yapılıyor mu?
2) src/app/api/mentorship/[id]/route.ts:140-170'i oku. Orada StatusChange yazımı korumalı mı?
3) İstemci tarafını bul (grep -rn "status-changes" src/app src/components) — 400 dönmek UI'ı
   kırar mı? Hangi yanıt daha güvenli?
4) Gösterim tarafında no-op kayıtları nerede filtrelemeliyim (sorguda mı, render'da mı)?
Dikkat: pipelineStatus artık String (özel aşamalar var) — enum karşılaştırması yazma.
Sonra diff'i ver.
```

**Zorluk** · Kolay 🌱 · **Tahmini** · 2-3 saat · İlgili: #926 (iki yazma yolunun birleştirilmesi)

mersahin:
## Summary

A `StatusChange` whose `fromStatus` equals its `toStatus` says "this person moved from stage X to stage X". Nobody moved. It is noise in the one table that is supposed to be a clean record of a person's progress — and, less visibly, it is arithmetic.

Closes #934

Preview: https://pr2261.interncrm.com

> **Revised after review.** The first version of this PR also shipped `prisma/backfill-noop-status-changes.mjs`, an unattended mass `DELETE` of the existing rows wired into `infra/deploy-prod.sh` and `infra/server/topic-deploy.sh`. That has been **removed** — see "What is *not* here" below. Everything else is unchanged.

### One gate, not N guards

The from → to check already existed at two of the write paths (#1338), written out inline at each. Every path that can create a `StatusChange` now builds the row through a single helper, `statusChangeData()` in `src/lib/stageChange.ts`, which returns `null` when the row would be a no-op:

- `POST /api/status-changes` — the admin's manual history entry (had a guard)
- `PUT /api/mentorship/[id]` — board drag/drop and the candidate-detail stage select (had a guard)
- `POST /api/admin/candidates/bulk` `advanceStage` — had no guard written out, though it could not actually reach the case: `nextOnPathStatus()` returns the *next* element of a hardcoded, duplicate-free list and never consults a tenant's own stage keys (those give `idx < 0` and are dropped a line earlier by `if (!nextStatus) continue`, #740). The call is **defence-in-depth over an input that is duplicate-free by construction**, so this path cannot drift from the other two if the source of `nextStatus` ever changes — not a live defect that was closed. The comment there and the new spec both say so explicitly.
- `prisma/seed-demo.mjs` — **wrote the reported row itself.** It seeded `APPLICATION_100 → <current stage>` for every demo mentee, including the ones still on `APPLICATION_100`. That is exactly the reproduction in the issue: Deniz Demo, "Stage history (1) · `100 · İlk temas → 100 · İlk temas`".
- `prisma/backfill-relation-start-stage.mjs` — cannot produce a no-op by construction, but a backfill running unattended on production is the last place to rely on "cannot happen", so it now skips explicitly. (`src/lib/mergeUsers.ts` only reassigns the FK of existing rows; `from`/`to` are untouched.)

The helper returns the `data` payload rather than performing the write, so it still works from inside the `$transaction([...])` arrays two of these paths use — the audit row and the relation's `pipelineStatus` continue to land together or not at all. It also absorbs the `reasonNote?.trim() || null` normalisation that was copied at each call site.

### The response: silent no-op, not 400

Setting a stage to its current value is not an error from the caller's point of view. An admin who re-picks the current option in a `<select>` has not made a mistake, and the clients here fire the request on every change event and treat a non-2xx as a failure to surface — a 400 would make correct behaviour look broken. `POST /api/status-changes` keeps returning `200 { change: null, changed: false }`, `PUT` keeps returning `200`, and bulk advance simply does not count the relation in `updated`.

### What is *not* here: no rows are deleted

#934 decides this in as many words — *"Gösterimde mevcut no-op kayıtları filtrele (eski veri geriye dönük temizlenmeyecek)"*: filter them in the display, existing data will **not** be cleaned retroactively. The display filters already exist from #1338 (`isStageTransition` in `/api/mentorship/[id]`, `/api/users/[id]` and `src/lib/relationTimeline.ts`), so the issue's acceptance criteria are met without touching a row.

Three reasons the deletion is gone rather than narrowed:

1. **It can destroy an admin's own words.** `StatusChange.reasonCode` / `reasonNote` landed with the drop-off-reason work (#810); the no-op gate at `POST /api/status-changes` landed later (#1338). In that window a backdated entry with `fromStatus === toStatus === INTERNSHIP_DROPPED_460` was accepted *and* `validateDropoffReason()` required a whitelisted `reasonCode` — plus, for `OTHER`, non-empty free text. Such a row matches `fromStatus === toStatus` and nothing in the predicate looked at either column.
2. **Nothing would have survived it.** The deploy line ended in `|| true`, and `--apply` short-circuits the dry-run listing, so the rows would have gone unlogged and unrecoverable on the first prod deploy.
3. **It silently widened another unattended backfill.** `prisma/backfill-relation-start-stage.mjs` deliberately skips any relation that has a `StatusChange` row ("a relation that was ever placed deliberately … must not be dragged back to stage 1"). Deleting a relation's only row would have made it match `statusChanges: { none: {} }` on the *next* deploy and rewritten its `pipelineStatus`, one deploy after the change that caused it.

The arithmetic those rows still distort is real and now tracked separately as **#2264**, to be fixed at the readers (`stageEnteredAt()` / `computeStageAging()`), not by destroying audit rows. The gate's header comment in `src/lib/stageChange.ts` points at it.

### The quieter bug this prevents going forward

Filtering at render time was never enough for *new* rows either, because the numbers are computed from the **table**, not from the list the UI renders:

- `stageEnteredAt()` (`src/lib/stageClock.ts`) takes the newest `StatusChange` whatever it says, so a no-op row **restarts the "days in stage" clock** — the mentor board chip, `GET /api/mentorship`, the mentee portal and the admin aging report all under-report, and an overdue relation can quietly stop reading as overdue.
- `computeStageAging()` (`src/lib/stageAging.ts`) reads the row as "left stage X, entered stage X" and counts **two visits where there was one**, with a near-zero dwell dragging that stage's average and median down.
- `GET /api/mentor/analytics` and `src/lib/activityReport.ts` count `statusChanges` rows directly as the "stage moves" signal.

No path can add to that set any more; #2264 removes the effect of the ones already there.

## Changes

- `src/lib/stageChange.ts` — `statusChangeData()`, the single gate, with the reasoning above written down next to it.
- `src/app/api/status-changes/route.ts`, `src/app/api/mentorship/[id]/route.ts`, `src/app/api/admin/candidates/bulk/route.ts` — all three build their row through it.
- `prisma/seed-demo.mjs` — no history row for a mentee who has not moved off the first stage.
- `prisma/backfill-relation-start-stage.mjs` — explicit skip rather than an implicit guarantee.
- `e2e/status-history-noop.spec.ts` — new spec.
- `releases/unreleased/no-noop-stage-history-rows.json` — `patch`, with EN/TR/DE notes.

## Verification

- [x] `npm run check:i18n` (4413 keys × 3 locales) and `npm run check:release-fragments` green
- [x] Added an e2e test — `e2e/status-history-noop.spec.ts` sets a relation to its current stage through all three write paths and asserts each returns 2xx, writes **no** `StatusChange` row and sends the mentee no notification, then that a real bulk advance still records exactly one row. Its header states which steps reach the new gate (the first two) and which exercise the pre-existing `if (!nextStatus) continue` from #740 (step 3), so the spec does not claim coverage it does not provide. `e2e/admin-status.spec.ts` keeps its existing coverage of the two API-shaped paths.
- [ ] Not run here: this container has no database and no Docker daemon, so the Playwright suite was not executed — verified by static checks and by CI.

Note, pre-existing on `main` and untouched by this PR: `npm run check:stage-keys` fails on `src/lib/stageClock.ts:84 HIRED_660` (it fails identically with this branch stashed).

## Contributor terms


mersahin:
## Review findings applied (0f03fc6)

**1. Major — the mass `DELETE`: agreed, removed entirely.**

I verified the issue text before touching anything, and it is unambiguous: #934 step 2 reads *"Gösterimde mevcut no-op kayıtları filtrele (eski veri geriye dönük temizlenmeyecek)"*, and the acceptance criterion is *"Arayüzde mevcut no-op kayıtlar görünmüyor"* — the UI, not the table. The display filters that satisfy that already exist from #1338 (`isStageTransition` in `/api/mentorship/[id]:117`, `/api/users/[id]:101`, `src/lib/relationTimeline.ts:315`), so the deletion was buying nothing the issue asked for.

The data-loss path checks out too. `reasonCode`/`reasonNote` are on `StatusChange` (`prisma/schema.prisma`) and `validateDropoffReason()` *requires* a whitelisted code — plus non-empty free text for `OTHER` — for a move into an off-path stage, while `POST /api/status-changes` only gained its `fromStatus === toStatus` refusal in #1338. A backdated `INTERNSHIP_DROPPED_460 → INTERNSHIP_DROPPED_460` row with an admin's note on it is therefore possible, matches the predicate on line 87, and was never read by it. With `--apply` skipping the dry-run listing and `|| true` swallowing the outcome, the note would have been gone on the first prod deploy with nothing written down.

Rather than narrow the predicate, I took the other option: `prisma/backfill-noop-status-changes.mjs` is deleted and both deploy hooks (`infra/deploy-prod.sh`, `infra/server/topic-deploy.sh`) are reverted to their `origin/main` state — `git diff origin/main -- infra/` is now empty. The arithmetic that motivated it is real, so I filed **#2264** to fix it at the readers (`stageEnteredAt()` in `stageClock.ts`, `computeStageAging()` in `stageAging.ts`, and the `take: 1` + `select: { createdAt: true }` query in `/api/mentorship/route.ts` that cannot filter as written), with "no rows are deleted" as an acceptance criterion. `src/lib/stageChange.ts`'s header comment now points there.

**2. Minor — the unreachable bulk guard: agreed, all three claims corrected.**

Confirmed against `src/lib/pipeline.ts:46-50`: `ON_PATH_STATUSES` is a hardcoded `as const` array with no duplicates, `nextOnPathStatus` returns `ON_PATH_STATUSES[idx + 1]`, and it never consults a tenant's `PipelineStage` rows — a custom key gives `idx < 0` and is dropped by `if (!nextStatus) continue`. So the branch is unreachable and the old comment's "a tenant pipeline carrying a duplicate key would have written the no-op row unchallenged" was simply wrong. Fixed in three places:

- the comment in `bulk/route.ts` now says the call is defence-in-depth over an input that is duplicate-free by construction;
- the release fragment drops "neither of which had the check" (the demo-seeder half of that claim was true and is kept);
- `e2e/status-history-noop.spec.ts` states in its header, and again at step 3, that only the first two steps reach the new gate and that step 3 exercises the pre-existing `if (!nextStatus) continue` from #740.

I did **not** add the suggested unit test for `statusChangeData()`. `src/lib/stageChange.ts` imports `./pipelineStages` → `./prisma`, which constructs a `PrismaClient` at module load, so it cannot be imported by the `node --test --experimental-strip-types` harness in `scripts/test/` the way `stageAging.ts` (deliberately dependency-free) can. Making it testable means splitting the pure gate out of the module this PR just consolidated *into*, which seemed the wrong trade for a one-line predicate that the e2e spec covers end-to-end on two real write paths. Happy to do it if you'd rather have the unit test.

**3. Minor — the coupling with `backfill-relation-start-stage.mjs`: agreed, and it disappears with (1).**

The mechanism is real — that script selects `statusChanges: { none: {} }` precisely to skip relations that were ever placed deliberately, and deleting a relation's only row would have made it match on the *following* deploy. With the deletion gone there is no coupling left to document, so I reverted `infra/deploy-prod.sh` rather than adding a comment about an interaction that no longer exists. The defensive `rel.pipelineStatus === target` skip added to that script stays: it is unrelated to the deletion and is a real (if unreachable today) guard.

Static checks after the changes: `npx tsc --noEmit` clean, `npm run lint` exit 0 (only the two pre-existing `react-hooks/exhaustive-deps` warnings on `admin/companies` and `admin/mentorship`), `npm run check:i18n` OK (4413 keys × 3 locales), `npm run check:release-fragments` exit 0. The schema is untouched, so no `prisma validate`. The fragment's user-facing notes were rewritten: they no longer claim existing entries were cleaned up or that the stage clock is retroactively accurate, only that a same-stage write no longer writes an entry or restarts the counter — which is what this PR actually does.

---

Issue #1252, opened 2026-08-19
Role conversion: offer it on the admin profile pages and inform the converted person


mersahin:
Follow-up to #1243/#1245 (maintainer request). Two gaps in the shipped conversion:

1. **It only lives on `/admin/users`.** The natural place an admin decides "this mentee should mentor now" is the person's profile — `/admin/candidates/[id]` and `/admin/mentors/[id]`. Extract the button + confirm panel into a shared `RoleConvertButton` and mount it on both detail headers (and reuse it on the users list instead of the duplicated inline panel).

2. **The converted person hears nothing.** The conversion revokes every session they have — they are silently signed out mid-work and land in a different shell at next sign-in. They should be told what happened and why:
   - an in-app notification (waiting after the forced re-login), linking to their new home shell;
   - an email (EN/TR/DE via `preferredLanguage`), sent unconditionally — an account-level change that signs you out everywhere is a transactional/security notice, not an opt-out-able digest.

PR #2269, opened 2026-09-07
fix: mentor profile message shortcut, and stage labels that stopped translating

### LINKED ISSUE #2268: Admin mentor profile has no message shortcut, and pipeline stage labels render in English in a Turkish UI
Two small defects on adjacent admin surfaces, reported together.

## 1. `/admin/mentors/[id]` has no "send message" button

`src/lib/personHref.ts` makes `/admin/mentors/<id>` the canonical admin-side mentor profile — it is where every `PersonHoverCard` "open profile" for a `MENTOR` lands. But that page never imported `UserQuickActions`, so it is the one person page in the product with no way to contact the person you are reading about. The affordance was added to `/admin/candidates/[id]` for #51 and to the hover cards for #1166; the mentor page (EPIC B / #415) was skipped.

Nothing in the authorization layer or the dictionaries blocks it:

- `canMessage()` in `src/lib/conversations.ts` short-circuits to `true` when either side is an `ADMIN`, before the `sharesProject` / `hasMentorship` checks, so ADMIN to MENTOR messaging is already supported and `POST /api/conversations` already creates the DIRECT conversation.
- `/api/admin/impersonate` refuses only `ADMIN` targets, and `/admin/users` already offers `loginAs` on mentor rows.
- Every string the component uses (`messages.sendMessage`, `usersAdmin.loginAs`, ...) already exists in EN/TR/DE.

So it is a pure UI omission.

**Not in scope:** `/admin/mentor-applications/[id]` should NOT get the affordance — a `MentorApplication` has no `userId` and no applicant account until it is approved, so there is no conversation to open. Contacting an applicant is e-mail's job.

## 2. Pipeline stage labels render in English in a Turkish UI

On `/admin/candidates/[id]` the stage `<Select>` shows `100 · First contact` while every label around it is Turkish.

Chain:

1. `resolvePipelineStages(orgId, locale: Locale = 'en')` has a silently-English default (`src/lib/pipelineStages.ts`).
2. `GET /api/admin/organizations/[id]/pipeline-stages` calls it **with no locale**, so it answers with the 13 built-in labels in English.
3. The editor (`src/app/admin/organizations/[id]/pipeline/page.tsx`) treats that response as editable state, prefilling 13 text inputs, and Save PUTs the whole prefilled set straight back.
4. The PUT persists every label unconditionally (`label: s.label.trim()`).

One Save on an untouched editor therefore converts "13 localized built-in stages" into "13 hardcoded English rows" — after which `useResolvedStages()` correctly renders them as-is in all three languages, because the label column is a single string with no per-locale variant. Recolouring or reordering a stage is enough to trigger it.

`DELETE` ("Reset to defaults") has the same omission, so the reset repopulated the editor in English too.

One further genuine leak on the same root cause: `GET /api/admin/stage-sla` ships `label` to `StageSlaEditor`, which prints it verbatim, so the whole stage-SLA settings list read English for a default-stage org.

### Known remaining leaks (separate issues, not this one)

- `src/app/api/calendar/feed/[token]/route.ts` calls `pipelineLabel(...)` with no locale **and** bypasses custom stages entirely, so a renamed stage shows its built-in name in a subscribed calendar. The token resolves to a user, so the locale is available.
- The weekly analytics report e-mail (`src/services/emailService.ts`) interpolates the raw enum key (`APPLICATION_100`) into the HTML table and ignores custom stages. Localizing it means moving the row construction inside the per-admin loop.
- `src/app/admin/organizations/[id]/pipeline/page.tsx` is 100% hardcoded English.
- Per-locale custom labels (`PipelineStage.labels Json` as `{ en?, tr?, de? }`, the way `EvaluationCriterion.labels` already works) would let a tenant rename a stage in more than one language. Only needed by a tenant that renames stages AND has readers in several languages.

mersahin:
Closes #2268.

Two small defects on adjacent admin surfaces, one PR.

## Defect 1 — the admin mentor profile had no "send message" button

**Root cause: a pure UI omission.** `src/app/admin/mentors/[id]/page.tsx` never imported `UserQuickActions`. The component went onto the candidate profile for #51 and into the hover cards for #1166; the mentor detail page (EPIC B / #415, later given `RoleConvertButton` for #1252) was skipped — even though `src/lib/personHref.ts` makes `/admin/mentors/<id>` the canonical admin-side mentor profile, so it is where every `PersonHoverCard` "open profile" for a `MENTOR` lands.

Nothing anywhere else blocked it. `canMessage()` short-circuits to `true` when either side is an `ADMIN`, `POST /api/conversations` creates the DIRECT conversation, `/api/admin/impersonate` refuses only `ADMIN` targets, `/admin/users` already offers `loginAs` on mentor rows, and every string the component needs already exists in EN/TR/DE. **No i18n, API or authorization change was needed.**

**What changed:** the full `<UserQuickActions userId={user.id} role={user.role} />`, in the candidate page's position (badge → quick actions → convert).

The whole component rather than a message-only variant, deliberately:

- its public API is `{ userId, role?, className? }` — "message only" would mean widening a shared component's props for one caller;
- the impersonate button is already self-gating (`session?.user?.role === 'ADMIN'`, plus `if (isSelf) return null`), so on an admin-only route it renders exactly what the candidate page renders;
- the server already sanctions impersonating a mentor, and `roleHome('MENTOR')` maps to `/mentor`, which is precisely why the component takes a `role` prop — so "view the app as this mentor" lands on the mentor dashboard, not `/portal`.

Worth knowing, since it is a product call: **the new view-as button can generate mentor-facing notifications.** `/api/admin/impersonate` writes an `IMPERSONATE_START` audit row, a warning-level activity entry, and notifies the target that their account was accessed. That is by design (transparency) and already reachable from `/admin/users` — but say the word and I will drop the view-as half and keep only "send a message".

The header also needed fixing before gaining two more buttons. It was a `flex-shrink-0` action column opposite an identity block with **no `min-w-0`** — exactly the #1305 shape, where the name/email column is the only thing that can give, and TR "Bu kullanıcı gibi gir" / DE "Nachricht senden" are the long labels. It now uses the candidate page's proven classes: `flex-wrap items-start`, `min-w-0` + `break-words` on the identity column, no `flex-shrink-0` on the actions.

## Defect 2 — pipeline stage labels rendered in English in a Turkish UI

**Root cause: `resolvePipelineStages(orgId, locale: Locale = 'en')` has a silently-English default, and the one caller whose output is written back to the database omitted it.**

1. `GET /api/admin/organizations/[id]/pipeline-stages` resolved the stages with no locale, so it answered with the 13 built-in labels in **English**, whatever the admin's language.
2. The stage editor treats that response as editable state and prefills 13 text inputs from it.
3. Save PUTs the whole prefilled set straight back, and the PUT persisted every label unconditionally (`label: s.label.trim()`).

So **one Save on an untouched editor** — or a recolour, or a reorder — converted "13 localized built-in stages" into "13 hardcoded English rows". After that `useResolvedStages()` renders them correctly as-is in all three languages, because `PipelineStage.label` is a single string with no per-locale variant. `DELETE` ("Reset to defaults") had the same omission, so the reset repopulated the editor in English too.

**What changed** (~40 lines, no schema change, no data migration):

- `isDefaultLabel(key, label)` in `src/lib/pipeline.ts` — true when a stored label is blank or is a byte-for-byte copy of one of *our own* built-in labels, in **any** of the three locales. Matching any locale is what closes the recolour/reorder hole; the cost, stated in the code comment, is that a tenant cannot pin our German string as a fixed label for Turkish readers.
- `PUT` now stores `""` for a stage nobody renamed. The public contract is unchanged (zod still requires `label.min(1)`) — the server normalizes, so no client needs to know about the sentinel. Only labels an admin actually typed are persisted, and those still render verbatim to everyone.
- Both read paths resolve it: `resolvePipelineStages()` on the server, and `useResolvedStages()` on the client for the labels `resolveCustomStages()` sends down raw. This makes the invariant the client file's comment already claimed true for a *partially* customized tenant, which it silently was not.
- `stageLabel()` falls through on a blank label (`??` → `||`).
- The editor's `GET`, `PUT` response and `DELETE` now resolve in the viewer's locale via `getLocale()`, so an admin sees Turkish while editing.
- `GET /api/admin/stage-sla` likewise — it ships `label` to `StageSlaEditor`, which prints it verbatim, so the whole stage-SLA settings list read English for a default-stage org.

A comment on `PipelineStage.label` in `prisma/schema.prisma` records the `""` contract. **It is a comment only — no column, type or index changed, so `db push` is a no-op.**

### Operator note: the org whose labels are already frozen

Rows that the editor already froze **heal by themselves on deploy** — `isDefaultLabel` recognizes them at read time, so a row storing `220 · Awaiting approval` renders `220 · Onay bekliyor` for a Turkish reader with no data change. I did not connect to any database and shipped no backfill.

The one case the code cannot fix is a tenant whose stage labels are **hand-written English that differs from our built-in strings** (e.g. someone typed `First contact` rather than `100 · First contact`). Those are indistinguishable from a deliberate rename, so they are honoured verbatim — by design. Clearing them is an operator action: open `/admin/organizations/<orgId>/pipeline` as that tenant's admin and press **Reset to defaults**, which drops the custom rows and returns the org to the localized built-ins. Worth checking first which case you are in — the editor's own sentence says whether the tenant uses custom stages, and `GET /api/admin/organizations/<orgId>/pipeline-stages` (two reads, writes nothing) returns the 13 labels to compare.

## Tests

- `e2e/mentor-detail.spec.ts` — new test on the affordance: `user-quick-actions` and `quick-login-as` visible, then one click on `quick-message` and assert the URL lands on `/messages/c/<id>`. **These three testids had zero coverage anywhere before this**, on either page, so this is also the first end-to-end proof of the ADMIN → MENTOR branch of `canMessage`. `quick-login-as` is asserted but never clicked (`e2e/impersonation.spec.ts` owns that flow, and starting one here would notify the mentor). One click only — `/api/conversations` is rate-limited to 20 per 15 minutes per shared runner IP. The conversation is deleted in `finally`; participant rows cascade with the user, the conversation row does not.
- `e2e/pipeline-stages.spec.ts` — `isDefaultLabel` across all three locales, blank, whitespace, a near-miss label and a custom key; plus a DB-backed test that a row stored `""`, a row stored with the frozen English default and a row with a genuinely typed label (`Ön görüşme`) resolve correctly in `tr`, `de` and `en`.
- `e2e/mobile-layout-audit.spec.ts` — `/admin/mentors/<id>` joins the 360px Turkish admin loop. The audit only ever measured the mentor **list**, which is where #1305 was filed; the detail page header is now the one with four controls.
- Not tagged `@smoke` (the mentor-detail test): CLAUDE.md caps that set at ~15–20 and this is not a critical path. The mobile-audit test it joins is already `@smoke`, so the count does not grow.

## Checks run locally — all green

| Check | Result |
|---|---|
| `npx prisma generate` | ok |
| `npx prisma validate` | schema valid |
| `npx tsc --noEmit` | 0 errors |
| `npm run lint` | 0 errors (2 pre-existing warnings, unrelated files) |
| `npm run check:i18n` | OK — 4460 keys × 3 locales |
| `npm run check:release-fragments` | OK — ships as `0.168.3-beta` |
| `npm run check:stage-keys` | OK — no new hardcoded key |
| `npm run check:locale-dates` | OK |
| `npm run test:unit` / `test:unit:coverage` | 59/59 pass, no floor breached |
| `npm run test:release`, `check:auth-reads`, `check:api-key-routes`, `check:query-scalars`, `check:tenant-models`, `check:demo-blocklist`, `check:events`, `check:openapi` | all OK |
| `npm run build` | succeeds |

**The e2e specs were not executed locally** — this box has no MySQL and I deliberately connected to no database. They run in CI.

`check:release-fragments` also prints a pre-existing warning that 39 fragments are pending and `release-compact.yml` looks stuck; the script itself flags it as "Not a problem with this PR", but it is worth a look.

## Deliberately out of scope

Filed as remaining leaks in #2268: the ICS feed (`api/calendar/feed/[token]`, locale-less **and** bypasses custom stages — calendar clients cache hard and it deserves its own verification pass), the weekly analytics report e-mail (interpolates the raw enum key; needs the row construction moved inside the per-admin loop), localizing the pipeline editor page itself (100% hardcoded English — a whole new dictionary namespace × 3 locales would bury this fix), and per-locale custom labels (`PipelineStage.labels Json`, the way `EvaluationCriterion.labels` already works). Also noted there: `/admin/mentors` renders the mentor name as a plain `<Link>` rather than a `PersonHoverCard`, so it is the one admin surface where a mentor's name offers neither a message nor a card.

### DIFF
--- a/docs/agent-experience.md
+++ b/docs/agent-experience.md
@@ -5975,3 +5975,40 @@ yaptığı atıfları düzeltmek gerekiyor. Marker kalmaması, metnin doğru olm
 kendiliğinden birleşti; yine de `grep -n "key: '"` ile iki girdinin de yerinde olduğuna
 baktım. Otomatik birleşme "çatışma yok" demektir, "iki taraf da korundu" demek değildir —
 kayıt defteri gibi dosyalarda sessizce bir girdi düşerse hiçbir tip hatası vermez.
+
+## 2026-09-07 — Bir kilit değerin kaynağı veriyse, kodu düzeltmek veriyi de iyileştirebilir (#2268)
+
+İki küçük kusuru tek PR'da kapattım (#2269). İkincisinin — Türkçe arayüzde İngilizce
+pipeline aşama etiketleri — bulgusu "isDataNotCode: true" idi ve doğruydu: semptomu üreten
+şey `PipelineStage.label` satırlarıydı. Refleks çözüm bir backfill betiği yazıp
+`infra/deploy-prod.sh`'e eklemek olurdu; bu, canlı satırları her ortamda yeniden yazmak
+demekti ve onay gerektirirdi. Bunun yerine **okuma yolunu** "bu etiket hâlâ bizim yerleşik
+etiketimizse, okuyanın diliyle çevir" diye değiştirdim. Aynı sonuç, sıfır veri mutasyonu,
+sıfır operatör adımı. Ders: veriden gelen bir semptomda önce "bu değeri kim okuyor ve
+okurken düzeltebilir miyim?" diye sor — backfill en son çare.
+
+**"Herhangi bir dilde eşleşiyor mu" kontrolü, kapatılması en zor deliği kapatıyor.**
+`isDefaultLabel` sadece okuyanın dilindeki yerleşik etiketle karşılaştırsaydı, editörde
+yalnızca **rengi** değiştirip kaydeden bir yönetici hâlâ bir dili dondururdu (prefill
+İngilizceydi, karşılaştırma Türkçeydi, eşleşme olmazdı). Üç dilin hepsine bakmak bu deliği
+kapatıyor; bedeli — bir kiracı Almanca yerleşik dizeyi Türkçe okuyuculara sabitleyemiyor —
+kod yorumuna yazılmalı, yoksa bir sonraki okuyan bunu hata sanar.
+
+**Konu numarasını issue'yu AÇTIKTAN sonra yaz.** Yorumlara ve release fragment'ına `#2270`
+yazdım (sıradaki numarayı tahmin ederek); `gh issue create` `#2268` verdi ve tahminim iki
+şey birden kaydırmıştı. Dokuz dosyada `sed` ile düzeltip kontrolleri baştan koşturmak
+gerekti. Ya issue'yu ilk iş aç, ya da numarayı yer tutucu bırakıp commit'ten önce doldur.
+
+**`npm install`, PR'ına versiyon değişikliği sızdırır.** Yerel `node_modules` eksikti
+(`@axe-core/playwright`, `web-push` yüzünden `tsc` 13 sahte hata veriyordu), `npm install`
+sorunu çözdü ama `package-lock.json`'ın `version` alanını 0.135.0 → 0.156.21 diye
+güncelledi. CLAUDE.md PR'ların versiyon dosyalarına dokunmamasını söylüyor: commit'ten önce
+`git checkout -- package-lock.json`. `git status`'u diff'e bakmadan `git add -A` ile
+geçmeyin.
+
+**`node --experimental-strip-types` `@/` takma adını çözmez.** `scripts/test/*.test.mjs`
+birim koşucusu `.ts` dosyalarını doğrudan import ediyor, ama `src/lib/pipeline.ts`
+`@/i18n/config`'ten import ettiği için o koşucudan erişilemiyor (coverage raporu da
+"Playwright altında, #1598 bekliyor" diyor). Saf yardımcıları yine de doğrulamak için
+`/tmp`'ye kopyalayıp import satırını `sed`'le göreceli hale getirdim: 30 saniye, ve üç
+dilin çıktısını gözle gördüm. Veritabanı olmadan da saf mantık kanıtlanabilir.
--- a/e2e/mentor-detail.spec.ts
+++ b/e2e/mentor-detail.spec.ts
@@ -39,3 +39,39 @@ test('mentor detail page shows expertise, capacity and active mentees', async ({
     await cleanupByEmail(menteeEmail);
   }
 });
+
+// #2268: the admin mentor profile is where every "open profile" for a MENTOR
+// lands, and it was the one person page with no way to contact them — the
+// affordance existed on the candidate profile and in the hover cards only. Also
+// the first end-to-end proof of the ADMIN → MENTOR branch of `canMessage`.
+test('mentor detail page offers the message and view-as shortcuts', async ({ page }) => {
+  const mentorEmail = uniqueEmail('mdqa-mentor');
+  const mentor = await seedUser(mentorEmail, 'Pass1234!', 'MENTOR', 'MDQA Mentor');
+  let conversationId: string | null = null;
+
+  try {
+    await page.goto('/auth/signin');
+    await page.fill('input[type="email"], input[name="email"]', ADMIN_EMAIL);
+    await page.fill('input[type="password"]', ADMIN_PASSWORD);
+    await page.click('button[type="submit"]');
+    await page.waitForURL((u) => !u.pathname.includes('/auth/signin'), { timeout: 20_000 });
+
+    await page.goto(`/admin/mentors/${mentor.id}`);
+    await expect(page.getByTestId('user-quick-actions')).toBeVisible();
+    // Present, but NOT clicked — e2e/impersonation.spec.ts owns that flow, and
+    // starting one here would notify the mentor and write an audit row.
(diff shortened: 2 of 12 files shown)


mersahin:
Preview environment for this PR (per `topic-preview.yml`): https://pr2269.interncrm.com

The two things to click:
1. `/admin/mentors/<any mentor id>` — the new **Mesaj gönder** / **Bu kullanıcı gibi gir** buttons in the header, and the header at phone width in Turkish.
2. `/admin/candidates/<id>` with the UI in Turkish — the pipeline stage `<Select>` options. On a topic env the DB is fresh, so it exercises the built-in path; the frozen-label path is what the new `e2e/pipeline-stages.spec.ts` case covers.

PR #2290, opened 2026-09-07
feat: render the monthly AI quota in settings and translate the pipeline-stage editor (#1625)

### LINKED ISSUE #1625: 🌱 [Admin] Task · Render `aiMonthlyQuota` in the settings form and translate the pipeline-stage editor
## 🎯 Goal
Two small credibility holes on screens a buyer clicks in the first ten minutes: a real setting that no UI ever renders (the AI budget can only be changed with a raw API call or a DB write), and an otherwise trilingual product whose per-tenant pipeline-stage editor is hardcoded English. Both are cheap; both are visible.

## 🔍 Current state
- `src/lib/settings.ts:32-35` — `aiMonthlyQuota: '200'` is a real, enforced setting (`src/lib/aiGate.ts:25` reads it and blocks AI calls when the pool is spent).
- `src/app/api/admin/settings/route.ts:27` — the PUT schema **accepts** `aiMonthlyQuota: z.string().regex(/^\d{1,6}$/)`.
- `src/app/admin/settings/page.tsx:128` — the save payload is
  `{ reminderDays, retentionMonths, supportEmail, weeklyDigest, require2fa, selfRegistration, earlyAccessWindowDays, premiumAnalytics, outcomeAutoSend, blindReview }` — `aiMonthlyQuota` is absent, and there is no state hook for it at `:13-24`.
- `src/app/admin/organizations/[id]/pipeline/page.tsx:88-101` — hardcoded English: `Pipeline stages`, `Customize this organization's pipeline stage labels, order, colors and grouping.`, `This tenant uses custom stages.`, `This tenant uses the built-in defaults.`, `Custom pipeline stages require a paid plan. You can preview the defaults below, but saving is disabled on the FREE plan.`, `← Organizations`, `Loading…`. Everything around it (`/admin/settings`) is translated through `useT()`.
- `src/i18n/dictionaries.ts` — a single typed EN/TR/DE source; `npm run check:i18n` enforces key parity across all three and fails CI on a missing key.

## 🛠 What to build
1. Add an `aiMonthlyQuota` state hook alongside the others at `src/app/admin/settings/page.tsx:13-24`, populate it in `load()` (next to `:110-117`), render a numeric `Input` in the settings form, and include it in the PUT payload at `:128`.
2. Give it a label and a short hint explaining what it caps (AI provider calls per calendar month across the tenant; `0` disables AI entirely) — EN/TR/DE keys under the existing `settings` block in `src/i18n/dictionaries.ts`.
3. Replace every hardcoded string in `src/app/admin/organizations/[id]/pipeline/page.tsx` with `useT()` lookups under a new `pipelineStages` (or existing `organizations`) block, EN/TR/DE.
4. Sweep the same file for any remaining English left in buttons, empty states and error text.

## ✅ Acceptance criteria
- [ ] An admin can read and change the monthly AI quota from `/admin/settings` and the new value is enforced by `runAiGated` without a server restart
- [ ] The numeric input rejects a non-numeric value client-side and the API still validates it server-side (the existing regex is unchanged)
- [ ] `/admin/organizations/<id>/pipeline` renders fully in Turkish and German with no English left, including the FREE-plan notice
- [ ] `npm run build` passes
- [ ] an e2e spec asserts the quota round-trips and the pipeline page has no untranslated string in TR
- [ ] a release fragment `releases/unreleased/settings-surface-completeness.json` is added

## 🇹🇷 Türkçe özet
İki küçük ama görünür eksik. Birincisi: `aiMonthlyQuota` gerçek ve işleyen bir ayar (kota dolunca AI çağrıları kesiliyor) ama hiçbir ekranda görünmüyor — bugün ancak elle API çağrısıyla ya da veritabanından değiştirilebiliyor. İkincisi: ürünün geri kalanı üç dilli iken organizasyonun pipeline aşama editörü baştan sona İngilizce. İkisi de yarım günlük iş ve alıcının ilk on dakikada gördüğü ekranlarda duruyor. Yeni metinlerin EN/TR/DE üçünde de anahtarı olmalı, yoksa CI kırmızı olur.

## 🤖 Agent prompt
```
Objective: two small admin-surface fixes.
(1) render aiMonthlyQuota in /admin/settings; (2) translate the org pipeline
    stage editor into EN/TR/DE.

Open first:
  src/app/admin/settings/page.tsx:13-24 (state), :110-117 (load), :128 (PUT)
  src/app/api/admin/settings/route.ts:27 (the API already accepts the key)
  src/lib/settings.ts:32-35 (the default + what it means)
  src/lib/aiGate.ts:25 (who enforces it)
  src/app/admin/organizations/[id]/pipeline/page.tsx:84-105 (hardcoded English)
  src/i18n/dictionaries.ts (add keys to en, tr AND de)

Constraints:
- `npm run check:i18n` enforces EN/TR/DE key parity and runs in CI. A key added
  to `en` only will fail the build. Write real Turkish and German, not English
  placeholders.
- Do not weaken the server-side zod validation to make the client simpler; the
  API is the authority.
- E2E locator pitfall: AdminNav renders its own input[type="search"] on every
  admin page, so an unscoped input[type="search"] selector hits the sidebar
  filter. Add a data-testid to any input you assert on.
- getByText('X') is substring matching in Playwright; use { exact: true } or
  scope to a container.
- Add a release fragment under releases/unreleased/ (bump: "patch").

Done when: the quota is editable from the UI and enforced, the pipeline editor
is fully translated, `npm run check:i18n` and `npm run build` pass.
```

**Zorluk** · S · **Tahmini** · 0.5-1 day · **Bağımlılık** · yok

mersahin:
## Summary

Two small holes on admin screens a buyer clicks early.

`aiMonthlyQuota` is a real, enforced setting — `runAiGated` reads it and refuses provider calls once the month's pool is spent — but no screen ever rendered it, so the only way to change it was a raw `PUT /api/admin/settings` or a DB write. It is now a numeric field in the system-settings form, loaded in `load()` and posted **when it changed**. Its hint says what the number actually caps: the pool is shared by CV extraction, CV feedback, interview prep, interaction summaries and mentor suggestions; only a call that succeeded is counted; the counter resets on the 1st; `0` switches AI off entirely; and the pool is measured **installation-wide**, not per organisation. That last part is the truth of the code rather than a wish: the limit resolves tenant row → global row → default like every setting, but `AiUsage` has no `orgId` and is not in `TENANT_MODELS`, so `getAiQuota()` counts every tenant's calls against whatever number this field holds. `docs/ai.md` now states the asymmetry and what scoping the meter would take. The API's `\d{1,6}` regex is untouched and remains the authority — `min={0} max={999999}` on the input only mirrors it.

The per-tenant pipeline-stage editor (`/admin/organizations/[id]/pipeline`) was hardcoded English in an otherwise trilingual product. Every string moves into a new `pipelineStages` dictionary block with real EN/TR/DE copy.

One deliberate wording decision: the PUT behind **Save** replaces the stage set with `deleteMany` + `createMany` and remaps no relation (noted during #1634's review). That behaviour is **not** changed here — but since a translation pass is where copy gets written, the save hint says the whole set is *replaced*, every row rewritten, not just the ones you touched (with the reassurance that is actually true: the stage keys are unchanged, so nobody moves stage). Nothing in the new copy makes the destructive path sound like a partial edit.

Closes #1625

## Changes

- `src/app/admin/settings/page.tsx` — `aiMonthlyQuota` state hook, populated in `load()`, rendered as a numeric `Input` (label + hint, `data-testid="ai-monthly-quota"`) next to the other budget/tier knobs. Three things guard the save: the value is validated against the API's own regex **before** the request and reported on the field when unusable (an empty box — the ordinary way to retype a number — used to 400 the whole payload and silently discard every other change on the form); a non-ok response or a network failure now renders a red `settings-save-error` box instead of looking like a success; and the key is sent only when it differs from what `load()` resolved, because a PUT writes the caller's own layer and posting an untouched inherited value would pin a tenant override at the operator's current number.
- `src/app/admin/organizations/[id]/pipeline/page.tsx` — `useT()` throughout: back link, heading, subtitle, custom-vs-defaults sentence, FREE-plan notice, `aria-label`s of the label and colour inputs, the off-path/final checkbox labels, both buttons, the loading line and every success/failure message. A server-side error detail (a developer-facing English string such as "Stage keys must be unique") is now shown as a small second line *under* the translated headline rather than replacing it. The custom-vs-defaults sentence sits in its own `<span data-testid="pipeline-stages-source">` so which of the two is on screen can be read exactly; new `data-testid`s on the page container and the save button.
- `src/i18n/dictionaries.ts` — `settings.aiMonthlyQuota`, `aiMonthlyQuotaHint`, `aiMonthlyQuotaInvalid`, `settings.saveFailed` and the new `pipelineStages` block, in EN/TR/DE.
- `docs/ai.md` — the limit and the counter are scoped differently; says so, and what scoping the meter per tenant would take.
- `docs/research/competitive-analysis-2026-08.md` — the AI-admin-console row no longer claims the quota is unsettable from the admin UI (🟡 quota control only; usage dashboard and provider status still missing, pool shared across tenants).
- `e2e/settings-ai-quota-and-stage-editor-i18n.spec.ts` — new spec: the quota round-trips through the form (resolved value rendered → change → persisted → shown again on reload, then restored through the API so a run against a deployed `BASE_URL` leaves that environment as it found it) and the stage editor renders in Turkish with no English left in the page container. Not tagged `@smoke` (neither flow is critical-path, and the smoke set is kept small).
- `releases/unreleased/ai-quota-field-and-stage-editor-i18n.json` — `patch` fragment with EN/TR/DE notes (both halves are user-visible).

Still out of scope: the same whole-payload rejection is reachable through the eleven pre-existing numeric keys on this form, and the same override pinning applies to all of them. A save that fails for any reason is now visible, which is the part worth fixing here; a whole-form dirty diff is a wider behaviour change than this PR's field.

Preview: https://pr2290.interncrm.com

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (lint prints only the two pre-existing `exhaustive-deps` warnings in unrelated files)
- [x] `npm run check:i18n` — 4551 keys × 3 locales OK
- [x] `npm run check:release-fragments` — fragment valid (it also warns that 49 fragments are pending, i.e. compaction is behind; unrelated to this PR)
- [x] `npm run check:stage-keys` — no new hardcoded canonical key
- [x] `npm run build` passes
- [ ] e2e not executed: this container has no database and no browser, so the new spec is written but unrun — CI is the first run

## Contributor terms


mersahin:
Review findings applied in 4935957. All five reproduced against the code; nothing was dismissed as a false positive, but two suggested fixes were narrowed — reasons below.

**1. `{ exact: true }` on the Turkish stage-editor assertion (major) — fixed.** Confirmed: `page.tsx` renders one paragraph made of subtitle + `' '` + the custom-vs-defaults sentence, so no element's text equals that sentence alone and the locator resolved to 0. Rather than dropping `exact` (which would have left the assertion satisfied by a substring of a paragraph that also contains the subtitle), the sentence now has its own element — `<span data-testid="pipeline-stages-source">` — and the spec asserts `toHaveText(...)` against it. Which of the two sentences is on screen is the one fact this test exists to read exactly.

**2. The quota field could 400 the whole settings PUT and discard unrelated changes (major) — fixed.** Confirmed: `aiMonthlyQuota` was in the payload unconditionally, `route.ts` rejects the entire body on a `\d{1,6}` miss, and the page reacted only to `res.ok`. Three changes: the value is checked before the request and, when unusable, reported on the field itself (`t.settings.aiMonthlyQuotaInvalid`) while the rest of the form still saves; a non-ok response or a network failure now renders a red `settings-save-error` box (`t.settings.saveFailed`) instead of looking like a successful save — that silence covered the eleven pre-existing keys too; and the field is omitted when unchanged, which is what makes Path B (a legacy non-numeric row rendering as an empty box) harmless rather than a permanent 400. I did not add `required`: it would put a red asterisk on the only "required" field of a form where nothing else carries one, and a legacy junk row would then block every save behind a native validation bubble.

**3. The hint promised a per-organisation pool the metering does not implement (minor) — fixed by wording.** Confirmed: `AiUsage` has no `orgId`, is absent from `TENANT_MODELS`, and `getAiQuota()` counts the month unfiltered, while the limit itself resolves per tenant. The EN/TR/DE hint now says usage is metered across the whole installation and that on a shared installation every organisation spends from the same month's pool; `docs/ai.md` states the asymmetry and what scoping the meter would take (`orgId` on `AiUsage`, a `TENANT_MODELS` entry, a filtered count). I did **not** gate the field to super-admin: `PUT /api/admin/settings` has accepted this key from any tenant ADMIN since #537, so hiding the input would change nothing about who can set the number — it would only put the control back out of sight, which is the problem #1625 was filed about. If the cap should be an operator-only decision, that is an API change and its own issue.

**4. Posting the quota untouched pins a tenant override (minor) — fixed for this field.** Confirmed: GET resolves tenant → global → default, PUT writes the caller's own layer, so the first save of any kind created `Setting(orgId=B, aiMonthlyQuota, <the operator's current number>)` and cut tenant B off from later changes to the global budget. The page now remembers what `load()` resolved and sends the key only when the string differs. Deliberately not generalised to a whole-form dirty diff in this PR: the same is true of the eleven pre-existing keys, and changing what all of them write is a wider behaviour change than a review fix on the field this PR adds.

**5. Stale gap-register row (minor) — fixed.** `docs/research/competitive-analysis-2026-08.md` now reads 🟡 quota control only, notes that the usage dashboard and provider status are still missing and that the pool is shared across tenants, with refreshed line references.

Also picked up the housekeeping note: the round-trip test no longer deletes a `Setting` row through Prisma. It reads the resolved value from the API first and puts it back the same way in `finally`, so a run against a deployed `BASE_URL` restores the environment it actually changed instead of editing the local database. The release fragment prose was updated to describe the error surfacing, the changed-only write and the installation-wide wording, and gained an EN/TR/DE note about the failed-save message.

The spec is still untagged, so the smoke gate will not execute it — the fix above is the reason it will pass in the scheduled full suite rather than a test run I could do here (no database and no browser in this container).

Static checks after the fixes: `npx tsc --noEmit` clean, `npm run lint` exit 0 (only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`), `npm run check:i18n` OK — 4551 keys × 3 locales, `npm run check:release-fragments` exit 0, `npm run check:stage-keys` OK (no new hardcoded key), `npm run build` succeeded. Schema untouched, so no `prisma validate` needed.

---

Issue #1732, opened 2026-08-30
🌱 🔗 [UX] Task · Pricing in the header, the footer and the landing free-core band


mersahin:
## 🎯 Goal
A pricing page nobody can find is the same as no pricing page. The public chrome currently offers Features / For companies / Showcase and never mentions money — so a visitor who wants the number leaves. This task makes the price one click from anywhere on the marketing site and puts the free-core promise where a mentor or mentee will actually read it.

## 🔍 Current state
- Header links array: `src/components/landing/PublicHeader.tsx:64-68` — `/features`, `/for-companies`, `/projects`. No pricing entry.
- Footer columns: `src/components/landing/PublicFooter.tsx:24-33` (Product column: features, for-companies, projects, release-notes, demo), `:36-44` (Community), plus a Legal column.
- Nav labels live in one i18n block per locale: `src/i18n/dictionaries.ts:615-627` (EN), `:4127-4139` (TR), `:7615-7627` (DE) — `home, dashboard, features, forCompanies, showcase, whatsNew, demo, becomeMentor, signIn, register, github, imprint, …`.
- The free-core sentence exists on the landing hero already: `src/i18n/dictionaries.ts:706` / `:4218` / `:7706` (`landing.heroModel`) — one line inside the hero, easy to miss.
- The landing page itself is `src/app/page.tsx`; it links to `/features` at `:443` and `:505`.
- `/pricing` itself is built in the sibling task — this task assumes it exists.

## 🛠 What to build
1. Add a `pricing` key to the `publicNav` block in **all three locales** (`Pricing` / `Fiyatlandırma` / `Preise`).
2. Add `{ href: '/pricing', label: n.pricing }` to the `links` array in `PublicHeader.tsx` (after `forCompanies`) and to the Product column in `PublicFooter.tsx`.
3. On the landing page, promote the free-core promise from a hero sub-line into a small dedicated band ("Always free for mentors and mentees — companies and programmes are what we charge for") whose only CTA is `/pricing`. Reuse the existing `landing.heroModel` string as the headline so no e2e assertion breaks; add one new supporting line + the CTA label.
4. Check the mobile disclosure menu in `PublicHeader.tsx` renders the new link too (the same `links` array drives it — confirm visually at a 375px viewport).

## ✅ Acceptance criteria
- [ ] "Pricing" appears in the public header (desktop **and** mobile menu) and in the footer's Product column, in EN/TR/DE
[...]

Issue #1730, opened 2026-08-30
💶 [Growth] Task · /pricing page — plan columns, comparison table and the metering rule in plain language


mersahin:
## 🎯 Goal
The page that ends the comparison. A program owner who lands here must be able to answer, without talking to anyone: what does this cost, what counts against the price, what am I never charged for, and what happens if I go over. This is also the page that has to make the free-core promise readable as a *structure* rather than marketing — the metering unit is a matched pair, so mentors and mentees literally cannot be billed.

## 🔍 Current state
- No `/pricing` route exists — `src/app/` has `features/`, `for-companies/`, `demo/`, `projects/`, `release-notes/`, but no `pricing/`.
- The public page pattern to follow: `src/app/for-companies/page.tsx:1-16` (`PublicShell`, `getServerDictionary`, `export const dynamic = 'force-dynamic'`), and `src/app/features/page.tsx`.
- The free-core claim already shipped on the hero and must be echoed, not contradicted: `src/i18n/dictionaries.ts:706` (EN), `:4218` (TR), `:7706` (DE).
- The pilot framing this page replaces: `src/app/for-companies/page.tsx:69-72` renders `t.forCompanies.pilotTitle` / `pilotBody`; the body text is `src/i18n/dictionaries.ts:2922` ("We charge nothing… pricing will be worked out together with the first companies") and `t.landing.faqCompany3A` at `:860` says the same. Both must be updated in this PR or the site will state two contradictory pricing policies.
- The catalogue module `src/lib/pricing.ts` lands in the sibling task — this page must read every number from it.
- **Related: #1403** (the "don't invent a number" stub) — this task is its replacement; close #1403 as superseded in the PR body.

## 🛠 What to build
1. **`src/app/pricing/page.tsx`** — server component in `PublicShell`, `force-dynamic`, sections in this order:
   - **Free-core band** at the very top (distinct visual strip): mentors and mentees are free forever, no card, no seat count — plus the explicit list rendered from `PRICING.NEVER_METERED` (messaging, meetings, video, goals, evaluations, CV, portal, pipeline).
   - **Program-side plans**: four columns (Community / Program / Program Plus / Enterprise) with the published monthly and annual prices, the active-pair band, admin seats, and the headline capabilities.
[...]

PR #2296, opened 2026-09-07
fix: park the /pricing links until the page exists (#1730)

mersahin:
## main's merge gate is red, and this is why

Every `Playwright smoke` run since `012d03d2` fails — on `main` itself, and therefore on every open PR that gets rebased onto it. One test:

```
✘ e2e/smoke.spec.ts:30 › home page loads with favicon and no console errors @smoke
  Error: unexpected diagnostics: [console.error] Failed to load resource: the server responded with a status of 404 (Not Found)
```

Last green `main`: `50475136`. Red from `012d03d2` ("feat(landing): pricing in the header, the footer and a free-core band (#1732) (#2284)") onward.

## Root cause

#2284 shipped three links to `/pricing` — `PublicHeader`'s `links` array, the footer's Product column and the free-core band's CTA — but **the page itself ships in #1730 and does not exist**. `src/app/pricing/` is absent; no open PR adds it.

The App Router prefetches every in-viewport `<Link>`, so the RSC prefetch of a route with no page returns 404 directly into the browser console, and `smoke.spec.ts:30` asserts the landing page emits no console diagnostics at all.

Why #2284's own gate stayed green is worth knowing: `e2e/pricing-nav.spec.ts` asserted `toHaveURL(/\/pricing$/)` after clicking, and the URL changes whether or not the destination exists. It also referenced an `e2e/pricing.spec.ts` that was never written.

## Parked, not reverted

- `publicNav.pricing`, `landing.freeCoreBadge`, `freeCoreBody`, `freeCoreCta` stay in **all three** dictionaries. #1730 restores three lines plus its own nav assertions.
- The band keeps its badge, headline and body — only the click is missing.
- The CTA is not repointed: `freeCoreCta` promises the price in EN/TR/DE, and a button reading "See pricing" that lands on `/features` is worse than no button.
- Every comment left in place records where the parked entry goes back.

`e2e/pricing-nav.spec.ts` → `e2e/free-core-band.spec.ts`: the three link tests and the phone-menu pricing test go; the band content assertion and the 375px overflow check stay.

## Checks

`npx tsc --noEmit` clean · `npm run lint` (only the two pre-existing `exhaustive-deps` warnings) · `npm run check:i18n` OK, 4529 keys × 3 locales · `npm run check:release-fragments` OK, ships as `0.173.1-beta`.

Not run locally: the e2e suite — this machine has no reachable database (`.env` still points `DATABASE_URL` at `crm-preview.ersah.in:3306`, the pre-migration host), so `/` 500s locally on `prisma.evaluation.findMany()`. CI proves it instead.

## Related, and worth a maintainer's eye

`npm run check:release-fragments` warns that **49 fragments are pending** (threshold 12) — `release-compact.yml` appears to have stopped folding, so `CHANGELOG.md` and `src/lib/releaseNotes.ts` have not recorded what shipped. Pre-existing and unrelated to this PR; `releases/README.md` → "When compaction is stuck" has the recovery.

Closes nothing — #1730 stays open and is now the ticket that un-parks these links.


mersahin:
@copilot interncrm.com/pricing bu sayfa 404 veriyor nedendir?

Issue #2289, opened 2026-09-07
🔁 [Mentor] Task · A first-class mentor transfer — one transaction, a recorded reason, and a decision on what happens to the history


mersahin:
Follow-up to #2281 / PR https://github.com/21072026/Internship/pull/2283.

## The gap

There is no first-class **"change this mentee's mentor"** operation anywhere in the product. The only way to do it is close-then-create: mark the relation `COMPLETED` on `PUT /api/mentorship/[id]`, then `POST /api/mentorship` with the new mentor. Two independent writes, no link between them, no reason recorded, and the record afterwards claims the first mentorship *completed successfully* — which is the exact corruption #1801 was filed to stop for the mentee-initiated case.

Two things make it worth its own issue now.

**1. The guard turned a silent mistake into a visible dead end.** Before #2283, an admin who assigned a new mentor without closing the old relation got a second `ACTIVE` row — wrong, but it appeared to work. Now every write path asks `hasOtherActiveMentorship()` and answers `409 { code: 'already_mentored' }`. That is the correct refusal, but from the admin's side it is an unexplained wall in the middle of a legitimate task: *the mentor left the company and this mentee needs a new one.* The refusal tells them the mentee already has a mentor; it does not tell them the sequence is close-first-then-assign, and there is no control that does it for them. A correct guard with no supported path through it is how people learn to fake a completion.

**2. Every child record stays on the closed relation.** History hangs off `relationId`, not off the (mentor, mentee) pair. From `prisma/schema.prisma` (`model MentorshipRelation`), the relation owns:

`interactions InteractionLog[]`, `statusChanges StatusChange[]`, `meetings Meeting[]`, `messages Message[]`, `evaluations Evaluation[]`, `goals Goal[]`, `meetingRequests MeetingRequest[]`, `questions MentorQuestion[]`, `relationNotes RelationNote[]`, `offers Offer[]`, `weeklyReports WeeklyReport[]`, `weeklyReportReminders WeeklyReportReminder[]`.

[...]

PR #2319, opened 2026-09-08
docs: session retrospective — four lessons from the mentor-transfer work (#2289)

mersahin:
## Summary

The end-of-session retrospective CLAUDE.md asks for, from the #2289 work (merged as #2318). Pure docs — no release fragment (per `releases/README.md`, trivial non-user-facing changes need none).

Closes nothing.

## Changes

Four entries in `docs/agent-experience.md`:

- **A new column on `MentorshipRelation` is public to mentors by default.** ~40 route files read relations with Prisma `include`, and `include` selects every scalar — so an `endReasonNote` column would have shipped an admin's candid prose to the mentor it was about, with no guard failing. That is why #2289 stores a coarse code on the relation and keeps the note in the ActivityLog. `Offer.compensationNote` is the precedent.
- **Guards forbid the obvious next line before you write it.** `check:events` caps direct `dispatchWebhook` call sites at ten (#1697), so "just fire `mentorship.created`" is a red build. Read the `check:*` script that owns the surface first; the refusal is usually the right design advice.
- **The Playwright browser symlink, with the inner path spelled out** — CLAUDE.md says symlink it, the security playbook says that doesn't work and to use `executablePath`; both are half right. The missing detail: link the build's `chrome-linux/` directory (not the build root) and rename the binary to `chrome-headless-shell`. Runnable snippet included; with it the real suite runs in this container.
- **13 local `@smoke` failures were an unseeded database**, not a regression — every one was a spec signing in as the seeded admin. And the one that survived seeding reproduced identically on `origin/main`, which is a three-minute way to settle "did I break this?" with evidence.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (docs-only change; nothing in `src/`)
- [x] `npm run test:e2e` passing (or CI green) — CI will confirm; the branch touches no code
- [x] Tested the change manually / added an e2e test — n/a, documentation

## Contributor terms

https://claude.ai/code/session_01MNKY4pLa4kEzdF87VRyeCX

---
_Generated by [Claude Code](https://claude.ai/code/session_01MNKY4pLa4kEzdF87VRyeCX)_

### DIFF
--- a/docs/agent-experience.md
+++ b/docs/agent-experience.md
@@ -10,6 +10,71 @@ Newest entries on top.
 
 ---
 
+## 2026-09-08 — A new column on `MentorshipRelation` is public to mentors by default (#2289)
+
+Roughly forty route files read relations with Prisma **`include`**, and `include` selects
+*every* scalar. So `endReasonNote` — an admin's candid free text about why a pairing ended —
+would have shipped straight to the mentor it was about, on `GET /api/mentorship` alone,
+with no guard failing and no test noticing. #1801 makes "the outgoing mentor is never shown
+the reason" a rule, so the column had to go: the reason stays as a coarse **code** on the
+relation (the same kind of fact `StatusChange.reasonCode` already exposes there) and the
+prose lives in the `ActivityLog` entry, which only admins read. **Rule: before adding a
+column to a model that mentors or mentees can read, ask who receives it — the answer is
+"everyone with an `include`", not "whoever I wrote a `select` for."** `Offer.compensationNote`
+is the precedent to copy, comment and all.
+
+## 2026-09-08 — Guards can forbid the obvious next line before you write it (#2289)
+
+A new write path that creates a relation "obviously" wants
+`dispatchWebhook('mentorship.created')`. `npm run check:events` caps direct dispatcher call
+sites at **ten**, and an eleventh is a red build (#1697) — so the obvious line is a CI
+failure, discovered by grepping the guard rather than by pushing. Same shape as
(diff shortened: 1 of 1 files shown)


mersahin:
**Playwright smoke** is red on `bfa69db`, and it is not this PR's failure. Re-running the job once; details so the claim is checkable rather than asserted.

The failure: `e2e/project-mentee-owner.spec.ts:25` — *a mentee creates a project from the portal and edits the one they own* — 1 failed / 138 passed, both the initial attempt and retry #1. The mentee renames their project, the `PUT /api/projects/<id>` comes back `200`, and then `portal-projects-list` never shows `MO Solar Tracker v2` within 10s. A stale list after a successful save.

Why it cannot be this PR's:

- **The diff is one markdown file.** `docs/agent-experience.md` only — no `src/`, no `e2e/`, no schema, no config. There is no mechanism by which it reaches a portal project list.
- **The same spec passed on the immediately preceding run of identical application code.** This branch is `origin/main` + that one doc, and `origin/main`'s tip is #2318, whose own smoke run was green including this spec.
- **It passes locally on `origin/main`**: `npx playwright test e2e/project-mentee-owner.spec.ts` → 3/3, against a local MariaDB.

So the honest reading is a flaky refresh race in that spec rather than a regression, and there is no fix to port — nothing has changed in the code it exercises. It is not on the known-flake list in CLAUDE.md (`account-self-service`, `sign-out-all`), so if it fails again on the re-run it is worth its own issue about the list not refreshing after a save; that would be a bug in the portal projects list, still not in this PR.

Re-running the failed job now — the one re-run, spent here.

---

Issue #1819, opened 2026-08-30
🧹 [Migration] Task · Backfill every free-text skill onto the taxonomy, with a queue for what did not map


mersahin:
## 🎯 Goal
Every skill already in the database was typed by a real person describing themselves. The migration onto canonical rows must map what it can and *show* what it could not — an admin decides whether `Bilgisayar Görüsü` becomes a new canonical skill or an alias of `Computer Vision`. Nothing gets dropped on the floor.

## 🔍 Current state
- Three JSON columns hold the data to migrate: `User.skills` (`prisma/schema.prisma:221`), `MentorApplication.expertise` (written at `src/app/api/mentor-applications/route.ts:89`), `Requisition.requiredSkills` (`prisma/schema.prisma:2562`).
- Backfills in this repo are standalone `prisma/backfill-*.mjs` scripts invoked from the deploy script, several of them `|| true` so a failure cannot wedge a deploy: `infra/deploy-prod.sh:489-521` (`backfill-company-interest-scope.mjs`, `backfill-json-columns.mjs --repair`, the idempotent seed block) and the legacy path `.github/workflows/deploy.yml:187-192`. #1540 proposes a `prisma/backfills/` registry to replace the `|| true` lines — follow it if it has landed, otherwise match the existing convention.
- `scripts/backfill-requisitions.mjs` + `scripts/requisition-backfill-lib.mjs` are the closest existing example of a backfill with its own test-covered library.
- Turkish text needs the `transliterate()` path already used by `src/lib/requisitions.ts:47-59` — `İ`/`ı` will otherwise produce false non-matches.
- The demo seeder is local-only by policy and refuses non-local `DATABASE_URL`s (`docs/DATA_ACCESS_POLICY.md`); develop and verify this against synthetic data, never against preview or production PII.

## 🛠 What to build
1. `prisma/backfill-user-skills.mjs` — for every user, application and requisition:
   - resolve each free-text entry through `resolveSkill()` (exact key → alias → transliterated key → case-folded);
   - create the `UserSkill` / `RequisitionSkill` row (skip on the unique key, so re-running is a no-op);
   - when nothing resolves, record an `UnmappedSkillTerm` row (`orgId`, `term`, `normalizedKey`, `occurrences`, `firstSeenAt`, `resolution` = `PENDING` | `PROMOTED` | `ALIASED` | `DISCARDED`) instead of guessing.
[...]

Issue #1816, opened 2026-08-30
🌱 🌍 [i18n] Task · Starter skill vocabulary (EN/TR/DE) with aliases and adjacency, plus a check:skills gate


mersahin:
## 🎯 Goal
An empty taxonomy is a worse experience than free text — the first mentee to open the picker would see nothing and give up. Ship a curated starter vocabulary so every new organisation begins with a few hundred real, trilingual, aliased skills, and a CI check that stops the file rotting.

## 🔍 Current state
- A usable seed already exists but is locked inside the CV parser: `SKILL_VOCABULARY` at `src/lib/cvParse.ts:23-32` — roughly 70 terms across languages, frameworks, databases, cloud/devops, ML, tooling and spoken languages. It is used only for word-boundary matching at `src/lib/cvParse.ts:76-88` and is exposed to no input anywhere.
- Nothing in it has a Turkish or German label, an alias, or a relationship to any other term.
- The repo's precedent for a curated, trilingual content file is `src/lib/newsletterContent.ts` (house style documented in the file) and `src/lib/features.ts` + the `featureCatalog` i18n block. `npm run check:i18n` (`scripts/check-i18n.ts`) enforces EN/TR/DE key parity in CI; other `check:*` validators live alongside it (`scripts/check-k6.mjs`, `scripts/check-release-fragments.mjs`, `scripts/check-openapi.mjs`) and are wired into `.github/workflows/ci.yml`.
- The seeders this must match in style: `prisma/seed.mjs` (first admin) and `prisma/seed-demo.mjs`; deploy-time idempotent seeds are invoked from `infra/deploy-prod.sh:520-521`.

## 🛠 What to build
1. `src/lib/skillSeed.ts` — a typed, exported array. Each entry: `key`, `labelEn`, `labelTr`, `labelDe`, `category`, `aliases: string[]`, `adjacent: string[]` (keys, with an optional weight).
[...]

PR #2327, opened 2026-09-09
feat: starter skill vocabulary (EN/TR/DE) with aliases and adjacency (#1816)

mersahin:
## Summary

Skills are free text today, so `React`, `react`, `ReactJS` and `React.js` are four different
skills and no filter over them can be right. The taxonomy work needs something to normalise
*to*, and an empty vocabulary is a worse experience than free text: the first mentee to open a
picker that offers nothing gives up and types a blob.

This ships that vocabulary — **101 curated skills in 13 categories**, each with a stable key,
EN/TR/DE labels, the spellings people actually type, and near-neighbour links — plus the CI gate
that stops the file rotting.

It was grown from `SKILL_VOCABULARY` in `src/lib/cvParse.ts` (the only curated list that
existed; a test asserts **every** term of it resolves here), plus the umbrella terms it lacked —
Frontend, Backend, Full Stack, Mobile, DevOps, Data Engineering, QA/Testing, UI/UX Design,
Product Management — and the composite strings that appear *verbatim* in real profile data
(`Git / GitHub`, `Entity Framework Core`, `Generative AI / AI Tools`, `REST API / API
Development` all come from the pasted CV blob quoted at the top of `skills.ts`, so those exact
strings are aliases).

On the Turkish and German: proper nouns (`React`, `Docker`, `PostgreSQL`) are the same string in
all three locales, which is correct and which the checker allows on purpose. Where the English
term is what a Turkish or German practitioner actually writes (`Agile`, `DevOps`, `Data
Science`), the label stays as it is and the honest local phrasing goes into the aliases
(`Çevik`, `Agiles Arbeiten`, `Datenwissenschaft`) — a label nobody would ever type is worse than
no translation. Role and concept terms do get real translations (`Veri Mühendisliği`,
`Datenbankadministration`, `Maschinelles Lernen`, `Ön Yüz Geliştirme`).

Refs #1816

### Scope boundary — deliberately not in this PR

**This is the vocabulary only.** No `Skill`/`UserSkill` schema model, no filter or matcher
rewiring, and no migration of the existing free-text `User.skills` values — those are
#1815/#1819/#1820 and touching them here would collide with work another session may hold.
`src/lib/skillVocabulary.ts` is exported but read by nothing yet, on purpose, so #1819/#1820 can
be written against a stable validated vocabulary instead of inventing one on the way past.

**Left out because it is blocked, so the issue stays open:** `prisma/seed-skills.mjs` (issue item
2, and two of the acceptance criteria) upserts on `[orgId, key]` — but `prisma/schema.prisma` has
no `Skill` model, `#1815` has not landed, and the schema is explicitly out of scope for this
task. There is nothing for a seeder to upsert into yet. It belongs in the PR that adds the model,
where its idempotence can actually be run. Hence `Refs`, not `Closes`.

## Changes

- **`src/lib/skillSeed.ts`** — the vocabulary. `SkillEntry` = `key` + `labelEn/labelTr/labelDe` +
  `category` + `aliases` + `adjacent`, and `SKILL_CATEGORIES` for the picker grouping. House style
  for adding an entry is documented in the file header. Adjacency is authored in **one** direction
  and symmetrised by `buildSkillIndex`, and is deliberately **unweighted** — the matcher (#1819)
  owns scoring, and a weight guessed here would be a number nobody measured. The module is
  **dependency-free** (same reason as `lastContactRule.ts`) so a plain Node script can load it;
  that is why `buildSkillIndex` takes the fold as an argument rather than importing it, so there
  is no second copy of `skillKey()` to drift.
- **`src/lib/skillVocabulary.ts`** — the app-facing wrapper: binds the seed to `skillKey()` (the
  one Turkish-aware fold this repo has) and exposes `resolveSkill`, `resolveSkillKey`,
  `skillNeighbours`, `skillLabel`.
- **`scripts/check-skills.mjs` + `npm run check:skills`**, wired into `ci.yml`.
- **`scripts/test/skill-vocabulary.test.mjs` + `npm run test:skill-vocabulary`**, also in `ci.yml`:
  every spelling of React collapsing to one key, adjacency coming out symmetric, TR/DE spellings
  resolving, free text resolving to `null` rather than a near miss, and the CV vocabulary being
  fully covered.
- **`releases/unreleased/starter-skill-vocabulary.json`** (`minor`). No `notes`: nothing is
  user-visible yet — no screen reads the vocabulary in this PR.

### Why a checker, and what it catches

Every way of breaking this file is invisible to TypeScript, because every field is a `string`. A
duplicated key, an alias two skills both claim, an adjacency pointing at a key somebody renamed,
or an empty Turkish label all compile perfectly and surface later as a picker with two "React"
rows or a lookup that resolves to whichever entry happened to be first. `buildSkillIndex()` is
deliberately *tolerant* of all of it (first claimant wins, unknown adjacency dropped) so a page
never 500s over a typo in a data file — this check is the other half of that deal.

Green on this branch:

```
skills OK — 101 skills in 13 categories, 361 lookup tokens (191 aliases), 304 adjacency link(s), all with EN/TR/DE labels.
```

The guard takes an optional path to an alternate seed module (the same trick as
`check-stage-keys.mjs`) so it can be exercised against a fixture. Each of the four error classes,
broken once:

**Duplicated key** (a second entry with `key: 'react'`) — exit 1:

```
skills check FAILED — 1 issue(s):
  • key "react": duplicated — aliases and adjacency cannot point at two entries
```

**Colliding alias** (`Postgres` added to `mysql`, where `postgresql` already claims it) — exit 1:

```
skills check FAILED — 1 issue(s):
  • token "postgres" is claimed twice: mysql (alias "Postgres") and postgresql (alias "Postgres") — a lookup would silently resolve to whichever comes first
```

**Dangling adjacency** (`react` adjacent to `reactjs`, which is not a key) — exit 1:

```
skills check FAILED — 1 issue(s):
  • key "react": adjacent "reactjs" is not a skill key in the vocabulary
```

**Missing locale** (`docker` with `labelTr: ''`) — exit 1:

```
skills check FAILED — 1 issue(s):
  • key "docker": labelTr is empty — all three labels are mandatory
```

It also fails on a key that does not survive `skillKey()` unchanged (such a key could never be
typed and matched), a self or duplicated adjacency, an unknown category, a category missing a
label, and an **alias made redundant by its own entry's key or label** — the index already holds
those, so it is pure noise. That last rule caught ten redundant aliases in my own first draft
(`NodeJS`, `NextJS`, `CSharp`, `GCP`, `Full-Stack`, …), which is exactly the rot it exists to
prevent.

## Verification

- [x] `npm run lint` + `npx tsc --noEmit` clean (only the two pre-existing `exhaustive-deps`
      warnings in `admin/companies` and `admin/mentorship`)
- [x] `npm run build` passes
- [x] `npm run check:skills` green, and red on each of the four error classes above
- [x] `npm run test:skill-vocabulary` — 8/8 pass
- [x] `npm run check:i18n` (4777 keys × 3 locales), `npm run check:stage-keys`,
      `npm run check:release-fragments` all green
- [ ] `npm run test:e2e` — not run: no database, no browser and no Docker daemon in this
      container. No e2e spec was added, because nothing user-visible changed: this PR adds a data
      module, a validator and a unit test, with no screen reading any of it yet.

Preview: https://pr2327.interncrm.com

## Contributor terms


mersahin:
All six review findings applied in ab61a47. They were worth doing properly rather than filing as follow-ups: a key here is forever, and #1819's matcher will act on whatever this file claims, so a *wrong* entry costs more than a missing one.

**Composites that resolved to nothing.** `SQL / PostgreSQL`, `C# / .NET` and `Swagger / API Testing` are verbatim lines of the real pasted CV quoted at the top of `skills.ts`, and all three returned `null`. The first two are now aliases of the **more specific** skill they name (`postgresql`, `csharp`): the general one is a single adjacency hop away, whereas resolving to `sql`/`dotnet` would throw the product away. Swagger/OpenAPI had no entry at all — a REST developer's most common second skill resolved to nothing — so it gains one (`backend`, adjacent to `rest`, `qa-testing`, `backend`).

**`Java C# / .NET` stays unresolvable**, deliberately and now documented: that is one line naming three skills because the CV had no separator there, and aliasing it to any of them would silently drop the other two. Asserted as a test so nobody "fixes" it later.

**The umbrella-as-alias false positive.** `NoSQL` was an alias of `mongodb`, and Redis and Elasticsearch are in this vocabulary and are also NoSQL stores — so a Redis engineer who wrote "NoSQL" was tagged a MongoDB engineer. Removed, and the general rule written into the house style rather than left as a one-off fix: *an umbrella term may alias a product only where this vocabulary holds no competing product under it.* `Version Control` → `git` and `Containerization` → `docker` pass that test and stay; a competitor arriving is what would make them wrong.

**The adjacency invariant matched the docs, not the data.** The house style said edges are "authored in ONE direction"; 125 of the 179 undirected edges are in fact written from both ends. Rather than rewriting 125 entries to satisfy a rule that buys nothing (`buildSkillIndex` symmetrises, and the unit test asserts the result), the invariant is now stated as what it actually is — an undirected relation, expressed directionally, written wherever it reads best — with the counts recorded so the next author is not told a rule the data has never followed.

**The CV-coverage test was asserting against a copy.** It listed the 76 parser terms by hand, which is a test that passes forever: adding a term to `cvParse.ts` could never fail it, so the coverage it claimed to prove would quietly stop being true. `SKILL_VOCABULARY` moves to a dependency-free `src/lib/cvSkillVocabulary.ts` — `cvParse.ts` imports `mammoth` and `pdf-parse` at module scope, so nothing under `scripts/` can load it — and both the parser and the test now read the one array. `cvParse.ts` re-exports it, so the import path is unchanged.

**Coverage.** `src/lib/skillSeed.ts` was in neither `FLOORS` nor `AWAITING_TESTS`, so a refactor of the index functions could have landed uncovered. Added at a 90% floor (measured 100%), with the caveat recorded in the entry that most of the file is the vocabulary *table*, so the line score is generous by construction and the real guards are `check:skills` on the data plus the behavioural test.

Verified: `check:skills` clean, `test:skill-vocabulary` 12/12, `npm run test:unit` 252/252, `test:unit:coverage` no floor breached, lint clean, `tsc --noEmit` clean, full `npm run build` passes, release fragment valid.

---

Issue #1556, opened 2026-08-30
📊 [Analytics] Task · ActivityLog, PageView and EmailLog get an orgId — and the admin telemetry routes get scoped


mersahin:
## 🎯 Goal
`/api/admin/activity` and `/api/admin/email-log` are the only analytics-family routes not wrapped in `withTenantScope`, and neither `ActivityLog`, `PageView` nor `EmailLog` is a tenant-anchored model. That means the moment isolation is switched on for a real multi-tenant sale, the **audit trail** — who signed in, who changed what, who was invited — and the **behavioural telemetry** and **recipient email addresses** still cross tenants. An auditor finds this in five minutes, and it is exactly the finding that fails an enterprise security review after we have told them isolation is on.

## 🔍 Current state
- `prisma/schema.prisma:2412-2432` — `ActivityLog { level, action, actorId?, actorEmail?, targetType?, targetId?, detail?, ip?, userAgent?, createdAt }`. No `orgId`.
- `prisma/schema.prisma:407-419` — `PageView { userId, path, durationSec, createdAt }`. No `orgId`.
- `prisma/schema.prisma:2453-2478` — `EmailLog { to, subject, category?, transport?, status, error?, createdAt }`. No `orgId`. The `to` column is a recipient address — personal data belonging to another tenant's users.
- `src/app/api/admin/activity/route.ts:12-37` — `getServerSession` + `role !== 'ADMIN'` check, then `prisma.activityLog.findMany({ where })` where `where` only ever holds `level` / `action` / `actorEmail`. **No `withTenantScope`.**
- `src/app/api/admin/email-log/route.ts:16-33` — same shape, no `withTenantScope`.
- `src/app/api/track/pageview/route.ts:30-50` — *is* wrapped in `withTenantScope`, but since `PageView` is absent from `TENANT_MODELS` (`src/lib/orgContext.ts:43-55`) the wrapper is a no-op for the row it writes.
- Writers to instrument: `src/lib/activity.ts` (`logActivity`, called from ~60 places), the pageview beacon above, and the mail path that writes `EmailLog` from `src/services/emailService.ts`.

## 🛠 What to build
1. Schema (additive, safe under `db push`): `orgId String?` + `@@index([orgId, createdAt])` on `ActivityLog`, `PageView` and `EmailLog`.
[...]

PR #2328, opened 2026-09-09
feat: per-org broadcast sending quota on announcements and newsletters (#1754)

### LINKED ISSUE #1754: 🌱 📣 [Comms] Task · Per-org broadcast sending quota on announcements and newsletters
## 🎯 Goal
Broadcast comms — newsletters, segmented announcements, campaign features — are programme-owner value and a natural paid tier. Today there is no volume quota, no plan gate and nothing protecting the shared sending reputation from one tenant mailing its whole list every day. A cap gives the commercial layer something to sell in this domain and gives the sending domain a floor of protection. **1:1 messaging, notifications and every lifecycle e-mail stay free forever** — this counts broadcasts only.

## 🔍 Current state
- Newsletters are dispatched by `src/lib/newsletterDispatch.ts` (one `NewsletterSend` row per recipient — `prisma/schema.prisma:917`), scheduled from `/api/cron/start` per the note in that file's header, with cadence settings at `src/lib/settings.ts:49-61`.
- Announcements: `prisma/schema.prisma:763` (`Announcement`), sent through `src/services/emailService.ts`.
- Every outgoing mail is already logged with a category and a transport: `EmailLog` (`prisma/schema.prisma:2453-2475`), indexed on `category` (`:2471`) and `createdAt` (`:2473`). The `transport` column at `:2464` distinguishes the reputable relay from the bulk channel — the comment there says the point is to see how much of the relay's daily quota each category consumes.
- `EmailLog` has **no** `orgId` — #1556 adds it, and this task needs it to count per tenant.
- No sending limit of any kind exists anywhere.

## 🛠 What to build
1. Add `monthlyBroadcastRecipients` to the limits in `src/lib/plans.ts` (Community small, Program larger, Program Plus larger still, Enterprise unlimited).
2. At the two broadcast entry points — scheduling/sending a newsletter issue, and sending an announcement — call `enforceQuota(orgId, 'BROADCAST_RECIPIENTS')` **before** the audience is expanded, and record `recordUsage({ metric: 'BROADCAST_RECIPIENT', quantity: recipientCount })` after a send.
3. The check is on the **recipient count of this send plus the month to date**, so a tenant one recipient over the band gets a clear 403 naming the number rather than a half-delivered blast. A send that would exceed the band is refused whole; it is never truncated silently.
4. Copy for the block and for a "you have used X of Y this month" line on the newsletter and announcement composers, in EN/TR/DE.
5. Explicitly **not** counted, and asserted by a test: 1:1 messages, notification e-mails, verification/invitation/reset mail, meeting reminders, weekly digests, dormant-first-contact mails. Only `category` values belonging to a broadcast are metered — enumerate them in one exported constant rather than by string comparison at the call site.

## ✅ Acceptance criteria
- [ ] A newsletter send that would exceed the month's band returns 403 with the limit and the month-to-date figure, and sends **nothing**.
- [ ] A send inside the band works exactly as before and increments the meter by the recipient count.
- [ ] A mentor sending a 1:1 message, a verification mail and a meeting reminder are all unaffected with the band set to zero — asserted in the spec.
- [ ] The composer shows the month-to-date usage in all three languages.
- [ ] `npm run build` passes
- [ ] an e2e spec covers it
- [ ] a release fragment `releases/unreleased/<kebab-slug>.json` is added

## 🇹🇷 Türkçe özet
Bülten ve duyuru gibi toplu gönderimler program sahibine ait bir değer ve doğal bir ücretli katman — ama bugün ne hacim sınırı ne plan kapısı var; bir kiracı her gün tüm listesine yazarak ortak gönderim itibarını yakabilir. Bu görev plana aylık alıcı sınırı ekliyor ve gönderim öncesi kontrol ediyor: bandı aşacak bir gönderim yarım gitmiyor, tümüyle reddediliyor ve kaç alıcı kaldığını söylüyor. **Birebir mesajlar, bildirimler, doğrulama/davet/şifre mailleri, toplantı hatırlatmaları hiçbir zaman sayılmaz** — bu testle güvence altına alınıyor.

## 🤖 Agent prompt
```
Objective: add a per-organization monthly broadcast recipient quota to newsletters and
announcements.

Open first:
  src/lib/newsletterDispatch.ts (dispatch + the cron registration note in its header)
  docs/newsletter.md (read before touching anything newsletter-related)
  prisma/schema.prisma: EmailLog :2453-2475 (category :2461, transport :2464),
                        Newsletter :846, NewsletterSend :917, Announcement :763
  src/lib/plans.ts, src/lib/quota.ts (enforceQuota), src/lib/usage.ts (recordUsage)
  src/i18n/dictionaries.ts

Constraints:
  - FREE CORE: never meter 1:1 messages, notifications, verification/invitation/reset mail,
    meeting reminders, weekly digests or dormant-first-contact mail. List the metered
    categories in ONE exported constant; do not compare category strings at call sites.
  - Refuse an over-band send WHOLE. A half-delivered blast is worse than a refusal.
  - Check before expanding the audience, meter after the send.
  - EmailLog needs an orgId to count per tenant - that is #1556. Depend on it.
  - A sent newsletter issue is immutable and undeletable (docs/newsletter.md) - do not add
    anything that mutates a sent issue.
  - All new copy needs EN + TR + DE keys.

Done when: over-band send 403s and delivers nothing, in-band send unchanged, free-core mail
proven unaffected with the band at zero, build + check:i18n green, release fragment, PR open.
```

**Zorluk** · S · **Tahmini** · 1 day · **Bağımlılık** · #1752, #1556 (EmailLog orgId)

mersahin:
## Summary

One tenant could broadcast to its whole list every day, and every other tenant on the **same sending domain** paid for it — a burst of complaints against one org's blast lands the shared domain in everybody's spam folder. That is the same constraint that already makes the dormant sweep send no third mail, ever.

Announcement e-mails and newsletter issues now share **one monthly recipient band per organization**:

```
min( plan's monthlyBroadcastRecipients , the broadcastMonthlyRecipients operator setting )
over the calendar month, in UTC
```

`src/lib/broadcastQuota.ts` is the single source of the rule. Several properties are load-bearing:

- **Usage is derived, never counted.** The month-to-date figure comes from rows that already exist — `NewsletterSend` rows belonging to the recipient's org, and `Announcement.emailedCount` for announcements sent by one of that org's admins. A counter of its own could drift from what was really sent, and the whole point of a refusal is that the number it names is true. (`EmailLog` still has no `orgId` — that is #1556 — which is why attribution goes through those two rather than the mail log.)
- **The check runs before the audience is expanded, and refuses whole.** `dispatchNewsletter` resolves the recipients and evaluates every tenant in them *before* it claims the issue, so a refusal never leaves a half-delivered row behind and the issue stays editable and cancellable by a human. It is **all-or-nothing per issue** — excluding one over-band tenant would mark the issue `SENT`, and a sent issue is immutable and undeletable, so that tenant could never receive it. Nothing partial, ever. (That rule has a cost at scale, written up as #2335.)
- **A refusal disarms what the refusal armed.** `scheduledAt` for a *send now* is `new Date()`, so an issue left `SCHEDULED` after a 403 is *due*: the 15-minute cron would mail, unattended, exactly the issue the admin was just told had not been sent — and pressing Send again would arm a second one that also delivers. Both send-now paths therefore put the row back to `DRAFT` with `scheduledAt` cleared, and only for a row **they** armed in that request: an issue the admin had deliberately scheduled for a date keeps that date. The 403 carries `status` and `stillScheduled` so the composer says which of the two happened. The content is always kept — throwing away what the admin just wrote because the meter is full would be its own kind of data loss.
- **A resume is never metered.** `dispatchNewsletter` on an issue already in `SENDING` is finishing a run the quota authorised once. Metering it again would double-count (usage is derived from `NewsletterSend` rows in the window, which by then include this issue's own delivered half, so a large issue could meter itself over its own band) and, because the refusal returns before the claim, would strand the issue in `SENDING`: half delivered, un-editable, un-cancellable, refused again on every tick.
- **The operator setting can only tighten a plan's band, never lift it**, in either layer. A cap a tenant's own admin could raise is not a protection of anybody else's deliverability, so it is a floor and not a lift — which is also what makes it safe to expose on the tenant settings form.

`min()` is taken across the plan band and both setting layers; `''` (the documented default) means "no override, follow the plan"; `0` switches broadcasts off entirely while every free-core mail path keeps working. **Fails open when no org resolves**, exactly like `planGate.ts` — today's grandfathered single-tenant production runs on the ENTERPRISE `default` org, whose band is unlimited, so this is a no-op there by construction.

### Which send paths the quota covers

**Metered** — the two entries of the one exported `BROADCAST_EMAIL_CATEGORIES` constant, so no call site compares a category string by hand:

| Path | Metered as |
|---|---|
| `POST /api/admin/announcements` with `email: true` | `announcement` — the mailed recipients only |
| `POST /api/admin/newsletters` with `action: 'send'` | `newsletter` |
| `POST /api/admin/newsletters/[id]/send` with `mode: 'now'` | `newsletter` |
| the `*/15` dispatch cron (`dispatchDueNewsletters`) | `newsletter` |
| a **resume** of an issue already in `SENDING` | *not metered* — already authorised, see above |

**Deliberately NOT metered** — free core, forever, and asserted by the spec with the band set to `0`:

- 1:1 messages and message notification mail
- verification, invitation and password-reset mail
- meeting invitations, meeting reminders, meeting-series reminders, RSVP mail
- weekly digests, mentor digests, interaction reminders, stage deadlines, re-engagement
- dormant first-contact check-ins
- **the in-app half of an announcement** — a bell notification is not mail, so an announcement sent *without* the e-mail box ticked spends nothing and can never be refused
- **"Send me a test copy"** of a newsletter — one mail to the requesting admin's own verified address; making careful proofreading cost audience would be the wrong incentive

### Mid-send behaviour, stated

The recipient count of the **entire** fan-out is checked before a single message leaves, so a send cannot cross the band half-way through. The one residual case is spelled out in the module and in the doc: this is a *check, not a lock*, so two admins pressing Send in the same second can both pass it and overshoot the band by at most the smaller of the two sends. Because usage is derived from the rows that were actually written, the next check sees the overshoot and refuses. That is a better trade than serialising every broadcast behind a row lock for a limit whose purpose is a monthly average.

## Changes

- **`src/lib/broadcastQuota.ts`** (new) — `BROADCAST_EMAIL_CATEGORIES` + `isBroadcastEmailCategory`, the UTC month window, per-category usage counters keyed by the category itself (adding a category without teaching it to count is a *type error*, not a silently unmetered blast), `resolveBroadcastBand`, `checkBroadcastQuota`, `broadcastQuotaError`, `broadcastQuotaStatus`.
- **`src/lib/newsletterDispatch.ts`** — audience and `already`-sent set resolved before the claim; per-tenant quota evaluation, skipped entirely for a resume; a refusal returns `reason: 'broadcast_quota_exceeded'` with the figures and touches nothing. A resumed run only ever asks for what is *left* to send, so finishing a half-delivered issue is never blocked by the half already delivered. The send loop and every counter are unchanged.
- **`src/app/api/admin/announcements/route.ts`** — the mailed-recipient list is computed once, before the `Announcement` row is written, and is both what the quota checks and what the loop sends (two copies of "who gets mailed" is how a metered count and a delivered count drift apart). Over-band → 403, no row, no notifications.
- **`.../newsletters/route.ts`**, **`.../newsletters/[id]/send/route.ts`** — turn the dispatcher's refusal into a 403 **and disarm the row** they armed for it: the created/armed issue goes back to `DRAFT` with `scheduledAt` cleared, an already-scheduled one is left on its date, and the 403 says which (`status`, `stillScheduled`).
- **`GET /api/admin/broadcast-quota`** (new) — the composer's meter; `limit: null` (and `used: null`) means unlimited.
- **`src/components/ui/BroadcastQuota.tsx`** (new) — one hook + one line component for **both** composers, and the localized refusal rendered from `code` + the figures rather than printed from the response body (a server-authored message is never shown verbatim, per `apiErrorMessage.ts`); it appends the "waiting as a draft" / "still scheduled" sentence when the route reports one.
- **Composers** — announcement page shows the meter next to the "also send by email" box (the only control there that spends it); newsletter page shows it under the send row and re-reads it after every dispatch.
- **`src/lib/settings.ts` + settings API + settings form** — `broadcastMonthlyRecipients`, default `''`, with the reasoning next to it. Empty is a legal value with its own meaning, so the field is a text box and only a non-empty non-number is reported as invalid.
- **i18n** — real EN/TR/DE for the usage line, the unlimited line, the refusal, the two post-refusal sentences and the setting label/hint/validation.
- **`docs/newsletter.md`** — a "The broadcast quota" section under § Sending with the table and the load-bearing properties.
- **`e2e/broadcast-quota.spec.ts`** (new) — over-band announcement 403s with the figures and writes nothing; a send-now newsletter 403s and the row is left `DRAFT` with `scheduledAt` null and no `NewsletterSend`; an in-app-only announcement still goes out with the band at `0`; a mentor's 1:1 message, a meeting invitation and a mentee question all still work for a tenant whose band is `0`; plus pure assertions that only the two categories are metered and that the UTC month window rolls the year correctly. The band is driven by a **per-org** `Setting` row, never the global layer, so it cannot break specs running in parallel.

Nothing here mutates a sent newsletter issue, and no schema change was needed.

## Review round

Three findings from the adversarial review of the first commit are fixed in the second (`d4c6a11`): the armed-after-refusal issue, the metered resume, and the docs that described the old behaviour. The fourth — that one tenant's exhausted band blocks a shared cross-tenant issue for everybody, and silently stops the cadence at scale — is a product decision about what "sent" means for an installation-wide editorial object, so it is written up with four options in **#2335** rather than changed unilaterally here.

## Verification

Static only — this container has no database and no Docker daemon, so the Playwright spec was written but not executed.

- [x] `npm run lint` clean (only the two pre-existing `exhaustive-deps` warnings in `admin/companies` and `admin/mentorship`)
- [x] `npm run build` passes
- [x] `npm run check:i18n` clean across EN/TR/DE
- [x] `npm run check:stage-keys` — no new hardcoded key
- [x] `npm run check:release-fragments` — fragment valid (ships as `0.182.0-beta`)
- [x] `npm run check:tenant-models` — no new drift
- [x] `npm run test:release` — 0 failures
- [x] Added an e2e test (`e2e/broadcast-quota.spec.ts`, not run here)

One observation left out of scope: `Announcement.emailedCount` increments even when `sendEmail` returns `SKIPPED` (an environment with no SMTP configured), so on such an environment the meter counts mail that never left. That is pre-existing behaviour of that counter, not introduced here.

Preview: https://pr2328.interncrm.com

## Contributor terms

Closes #1754


mersahin:
Review findings applied in d4c6a11.

**Fixed — a quota-refused "send now" stayed armed.** This was the worst of the three. `scheduledAt` for a send-now is `new Date()`, so the row the admin was told "nothing was sent" about was *due*: the 15-minute cron would have mailed it unattended on its next tick, and pressing Send again (the natural response to a refusal) armed a second issue that would also deliver. Both send-now paths now put the row back to `DRAFT` with `scheduledAt` cleared — and only for a row **they** armed in that request, tracked by an `armedHere` flag in `[id]/send`: an issue the admin had deliberately scheduled for a date keeps that date, because that queueing was a human decision and the band will very likely have reset by then. The 403 carries `status` and `stillScheduled`, and `useBroadcastQuotaMessage` appends the matching sentence (`broadcastQuota.keptDraft` / `.stillScheduled`, EN/TR/DE) so the admin is told which of the two happened. The content is always kept either way.

**Fixed — a resume was metered.** `dispatchNewsletter` on an issue already in `SENDING` is finishing a run the quota authorised once, so metering it again is a second toll on one decision. Worse, it double-counts: `checkBroadcastQuota` derives usage from `NewsletterSend` rows inside the window, which by then include this issue's own delivered half — a large issue could meter itself over its own band. And since the refusal returns *before* the claim, that left the issue stranded in `SENDING`: half delivered, un-editable, un-cancellable, refused again on every tick. `const resuming = issue.status === 'SENDING'` now guards the metering loop.

**Left to the maintainer — all-or-nothing per issue.** The reviewer is right that one tenant's exhausted band blocks a shared newsletter issue for every other tenant, and that at scale this converges on "no newsletters at all" with no operator-visible signal. But the current rule is documented with its reasoning (excluding a tenant would mark the issue `SENT`, and a sent issue is immutable and undeletable, so the skipped tenant could never receive it) — flipping it changes what "sent" means for a cross-tenant editorial object. That is a product decision, not a bug fix, so it is written up with four options in #2335 rather than changed here.

**Docs corrected.** `docs/newsletter.md` § The broadcast quota claimed the opposite of what now happens ("a refusal leaves it SCHEDULED and re-sendable"); it and the release fragment now carry the disarm rule and the resume rule as the load-bearing properties they are.

**Test.** `e2e/broadcast-quota.spec.ts` gains the assertion that closes the gap: with the band at zero, a send-now newsletter comes back 403 with `status: 'DRAFT'`, and the row is `DRAFT` with `scheduledAt` null, `sentCount` 0, and no `NewsletterSend` rows.

Verified locally: `tsc --noEmit`, `next lint`, `check:i18n`, `check:stage-keys`, `check:release-fragments`, and a full `npm run build` — all clean.

---

Issue #2323, opened 2026-09-09
🔑 release-compact.yml cannot open its own PR — needs RELEASE_BOT_TOKEN or the Actions PR permission


mersahin:
## The failure

Every scheduled `release-compact.yml` run does its job and then cannot deliver it:

```
pull request create failed: GraphQL: GitHub Actions is not permitted to create or approve pull requests (createPullRequest)
##[error]The branch bot/release-compact is pushed and correct (65 fragment(s) -> 0.181.0-beta),
        but creating the PR failed. Usual cause: the RELEASE_BOT_TOKEN secret is missing, and this
        org forbids GITHUB_TOKEN from creating pull requests. Open the PR from bot/release-compact
        manually, or add the secret.
```

Four runs in a row: [18](https://github.com/21072026/Internship/actions/runs/34333371629), [17](https://github.com/21072026/Internship/actions/runs/34208137107), [16](https://github.com/21072026/Internship/actions/runs/34107752398), [15](https://github.com/21072026/Internship/actions/runs/34036344283). Each force-pushed a correct branch and stopped there.

Consequence while it lasted: 65 shipped changes had no `CHANGELOG.md` entry and no `/release-notes` line, and `package.json` stayed at `0.156.21-beta` while the derived build version had moved to `0.181.0-beta`. I opened the pending one by hand as #2321, so the immediate backlog is cleared — but the next scheduled run will fail identically.

## The fix is a repo setting, not code

The workflow's own error message already names both options. Someone with repo admin has to pick one:

1. **Allow Actions to create pull requests** — Settings → Actions → General → Workflow permissions → tick *Allow GitHub Actions to create and approve pull requests*. Note it grants that to **every** workflow in the repo, which is a real (if small) widening: any workflow that can be made to run can then open a PR.
2. **Add a `RELEASE_BOT_TOKEN` secret** — a fine-grained PAT with `contents: write` + `pull-requests: write` on this repo only. Narrower, but it is a credential someone must rotate, and it expires silently.

Option 2 is the better default for a public repo. Option 1 is one click and no rotation.

Either way this is **not something an agent should decide or do** — it changes the repo's permission surface. That is why this is an issue and not a PR.

## What code CAN do, and should

Regardless of which option is chosen, the workflow currently fails in a way that produced no trail anybody read for three days:

[...]

Issue #2322, opened 2026-09-09
🚨 Every failure-alert e-mail job times out on GitHub-hosted runners — four release-compact failures went unnoticed


mersahin:
## What happened

`release-compact.yml` failed four scheduled runs in a row (15–18, 2026-09-06 → 2026-09-09) and nobody found out. The compaction itself was fine; the PR-creation step failed, and `bot/release-compact` sat there correct and unreferenced while 65 fragments piled up and `CHANGELOG.md` recorded nothing.

The reason it went unnoticed is the second failure in the same run. From [run 18](https://github.com/21072026/Internship/actions/runs/34333371629), job `Email alert on failure`:

```
Failed to send alert email: Error: Connection timeout
    at SMTPConnection._formatError (nodemailer/lib/smtp-connection/index.js:938:19)
    ...
  code: 'ETIMEDOUT',
  command: 'CONN'
```

The runner cannot open a TCP connection to the SMTP host at all — it never gets as far as authenticating. So the one mechanism meant to surface a red scheduled run is itself red, silently.

## Why this is worth more than the one workflow

The same "run a Node script that mails `ALERT_EMAIL_TO` when the job is red" pattern guards at least three scheduled safety nets, all of which are **silent when green by design** — which means a broken mailer is indistinguishable from a healthy repo:

- `scripts/e2e-report-email.mjs` — the 4×/day full Playwright suite. `E2E_REPORT_MODE` defaults to red-only.
- `scripts/k6-report-email.mjs` — the nightly load test. Breach-only; green is silent.
- the `release-compact.yml` alert above.
- whatever the dead-letter alert from #1674 uses.

If the SMTP path is broken for all of them, the repo currently has **no working out-of-band failure channel**. Every scheduled net degrades to "someone happens to look".

## What to work out first (do not guess)

1. **Is it the runner or the credentials?** `ETIMEDOUT` on `CONN` is a network-reachability symptom, not an auth one. GitHub-hosted runners block outbound SMTP on 25 and are unreliable on 465/587 to arbitrary hosts. Check which port the scripts use and whether the host is reachable from `ubuntu-latest` at all.
2. **Did this ever work?** Look for a run where an alert e-mail was actually delivered. If the answer is "not since the server migration (#2166)", the cause is the SMTP host moving to `s.interncrm.com` and the workflows still pointing at the old one — a completely different fix from a port problem.
[...]

PR #2329, opened 2026-09-09
ci: release-compact opens an issue when it cannot open its PR (#2323)

mersahin:
## Summary

Four scheduled `release-compact.yml` runs in a row did the work and could not deliver it: each force-pushed a correct `bot/release-compact`, was refused at `gh pr create` ("GitHub Actions is not permitted to create or approve pull requests"), and stopped there. Nobody noticed for three days — 65 fragments piled up while `CHANGELOG.md`, `src/lib/releaseNotes.ts` and `package.json`'s version stopped recording what shipped. The workflow *did* print the right message; it printed it into a run log nobody opens for a workflow that has always been green, and the alert e-mail that should have flagged it is separately broken (#2322).

This closes that loop without touching the repository's permission surface. **A workflow that cannot open a pull request can still open an issue** — no new secret, and no permission beyond `issues: write` on the workflow's own `GITHUB_TOKEN`.

Refs #2323 — deliberately **not** `Closes`. The issue's actual ask is a repository setting (allow Actions to create PRs, or add a `RELEASE_BOT_TOKEN` secret), which is the maintainer's decision and explicitly out of scope for an agent. That ask must stay open after this merges. What this PR delivers is the "What code CAN do, and should" section of the issue; tick those three boxes and the remaining one is yours.

## Changes

- **`scripts/release-compact-alert.mjs` (new)** — the failure trail.
  - `stuck` mode: opens, or **updates in place**, one issue labelled `release-compact-stuck` naming the branch, the fragment count, the version those fragments ship as, the failed run, and a one-click `compare/main...bot/release-compact?expand=1` link — the entire manual step is clicking it and pressing the green button.
  - Updating rather than duplicating is the load-bearing part: this fires **daily**, so "open an issue on failure" without the lookup is 30 identical issues in a month, which is its own kind of silence. It also **comments only when the situation moved** (count or version changed) — a daily "still stuck, same numbers" notification is how an alert gets muted.
  - `resolved` mode: **closes** that issue as soon as a compaction PR exists, so it never sits open lying about the current state.
  - Every `gh` call is best effort. If the issue cannot be filed at all (a PAT without `issues: write`, `gh` missing), the `::error::` annotation still goes out and the step still fails.
- **`::error::` annotation** — same content on one line (`%0A` breaks), so the run summary carries the branch, the count, the version, the compare link, the tracking issue and who fixes it permanently, without opening the log.
- **`.github/workflows/release-compact.yml`** — `issues: write`, the two script calls, and a `FAILURE TRAIL` header note. The **success path is untouched**: the push, the force-push, the fragment arithmetic and the stamping are byte-for-byte what they were.
- **`scripts/test/release-compact-alert.test.mjs` (new, 11 assertions)** — the only thing this code produces is text somebody has to act on, generated once a day at 04:45 UTC with nobody watching, so it is not left untested inside a YAML file: the issue names branch/count/version, carries the manual step and the accountable person, the compare link URL-escapes the branch's slash, an unchanged backlog does not comment, a moved one does, and the annotation is single-line and stands alone when the issue could not be filed. Picked up by `npm run test:unit` / `test:unit:coverage` via the existing `scripts/test/*.test.mjs` glob — no new CI step.
- **`scripts/check-release-fragments.mjs`** — the backlog warning ended *"Not a problem with this PR"*, which was accurate and got it correctly skipped by every reviewer across a whole session. It now says **who acts** (the maintainer, in #2323), **what the action is** (open the pending PR from `bot/release-compact`, then choose how Actions may open it in future), and what to do if #2323 is already closed and it still fires. Still warn-never-fail.
- **`releases/README.md`** — "When compaction is stuck" documents the new trail and the 2026-09-05..09 recurrence.

### Does the warning change belong here?

Yes, here — asked in the brief, so stating the reasoning: it is the same failure loop and about ten lines of text, and splitting it would put the "who is accountable" half of one fix behind a second review. It changes no behaviour and no exit code; the warning still cannot fail a PR.

## Verification

Static only — no database and no Docker in this container, so the app and Playwright cannot run. This change has no runtime surface (a workflow, two scripts, one doc).

- [x] `npm run lint` clean (pre-existing `react-hooks/exhaustive-deps` warnings only, untouched files)
- [x] `npm run build` passes
- [x] `npm run check:i18n` — 4777 keys × 3 locales (no new strings: nothing here is user-facing)
- [x] `npm run check:stage-keys` — no new hardcoded key
- [x] `npm run check:release-fragments` — well-formed, and the reworded backlog warning verified in its output
- [x] `npm run test:unit:coverage` — all suites pass, no floor breached
- [x] `node --test scripts/test/release-compact-alert.test.mjs` — 11/11
- [x] YAML re-parsed with `js-yaml` and the changed step's shell body checked with `bash -n`
- [x] Both script modes exercised end-to-end against a stubbed `gh`: first failure (creates), repeat failure with identical numbers (edits, **no** comment), repeat failure with a moved backlog (edits **and** comments once), resolved with an alert open (closes), resolved with none (silent), `gh` absent (warning + annotation, exit 0 so the workflow's own `exit 1` is what fails the run)

**No release fragment**: CI plumbing with no user-visible surface — nothing to put in `/release-notes` or the changelog. No i18n keys and no schema change for the same reason.

Preview: https://pr2329.interncrm.com

## Contributor terms

### DIFF
--- a/.github/workflows/release-compact.yml
+++ b/.github/workflows/release-compact.yml
@@ -38,6 +38,19 @@ name: Compact release fragments
 # + alert email) naming the branch — a maintainer opens the PR in one click, or
 # any agent session runs `node scripts/release-compact.mjs` and opens it
 # normally.
+#
+# FAILURE TRAIL (#2323). "Fails LOUDLY" above was not loud enough: four
+# scheduled runs in a row pushed a correct branch, failed at the PR step and
+# went unnoticed for three days, because an annotation lives in a run log
+# nobody opens for a workflow that has always been green, and the alert e-mail
+# is separately broken (#2322). So the failure path now also OPENS — or, on the
+# next failed run, UPDATES — a labelled issue naming the branch, the fragment
+# count, the version and the one-click manual step
+# (scripts/release-compact-alert.mjs), and CLOSES it again as soon as a
+# compaction PR exists. Updating rather than duplicating is deliberate: this
+# fires daily, and 30 identical issues would be its own kind of silence.
+# Nothing about the success path changed — the push, the force-push, the
+# arithmetic and the stamping are untouched.
 
 on:
   schedule:
@@ -53,6 +66,11 @@ concurrency:
 permissions:
   contents: write
   pull-requests: write
+  # A workflow that is forbidden from opening a pull request can still open an
+  # ISSUE, which is the whole failure trail of #2323 — see the FAILURE TRAIL
+  # note above. This is the default GITHUB_TOKEN's own issues scope; it grants
+  # nothing to any other workflow and needs no secret.
+  issues: write
 
 env:
   # One branch for every compaction, ever. See the header note.
@@ -132,7 +150,14 @@ jobs:
                 --head "$BRANCH" \
                 --title "chore(release): compact fragments into ${VERSION}" \
                 --body "$BODY"; then
-              echo "::error title=Could not open the compaction PR::The branch ${BRANCH} is pushed and correct (${COUNT} fragment(s) -> ${VERSION}), but creating the PR failed. Usual cause: the RELEASE_BOT_TOKEN secret is missing, and this org forbids GITHUB_TOKEN from creating pull requests. Open the PR from ${BRANCH} manually, or add the secret."
+              # The branch is pushed and correct; only the PR is missing. Leave
+              # a trail a person actually encounters — an issue plus the
+              # ::error:: annotation, both naming the branch, the count, the
+              # version and the manual step (#2323). Best effort: if even the
+              # issue cannot be filed, the annotation still goes out and this
+              # step still fails.
+              node scripts/release-compact-alert.mjs stuck \
+                || echo "::error title=Could not open the compaction PR::The branch ${BRANCH} is pushed and correct (${COUNT} fragment(s) -> ${VERSION}), but creating the PR failed and the failure could not be recorded as an issue either. Open the PR from ${BRANCH} manually; see #2323."
               exit 1
             fi
             PR="$(gh pr list --head "$BRANCH" --state open --json number --jq '.[0].number // empty' || true)"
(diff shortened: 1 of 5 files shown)


mersahin:
Both review findings applied in 23b1315.

**Fixed — the alert could not use the permission it was granted.** `permissions: issues: write` belongs to the workflow's own `GITHUB_TOKEN`, but the PR step runs with `GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN || secrets.GITHUB_TOKEN }}` and the script inherited it. So in exactly the configuration this alert *recommends adopting* — set `RELEASE_BOT_TOKEN` — every `gh` call in the script would have run as a PAT nobody promised carries an issues scope, and the failure trail would be silent in the one setup a maintainer had just been told to switch to. The step now also exports `ISSUE_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` and `gh()` prefers it, falling back to `GH_TOKEN` and then `GITHUB_TOKEN`, passing the chosen token as both `GH_TOKEN` and `GITHUB_TOKEN` in the child env so gh's own precedence cannot pick the other one.

**Fixed — the alert asserted a cause it had not verified.** Both the issue body and the annotation stated "GitHub Actions is not permitted to create pull requests" as fact, because that is what the four observed runs hit. `gh pr create` also fails on an expired PAT (`HTTP 401: Bad credentials`), a branch protection rule and an exhausted GraphQL quota — each needing a different fix — and an alert that confidently names the wrong cause sends the reader to the wrong repository setting, which is worse than one that quotes. So:

- the workflow captures stderr to `$RUNNER_TEMP/pr-create.err` (still `cat`-ed back to the log so the run reads normally) and passes it in as `PR_CREATE_ERROR`;
- the issue quotes it verbatim in a fenced block under **What `gh pr create` said**;
- the familiar refusal is *recognised* rather than assumed: when it matches, the body still names the repository decision and @mersahin; when it does not, the body says explicitly that this is not that refusal and warns against reaching for the setting; when nothing was captured (an older workflow revision, or a crash before the redirect) it says so and falls back to the known cause as the likely guess;
- the annotation carries one line of the real stderr instead of the hardcoded sentence;
- `formatCreateError` keeps the **tail** (gh prints its real message last), drops blank lines, and caps lines and characters so a pathological failure cannot post a megabyte into an issue.

**Tests.** Six new assertions cover both directions of the cause detection, the empty-capture case and the truncation. `node --test scripts/test/release-compact-alert.test.mjs` 16/16, `npm run test:unit` 273/273, `npm run test:unit:coverage` no floor breached, and the workflow parses as YAML.

---

